PR #57, opened 2012-10-09
BATCH-1684: Forcing the use of ISO-8859-1 encoding for persistence

mminella:
When converting from a ByteArrayInputStream to a String, default encoding is causing the round trip to not be symmetrical.


spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcExecutionContextDao.java
```diff
@@ -224,14 +224,17 @@ private String serializeContext(ExecutionContext ctx) {
 		}
 
 		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		String results = "";
+
 		try {
 			serializer.serialize(m, out);
+			results = new String(out.toByteArray(), "ISO-8859-1");
```
sudhishmadhavan: Hey Michael,
Isnt ISO-8859-1 making an ASCII assumption. Isnt UTF-8 or 16 preferrable? We are running into an issue when we run spring-batch on a UTF supported environment but isnt ASCII ( In this case its EBCDIC).

mminella: Please create issues or use StackOverflow to ask questions like this.

PR #214, opened 2013-08-13
Fix for BATCH-1767

pmouawad:
Hello,
Here is a fix for BATCH-1767, this is a blocker one for a project, I hope you can take it into account soon.

Note I have registered this day as SpringSource Community Contributor.
Regards
Philippe M.


bogste:
Well, any workaround for this issue then?
We're having trouble with parallel test execution, because of this optimistic lock strategy when a chunk write fails.

PR #292, opened 2014-03-09
Use bean names to avoid key collision in in execution context

jpraet:
Change ItemStreamSupport to implement BeanNameAware. The default name (in case the object is instantiated outside of a spring bean factory) is still the short form of the class name. The name property can still be explicitly configured by the user. When the user didn't configure the name, the BeanNameAware.setBeanName callback will initialize the name property to the bean name.


fmbenhassine:
As mentioned in https://github.com/spring-projects/spring-batch/pull/339#issuecomment-3381848754, I applied the patch in this PR and updated it as needed for v6: 5e48bc09ade567ad68a3d996cd6542f3b8138a9f

Credits to @jpraet for this!

PR #324, opened 2014-05-29
BATCH-2240 Call ItemStream open() and close() within a transaction

jpraet:
* Modify TaskletStep to call open() and close() of registered ItemStreams within a transaction
* Fix problem in TransactionAwareBufferedWriter where close callback was not being called if the transactional buffer was empty


mminella:
I'm closing this PR based on conversations in the Jira.  If the Jira conversation warrents, we can re-open this later.

PR #339, opened 2014-07-13
ItemStreamSupport component name collision detection

jpraet:
Alternative approach for https://github.com/spring-projects/spring-batch/pull/292.

Instead of trying to prevent ExecutionContext key collisions using
BeanNameAware, this PR throws an exception on update() when a duplicate
component name is detected.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ItemStreamSupport.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ItemStreamSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2013 the original author or authors.
+ * Copyright 2006-2014 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,17 +16,20 @@
 package org.springframework.batch.item;
 
 import org.springframework.batch.item.util.ExecutionContextUserSupport;
+import org.springframework.util.ClassUtils;
 
 /**
  * Empty method implementation of {@link ItemStream}.
  *
  * @author Dave Syer
  * @author Dean de Bree
- *
+ * @author Jimmy Praet
  */
 public abstract class ItemStreamSupport implements ItemStream {
 
 	private final ExecutionContextUserSupport executionContextUserSupport = new ExecutionContextUserSupport();
+	
+	private String collisionDetectionKey = ClassUtils.getShortName(getClass());
 
 	/**
 	 * No-op.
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@jpraet Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=339) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jpraet Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=339)!

jpraet:
We've had to deal with a production issue once again caused by this problem of multiple batch components of the same type persisting restart information in the execution context with colliding keys.

I feel like something should be done in the spring batch framework to prevent this common pitfall.

The issue specified in the reference docs, but it's just a little sidenote in chapter `6.13.1 Custom ItemReader Example`:

> It is also worth noting that the key used within the ExecutionContext should not be trivial. That is because the same ExecutionContext is used for all ItemStreams within a Step. In most cases, simply prepending the key with the class name should be enough to guarantee uniqueness. However, in the rare cases where two of the same type of ItemStream are used in the same step (which can happen if two files are need for output) then a more unique name will be needed. For this reason, many of the Spring Batch ItemReader and ItemWriter implementations have a setName() property that allows this key name to be overridden.

If nothing is done in the framework itself, I think it can be indicated more clearly as a warning in the docs. And should also be referred to from chapters `3.3 ExecutionContext` and `6.4. ItemStream`.

mminella:
My concern still stands about breaking restartability between versions.  I'm also not 100% sure that causing existing jobs to fail automatically because of this is also the ideal solution (if a step that uses a potentially colliding component is configured to not be restartable for example, the user wouldn't care).  Since we already call this out in the documentation, I'm wondering if a simple warning log message when it occurs may be enough so that it's clear what is happening when it occurs.

mminella:
To update on this, I'm having a conversation right now in [BATCH-2575](https://jira.spring.io/browse/BATCH-2575) about a potential breaking change between major versions related to this.  If we do that, then it may make more sense to allow for this kind of change.  Thoughts there are welcome!

jpraet:
Yes, if restart compatibility is being broken between 3.x and 4.x, then I think it's best to merge #292 (which is a better solution to this problem than this PR, as it prevents the problem, instead of simply detecting it).

jpraet:
As far as I know the underlying problem (key collisions in execution context when using multiple batch components of the same type without configuring a name) still exists.

#292 resolve this problem (but was not accepted because it's not backwards compatible regarding job restarts).

fmbenhassine:
I agree with https://github.com/spring-projects/spring-batch/pull/339#issuecomment-49204540, it's too late to throw an exception at runtime while updating the context. 

Ideally, this should be done at stream registration time or at last at startup time, but I don't see any extension point to handle that properly.

> #292 resolve this problem (but was not accepted because it's not backwards compatible regarding job restarts).

That's acceptable for a major version, I will merge #292 in favor of this one for v6.

PR #345, opened 2014-08-28
BATCH-2205: MethodInvokingTaskletAdaper does not support primitive arguments.

ksseo0701:
MethodInvokingTaskletAdaper does not support primitive arguments.
Because "isAssignableFrom(..)" method was used in AbstractMethodInvokingDelegator.targetClassDeclaresTargetMethod().

The original source are correct when Object[] are used as parameters.
(ex. public static class TestBean {
		public void execute(Integer integerValue) {
                    ...
		}
	}
)

But I think, using a primitive type parameter can be supported.
(ex. public static class TestBean {
		public void execute(int integerValue) {
                    ...
		}
	}
)


mminella:
Merged.  Thanks!

PR #351, opened 2014-11-02
BATCH-2311 JobLauncher allows restart of job_execution with status UNKNOWN

jpraet:
* additionally prevent restart of step executions with status STARTING,
STARTED or STOPPING
* additionally prevent restart of job executions with status UNKNOWN or
STOPPING


mminella:
Merged.  Thanks!

PR #352, opened 2014-11-04
BATCH-2317 Optimized JdbcPagingItemReader configuration for process indicator pattern

jpraet:
Allow absence of sortKeys when saveState=false.


jpraet:
I'm closing this because I don't think this is the right approach. I'll elaborate in the JIRA ticket.

PR #354, opened 2014-11-21
use current classloader to get the proxied repository

ChristianRamseier:
The job repository factory always tries to get the proxied object with the system's classloader. If it has been loaded by a class loader that has a parent, the lookup will fail as the class is not visible from the parent class loader.


mminella:
Merged.  Thanks!

PR #355, opened 2014-11-28
BATCH-2323: Added test for JobScope with JavaConfig and XML

thebignet:
I've added a test case for JobScope beans declared with XML as well as JavaConfig based on the test case for StepScope beans.  This test does not pass and displays the error described in the Jira issue.


mminella:
This test isn't ever going to work.  If you explicitly define a `JobScope` in your ApplicationContext and use `@EnableBatchProcessing`, things will break because you're adding both types.  I'm closing this and we can pick this conversation up in Jira.

PR #358, opened 2015-01-26
Expose createJobLauncher to sub-classes of DefaultBatchConfigurer

robfletcher:
Looks like this is supposed to be an extension point and I'd like to use it in tests so I can use a `JobLauncher` with a `TaskExecutor` that's under my control. I want to do things like killing the `TaskExecutor` running the jobs and test recovery, test multi-threading, etc.


mminella:
Merged.  Thanks!

PR #361, opened 2015-02-23
BATCH-2353 update sql server ddl to support ANSI NULL Default = false

pellcorp:
For BATCH-2353 master


mminella:
Merged.   Thanks!

PR #362, opened 2015-03-13
fix for BATCH-2359

sparsick:


mminella:
Merged.  Thanks for this!

PR #363, opened 2015-04-08
Add missing braces in example configuration

wilkinsona:


mminella:
Merged.  Thanks for the catch!

PR #364, opened 2015-04-30
BATCH-2380: disallow null + empty delimiter

mouyang:

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineTokenizer.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineTokenizer.java
@@ -76,6 +76,7 @@ public class DelimitedLineTokenizer extends AbstractLineTokenizer {
 	 * @param delimiter the desired delimiter
 	 */
 	public DelimitedLineTokenizer(String delimiter) {
+		Assert.notNull(delimiter);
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineTokenizer.java
```diff
@@ -133,6 +134,8 @@ public void setQuoteCharacter(char quoteCharacter) {
 	@Override
 	protected List<String> doTokenize(String line) {
 
+		Assert.state(null != delimiter && 0 != delimiter.length());
```
mminella: We don't want to be doing this for every record.  We also would want to know there's an issue before this point.  What about implementing `InitializerBean` and providing the check there?

mouyang: I agree.  It is not expected that the delimiter will change after a bean has been initialized.

mminella:
Two last, minor, requests:  

1. Can you create this PR against master instead of 3.0.x?  While the null check isn't really a breaking change, the empty string technically is.
2.  Going forward, the convention for branching is to name the branch after the Jira issue it's solving.  So in your case, the branch would be called BATCH-2380.

mouyang:
will do

PR #367, opened 2015-06-06
stop job execution if an exception is thrown in an after step

glefloch:
This pull request will stop the job if an exception is thrown in an after step.
This way we won't go into the next with a potential weird state due to an incomplete after step.
This way the developer can easily throw exception in the after step. If exception are non blocking he can just catch them and that's it.


mminella:
Can you create a Jira issue for this so we can discuss the functionality there?

glefloch:
Hi, 
I just created the issue here is the link: https://jira.spring.io/browse/BATCH-2391
so we can discuss the functionalit there.
Regards

mminella:
I'm going to close this PR until we can figure things out in Jira.  We can re-open it once the discussion in Jira is resolved.

PR #368, opened 2015-06-13
BATCH-2380: disallow null + empty delimiter

mouyang:
resubmission of https://github.com/spring-projects/spring-batch/pull/364
base off master instead of 3.0.x


mminella:
Thanks for the contribution!

PR #371, opened 2015-06-24
BATCH-2402 Refactored `DelimitedLineTokenizer` to avoid temporary String allocation

jongray:
Refactored the `DelimitedLineTokenizer` (behaviour is exactly the same) to avoid temporary `String` allocation in order to be more memory space efficient.  This helps to avoid garbage collection with large numbers of batch lines and large numbers of quoted tokens.

I made these changes after profiling a batch which was reading ~10 million lines from a tab-delimited file, each line with ~30 quoted columns. With these changes, the total memory of Strings allocated during the read was reduced by approximately 75%.  As a result of that, in my test environment (1.8.0_45, 64bit, 16GiB system memory, default heap sizes) the read runtime was reduced by approximately 30% due to the reduction of the number of garbage collections.

I think the tradeoff in terms of increase in code complexity is worth the performance benefit.


mminella:
Very nice work.  Thanks for the update!  Merged

PR #374, opened 2015-07-06
Remove remaining dependencies on ParameterizedRowMapper

wilkinsona:
This is a follow-on from adc2f79. It removes the remaining references
to ParameterizedRowMapper and should, therefore, make things fully
compatible with Spring Framework 4.2.


mminella:
Merged and thanks!

PR #375, opened 2015-07-08
Use the latest version of the Spring IO Plugin

wilkinsona:
This change is necessary to allow Spring IO Platform 2.0 to remove the platform-versions properties file that was deprecated in 1.1. To allow Spring Batch's Platform compliance to be verified as part of Spring IO Platform 2.0's build and release process, I'd like this to be included in Spring Batch 3.0.x which is the version that's currently in Spring IO Platform 2.0. If you have any questions, please let me know.


mminella:
Thanks!

PR #376, opened 2015-07-12
Add regression unit test covering behavior reported in BATCH_2346

ajablonski:
Observed that deleting the code fixing issue BATCH_2346 didn't cause
any tests to fail


mminella:
Merged and thanks!

PR #377, opened 2015-07-28
BATCH-2406 Don't synchronize on LinkedBlockingQueue

marschall:
LinkedBlockingQueue has its own concurrency control mechanisms that is
orthogonal to the synchronization provided by the Java keyword
synchronized.

See also the Findbugs pattern JLM_JSR166_UTILCONCURRENT_MONITORENTER [1]

 - remove synchronized on LinkedBlockingQueue

 [1] http://findbugs.sourceforge.net/bugDescriptions.html#JLM_JSR166_UTILCONCURRENT_MONITORENTER

Issue: BATCH-2406

### DIFF
--- a/spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/ChunkMessageChannelItemWriter.java
+++ b/spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/ChunkMessageChannelItemWriter.java
@@ -19,7 +19,7 @@ package org.springframework.batch.integration.chunk;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


marschall:

mminella:
Per the documentation of BlockingQueue, the bulk operations do not have the same guarantees that individual addition/removal methods do.  See: http://stackoverflow.com/questions/2695426/are-linkedblockingqueues-insert-and-remove-methods-thread-safe and http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/BlockingQueue.html

Also, the FindBugs note you reference points out that there are use cases where this usage is valid, it just should be reviewed carefully.

PR #378, opened 2015-07-29
BATCH-2407 Use StringBuilder instead of StringBuffer

marschall:
With Java 1.5 StringBuilder is preferred over StringBuffer for single
threaded access as is has less overhead.

This is evidenced by the class comment of StringBuffer and Effective
Java 2nd Edition Item 67: Avoid excessive synchronization.

> The StringBuilder class should generally be used in preference to
> this one, as it supports all of the same operations but it is faster,
> as it performs no synchronization.

 - replace StringBuffer with StringBuilder where possible

Issue: BATCH-2407

### DIFF
--- a/spring-batch-core-tests/src/test/java/test/jdbc/datasource/DataSourceInitializer.java
+++ b/spring-batch-core-tests/src/test/java/test/jdbc/datasource/DataSourceInitializer.java
@@ -137,10 +137,10 @@ public class DataSourceInitializer implements InitializingBean, DisposableBean {
 	}
 
 	private String stripComments(List<String> list) {
-		StringBuffer buffer = new StringBuffer();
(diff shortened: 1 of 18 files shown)


marschall:

mminella:
There are a large number of tests failing with this PR.  Please be sure to perform a full build before submitting a pull request.

marschall:
I fixed the tests and rebased and amended. 

`FaultTolerantStepFactoryBeanRollbackTests.testMultithreadedSkipInWriter()`  fails from time to time but does so on `3.0.x` as well. I assume that's a normal threading issue not related to this change but maybe related to the non-atomic increment operations in `StepContribution`

mminella:
Merged.  Thanks!

PR #379, opened 2015-07-30
BATCH-2408 Use ArrayList instead of Stack

marschall:
Stack is a pre Java 1.2 class that is slower than ArrayList as it
incurs additional synchronization overhead.
- replace Stack with ArrayList where it's not public API

Issue: BATCH-2408


marschall:

mminella:
LGTM.  Thanks for the update!  Merged as dbe1679890b593fc7b53a436c61893aa33688785

PR #380, opened 2015-07-30
BATCH-2409 OracleLobHandler is deprecated

marschall:
Spring has deprecated OracleLobHandler in favor of DefaultLobHandler
for the Oracle 10g driver and higher and recommends using that driver
even against 9i databases. Extended Support for Oracle 9.2 ended in
July 2010.

 - switch OracleLobHandler to DefaultLobHandler

Issue: BATCH-2409


marschall:

mminella:
See comments made in the related jira.

PR #381, opened 2015-07-30
BATCH-2410 junit.framework.Assert is deprecated

marschall:
junit.framework.Assert has been deprecated in favor of
org.junit.Assert.

 - replace junit.framework.Assert with org.junit.Assert

Issue: BATCH-2410


marschall:

mminella:
Thanks for taking the time to clean this up!

PR #382, opened 2015-08-03
BATCH-2411 Remove @SuppressWarnings("unchecked")

marschall:
The source code contains various @SuppressWarnings("unchecked") that
are no longer unnecessary since the relevant APIs have been upgraded to
support Java 5 generics.

 - remove @SuppressWarnings("unchecked") where no longer unnecessary

Issue: BATCH-2411


marschall:

mminella:
Merged.  Thanks for cleaning that up!

PR #383, opened 2015-08-04
BATCH-2412 Add guards around debug statements

marschall:
In various places in the code base a debug message is constructed
unconditionally. This adds unnecessary overhead in the common case when
debug logging is off.

 - add guards around debug statements in non-test code

Issue: BATCH-2412


marschall:

PR #384, opened 2015-08-10
BATCH-2417 Drop JAXP 1.3 support

marschall:
Spring 4 requires JDK 6 which includes JAXP 1.4. In Spring 4
`org.springframework.util.xml.StaxUtils` is public and directly
delegates to `javax.xml.transform.stax.StAXSource`. This means that
`org.springframework.batch.item.xml.StaxUtils#hasSpring30StaxSupport`
will always be true. Therefore a lot of code an be simplified by
calling the constructors directly instead of through reflection.
- remove reflection no longer needed in Spring 4
- move methods only used in tests to custom test class

Issue: BATCH-2417


marschall:
have signed and agree to the terms of the SpringSource Individual

mminella:
Rebased and merged as 73ca73688d218aafa985830b7306f9ff83c961d3

PR #385, opened 2015-08-19
Improve logging - show job duration in human readable format

ptahchiev:


mminella:
Given that the time is already part of the logs by default (or is configurable based on your logging configuration), what is this adding that isn't already there?

pivotal-issuemaster:
@ptahchiev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=385) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ptahchiev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=385)!

pivotal-issuemaster:
@ptahchiev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=385)!

benas:
> what is this adding that isn't already there?

@mminella It adds the math. Certainly we can look at the first line and last line of the logs, extract the timestamps and calculate the difference. For small and quick jobs, that would be easy. However, for jobs that run for hours and produce a big chunk of logs, finding the first/last line manually and do the math can be tedious, so letting the framework do this task has an added value IMO. I see it like a maven or gradle job printing the job duration in the end. Do you agree?

> specify duration in milliseconds

I would not show the duration in milliseconds as these logs are intended for humans. I would print it in a human readable format, something like: `3h 15m 45s` instead of `11745000ms` for example.

The changes in this PR use:

```java
long start = System.currentTimeMillis();
...
System.currentTimeMillis() - start
```

I would instead use the `java.time.Duration` API instead now that we have Java 8 as a base line, something like:

```java
Duration duration = Duration.between(jobExecution.getStartTime().toInstant(), jobExecution.getEndTime().toInstant());

// and format it with duration.toHours(), duration.toMinutes(), etc
```

@ptahchiev What do you think?

ptahchiev:
@benas Sounds good to me.

benas:
@ptahchiev Great! I created a JIRA ticket for this improvement: [BATCH-2775](https://jira.spring.io/browse/BATCH-2775).

Can you please update the PR accordingly? Many thanks upfront!

benas:
@ptahchiev Any update on this PR?

benas:
> @benas Sounds good to me.

@ptahchiev I added the job duration in the logs as discussed previously (See 10f22819963145482deeed2ff13004b3a21ab5a9). Thank you for your contribution anyway!

PR #386, opened 2015-08-27
BATCH-2161 ScopeConfiguration methods should be static

marschall:
The scope bean factory methods in ScopeConfiguration are
BeanFactoryPostProcessors and should therefore be static.
- Declare the scope beans in ScopeConfiguration as static

Issue: BATCH-2161

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
@@ -129,19 +129,18 @@ public abstract class AbstractBatchConfiguration implements ImportAware {
 @Configuration
 class ScopeConfiguration {
 
-	private StepScope stepScope = new StepScope();
-
-	private JobScope jobScope = new JobScope();
-
 	@Bean
-	public StepScope stepScope() {
+	public static StepScope stepScope() {
+		StepScope stepScope = new StepScope();
 		stepScope.setAutoProxy(false);
 		return stepScope;
 	}
 
 	@Bean
-	public JobScope jobScope() {
+	public static JobScope jobScope() {
+		JobScope jobScope = new JobScope();
 		jobScope.setAutoProxy(false);
 		return jobScope;
 	}
+
 }


marschall:

ractive:
:+1:

OzoneGrif:
+1 for this fix.

AlexanderHagenhoff:
:+1: need it

jntakpe:
Sure need it 👍

thaarbach:
+1

bhuism:
+1

petrdvorak:
@mminella Michael, this seems like a tiny pull request - minor edits in one file. Could you please merge it? It causes unnecessary warnings in Spring Boot console...

fabn:
Really, 2 years to review and merge a simple PR like this one??

mminella:
There is a comment in the code that we did not want the methods to be static.  The warning this produces is not a Spring Batch one, it's Spring Core which does not apply to the use case implemented here (We are not processing `@Autowired`, `@Resource`, or `@PostConstruct` annotations).  I'm closing this PR as works as designed.  If this is an issue with the community, maybe we should take it up with the Spring Framework team to get that warning moved to a debug message.

petrdvorak:
OK, fair explanation.

> maybe we should take it up with the Spring Framework team to get that
> warning moved to a debug message

I will try little discussion here, in effort to fix the issue eventually.

An in-place warning suppression would IMO be a better solution in this case. Moving the warning to `DEBUG` level globally does not make much sense to me - there probably are legitimate situations where the warning could be handy. However, if this is a specific situation in Spring Batch where the warning is not correct, it should be suppressed...

This is the Spring Framework code that triggers the warning:

https://github.com/spring-projects/spring-framework/blob/7ae59d0c2a5313ad96ba10f9d8ee4e7f77c45c85/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassEnhancer.java#L351

I could attempt the fix, but need couple hints. ;-) Could someone suggest a preferred way to suppress the warning? Some custom annotation? Or is there some existing annotation that would suit the purpose (like `@SuppressWarning`, but for runtime, not for compile time)? Or maybe some logging scope related annotation (such as `@Debug`) that we could use here to override default log levels? ...

mminella:
To be clear, you'll want to take it up with the Spring Framework team via a Jira issue before issuing a PR.  They may not feel the same way we do about this.

marschall:
> There is a comment in the code that we did not want the methods to be static.

Could you explain the reasoning behind this as this may help with filing a JIRA against Spring Framework.

mminella:
I didn't write that code so I'm not 100% sure, but my bets are:
1.  There isn't a reason for it to be static in the first place so just avoid it.
2.  We probably don't want to limit the Job and Step scope to one per JVM, but one per context.

marschall:
> There isn't a reason for it to be static in the first place so just avoid it.

That makes it hard to get a JIRA against Spring Framework accepted.

> We probably don't want to limit the Job and Step scope to one per JVM, but one per context.

The PR / making the method static does not prevent this. It still allows for a Job and Step scope per Spring application context.

Reading the comment again I believe the comment suggests the intention was to avoid creating a static _class_ and not avoiding static methods.

jhoeller:
From a quick glance: Such factory methods should indeed be declared as static simply because they do not access any state in the configuration class instance. This is even recommendable from a design perspective, making it clear that no such stateful configuration class instance is needed for those beans. 

Since we are only talking about static methods here and not about global static state, it does not affect the sharing between contexts. Those beans will still be locally instantiated for contexts configured with that configuration class. In other words, there is no shared-per-JVM impact here, just a local indication for what it takes to call such a factory method within the containing configuration class.

As for the warn log message, it essentially indicates that we can't fully initialize the configuration class instance. To see the effect of this, try adding an autowired field to the configuration class: We can't populate it since we are being called too early, i.e. before any annotation post processors are ready. Even if no such need is present in that configuration class right now, it indicates a fragile lifecycle mismatch.

mminella:
@dsyer I realize you wrote this code a while ago but since you wrote it and made the specific comment on the class, any thoughts?

PascalSchumacher:
@benas I would be great if this was resolved.

dsyer:
I think it's fine to be static as long as it doesn't create global state (seems like I made a wrong assumption there). The comment on the class can be deleted if we make the methods static.

benas:
@PascalSchumacher We are preparing to release 3.0.9/4.0.1 soon and this PR will be part of this release. keep tuned.

PascalSchumacher:
Great. Thank you very much!

benas:
@marschall Thank you for your PR! It was rebased and merged as 1e938e3e2b6fab5d8382e869a9ba15f84d1947af and will be included in the upcoming 3.0.9/4.0.1 release.

Stay tuned.

PR #387, opened 2015-09-13
BATCH-2434 Improve TransactionAwareBufferedWriter

marschall:
TransactionAwareBufferedWriter offers a number of optimization
potentials. First it buffers at the char level. Buffering at the byte
level instead saves about 50% memory usage in common cases. Second it
does not overwrite any of the  #write(String) methods leading to
unnecessary intermediate copies.
- buffer in a ByteArrayOutputStream instead of StringBuilder
- overwrite #write(String) methods to avoid copies

Together these two changes should help to reduce both live set size and
allocation rate.

Issue: BATCH-2434
https://jira.spring.io/browse/BATCH-2434

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2012 the original author or authors.
+ * Copyright 2006-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -210,9 +210,7 @@ public class TransactionAwareBufferedWriter extends Writer {
 	public void write(char[] cbuf, int off, int len) throws IOException {
 
 		if (!transactionActive()) {
-			char [] subArray = new char[len];
-			System.arraycopy(cbuf, off, subArray, 0, len);
-			byte[] bytes = new String(subArray).getBytes(encoding);
+			byte[] bytes = new String(cbuf, off, len).getBytes(encoding);
 			int length = bytes.length;
 			ByteBuffer bb = ByteBuffer.wrap(bytes);
 			int bytesWritten = channel.write(bb);
@@ -224,5 +222,28 @@ public class TransactionAwareBufferedWriter extends Writer {
 
 		StringBuilder buffer = getCurrentBuffer();
 		buffer.append(cbuf, off, len);
+	}
+
+	/*
+	 * (non-Javadoc)
+	 *
+	 * @see java.io.Writer#write(char[], int, int)
+	 */
+	@Override
+	public void write(String str, int off, int len) throws IOException {
+
+		if (!transactionActive()) {
+			byte[] bytes = str.substring(off, off + len).getBytes(encoding);
+			int length = bytes.length;
+			ByteBuffer bb = ByteBuffer.wrap(bytes);
+			int bytesWritten = channel.write(bb);
+			if(bytesWritten != length) {
+				throw new IOException("Unable to write all data.  Bytes to write: " + len + ".  Bytes written: " + bytesWritten);
+			}
+			return;
+		}
+
+		StringBuilder buffer = getCurrentBuffer();
+		buffer.append(str, off, len);
 	}
 }


marschall:

benas:
Hi @marschall ,

Thank you for this PR!

> Buffering at the byte level instead saves about 50% memory usage in common cases

How did you measure this? I would like to be able to measure the improvement of this PR before applying any change.

marschall:
> Hi @marschall ,
> 
> Thank you for this PR!
> 
> > Buffering at the byte level instead saves about 50% memory usage in common cases
> 
> How did you measure this? I would like to be able to measure the improvement of this PR before applying any change.

`TransactionAwareBufferedWriter` uses a `StringBuilder` to buffer. `StringBuilder` uses a `char[]` to buffer. A `char` is 16 bit so we end up using 16 bit for every character.
With the change `TransactionAwareBufferedWriter` uses `ByteArrayOutputStream` to buffer.  `ByteArrayOutputStream` uses a `byte[]` to buffer. When a 1 byte encoding is used like ASCII or Latin-1 we end up using only 8 bit for every character. Even if we use UFT-8 most characters end up only using 8 bit.

This changes a bit with Java 9+ where `StringBuilder` uses a `byte[]` with a flexible encoding. When the whole string is Latin-1 then only a single byte is used for storing a character, otherwise two bytes are used as before.

Doing a heap dump while a transaction is active and looking at the retained heap of a `TransactionAwareBufferedWriter` should reveal the differences on Java 8. On Java 9+ a single non Latin-1 character like € has to be in the buffer, otherwise they should be equal.

benas:
Thank you for your answer, I was aware of the theory and just wanted to know how you practically measured the 50% improvement in terms of memory usage.

I see the PR is opened against the `3.0.x` branch, can you please rebase it on the latest master?

I will add some inline comments about the changes.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -210,19 +274,22 @@ public void flush() throws IOException {
...
 		if (!transactionActive()) {
-			char [] subArray = new char[len];
-			System.arraycopy(cbuf, off, subArray, 0, len);
-			byte[] bytes = new String(subArray).getBytes(encoding);
-			int length = bytes.length;
-			ByteBuffer bb = ByteBuffer.wrap(bytes);
-			int bytesWritten = channel.write(bb);
-			if(bytesWritten != length) {
-				throw new IOException("Unable to write all data.  Bytes to write: " + len + ".  Bytes written: " + bytesWritten);
+			if (len > 8192) {
```
benas: Same comment as previously mentioned for `8192`. And it looks like this block of code is the same as the previous one (Lines 249-260). If it is the case, can we extract this in a private method?

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -71,6 +75,47 @@ public TransactionAwareBufferedWriter(FileChannel channel, Runnable closeCallbac
 	public void setEncoding(String encoding) {
 		this.encoding = encoding;
 	}
+	
+	/**
+	 * Buffers the output on a {@code byte[]} level. Compared to
+	 * buffering on a {@code char[]} level this saves about 50% memory
+	 * for common encodings and input.
+	 */
+	static final class Buffer {
```
benas: Do we really need this new `Buffer` class? Can't we use `java.nio.ByteBuffer` or probably `org.springframework.core.io.buffer.DataBuffer`? I'm not sure if this is possible, I'm just trying to see if we can reuse an existing data strucuture before creating a new one.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -197,8 +239,30 @@ public void close() throws IOException {
...
 		if (!transactionActive() && forceSync) {
-			channel.force(false);
+			outputStream.getChannel().force(false);
+		}
+	}
+	
+	@Override
+	public void write(String str, int off, int len) throws IOException {
+		if (!transactionActive()) {
+			if (len > 8192) {
```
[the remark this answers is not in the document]
marschall: The code tries to be really clever but by looking at `OutputStreamWriter#write` it will eventually end up calling `sun.nio.cs.StreamEncoder#write(String, int, int)` which will end up allocating a `char[len]` so it won't actually end up saving anything. We would have to iterate for that.
I'll simply remove the code.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
marschall: I'll remove this as well.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
marschall: I don't think so. `java.nio.ByteBuffer` has a fixed maximum size (capacity) and can not grow beyond this. We don't really know how much we have to buffer, so we would have to allocate new ones once we exceed their capacity.
`org.springframework.core.io.buffer.DefaultDataBuffer` is just wrapper around a `java.nio.ByteBuffer` and ends up inheriting its limitations.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -113,18 +162,11 @@ public void beforeCommit(boolean readOnly) {
...
-					StringBuilder buffer = (StringBuilder) TransactionSynchronizationManager.getResource(bufferKey);
+					Buffer buffer = (Buffer) TransactionSynchronizationManager.getResource(bufferKey);
 					if (buffer != null) {
-						String string = buffer.toString();
-						byte[] bytes = string.getBytes(encoding);
-						int bufferLength = bytes.length;
-						ByteBuffer bb = ByteBuffer.wrap(bytes);
-						int bytesWritten = channel.write(bb);
-						if(bytesWritten != bufferLength) {
-							throw new IOException("All bytes to be written were not successfully written");
```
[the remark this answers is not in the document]
marschall: This can no longer happen because we now call `java.io.ByteArrayOutputStream#writeTo(OutputStream)` which writes the entire buffer (and propagates any exception).

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -57,12 +61,12 @@
 	 * to execute on close. The callback should clean up related resources like
 	 * output streams or channels.
 	 *
-	 * @param channel channel used to do the actual file IO
+	 * @param outputStream channel used to do the actual file IO
 	 * @param closeCallback callback to execute on close
 	 */
-	public TransactionAwareBufferedWriter(FileChannel channel, Runnable closeCallback) {
+	public TransactionAwareBufferedWriter(FileOutputStream outputStream, Runnable closeCallback) {
```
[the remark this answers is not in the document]
marschall: done

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemWriter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
```
benas: This change should be rolled back since the file has not been changed.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -72,6 +99,47 @@ public void setEncoding(String encoding) {
...
+
+		void append(char[] cbuf, int off, int len) throws IOException {
+			this.writer.write(cbuf, off, len);
+		}
+
+		void append(String str, int off, int len) throws IOException {
+			this.writer.write(str, off, len);
+		}
+
+		byte[] getBytes() throws IOException {
```
benas: This method is not used and can be removed.

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
```diff
@@ -72,6 +99,47 @@ public void setEncoding(String encoding) {
...
+		private ByteArrayOutputStream outputStream;
+
+		private OutputStreamWriter writer;
+
+		Buffer(String encoding) throws UnsupportedEncodingException {
+			this.outputStream = new ByteArrayOutputStream();
+			this.writer = new OutputStreamWriter(outputStream, encoding);
+		}
+
+		int length() throws IOException {
```
benas: This method is used in `org.springframework.batch.support.transaction.TransactionAwareBufferedWriter#getBufferSize`. According to its Javadoc, the method should return the size of unflushed buffered data. However, the method `length` flushes the writer before returning the size of the outputStream. Do we need to flush the writer to get the length of the buffer? A `get` method should be side-effect free IMO. I tried to remove the `writer.flush` but this makes several tests to fail. Wdyt?

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemWriter.java
marschall: fixed

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
marschall: fixed

spring-batch-infrastructure/src/main/java/org/springframework/batch/support/transaction/TransactionAwareBufferedWriter.java
marschall: The issue is that `OutputStreamWriter` has an internal buffer in `sun.nio.cs.StreamEncoder`. If we want to know how many bytes we will write this will be the bytes in the `ByteArrayOutputStream` plus the bytes in the `OutputStreamWriter` / `sun.nio.cs.StreamEncoder`. There is no way of accessing the number of bytes in the internal buffer in `OutputStreamWriter` / `sun.nio.cs.StreamEncoder`. The only way to be sure is to flush those bytes to the `ByteArrayOutputStream`.

benas:
@marschall Thank you for all these updates! While the build is passing with this PR, I still have a couple of concerns:

1. According to its javadoc, `TransactionAwareBufferedWriter` is designed to be a wrapper for a file channel and the fact that we are changing its constructor to accept an output stream instead of a file channel makes me nervous (even if an output stream has a file channel associated with it). I feel like we can implement the improvement without introducing this API breaking change (that we are trying to make non breaking by deprecating the current constructor in favor of the new one)
2. `TransactionAwareBufferedWriter` is used in all file item writers (flat, json and xml files). So any hidden regression related to this can have a huge impact. The reason I'm pointing that out is that while we have some pretty good unit tests for this class, I don't see any integration test covering the "file write failed + restart" scenario where Spring Batch should truncate the file to the last known offset before starting to write items from where it left off in the previous failed run. I can take care of creating such a test suite (a test for each file type json/flat/xml + different encodings: single byte and multi-byte), but I really want to make sure there are no regressions at this level before merging this PR.

@marschall What do you think?

I also would like to have a second opinion on that PR, @mminella What are your thoughts? Anything I could have missed or overlooked?

Thank you both.

marschall:
@benas can you be a bit more specific what changes you would like to see or which approach you would prefer? A deprecated constructor is provided for backwards comparability. Should I change the class Javadoc as well?

benas:
We have been discussing this with Michael and here are our thoughts. Since the PR contains two improvements, I will address them separately.

### 1. buffer in a ByteArrayOutputStream instead of StringBuilder

> This changes a bit with Java 9+ where StringBuilder uses a byte[] with a flexible encoding. 

Indeed, this basically [JEP 254](http://openjdk.java.net/jeps/254). Since v4.3 will be the last version in the v4 line (as [announced](https://spring.io/blog/2019/12/04/spring-batch-4-0-4-4-1-3-and-4-2-1-available-now) last year) and since the next major version will have a Java9+ baseline anyway, we think the deprecation + additional complexity here is not worth it for the remaining short lifetime of v4.

### 2. overwrite #write(String) methods to avoid copies

This enhancement makes sense and could be merged. I tried it [here](https://github.com/benas/spring-batch/commit/704f159fd0df1d954f126e7d3e754ce7506107f7) and the build passes without any regression.

@marschall Could you please update the PR to keep only the second enhancement? It should be good to merge with that. Thank you upfront.

marschall:
@benas yes I can do this

marschall:
I pushed the new version.

benas:
LGTM. Rebased and merged as 10fd3716f03a6de9b024cc054a9f31e3ad38c105. Thank you very much for your contribution!

PR #388, opened 2015-09-13
BATCH-2435 FlatFileItemWriter buffers entire chunk

marschall:
FlatFileItemWriter currently buffers the entire string conversion of a
chunk in memory. There is no need to do that, it can just as easily be
streamed out.

 * stream out individual lines of a chunk

This change should help to reduce both live set size and allocation
rate.

Issue: BATCH-2435
https://jira.spring.io/browse/BATCH-2435


mminella:
See my comments associated with this in the Jira issue.

PR #389, opened 2015-09-18
fix assert message in PatternMatchingCompositeLineMapper.java

beartech1:
per BATCH-2427, changed assertion message to say "The 'patternMatcher' property must be non-null"


mminella:
Thanks and merged as d23497b0c936ecb4611fc1b89117370abfa9a7cd

PR #390, opened 2015-10-01
Update TaskletStep.java

knowcean:
The save of the old stepExecution should be done after the thread get the lock, otherwise, it may overwrite committed changes made from another thread during roll back.


mminella:
The TaskletStep is a very sensitive piece of the framework.  Please be sure all tests pass before submitting a PR making any changes to it.  In the current state, there are two tests that fail.

PR #391, opened 2015-10-07
Add shutdown to ThreadPoolTaskExecutor

markrk505:

Associated with ticket https://jira.spring.io/browse/BATCH-2439

This fixes an issue where the thread pool is not properly cleaned up which leaves any threads it has created blocked in memory.


mminella:
Merged as f6d22f4f806592b3545c1a363c9ce8e8c60aeb06.  Thanks!

PR #392, opened 2015-10-17
Fix Java configuration example autowiring in docs

lkrnac:


lkrnac:
Signed Individual Contributor Agreement

mminella:
Updated the spelling error but will leave the other config difference (no need for `@Qualifier`).  Thanks for pointing this out!

lkrnac:
There are used `step1()` and `step2()` without parameters, but that class contains only methods with parameters:

    @Bean
    protected Step step1(ItemReader<Person> reader, ItemProcessor<Person, Person> processor, ItemWriter<Person> writer) {
        return steps.get("step1")
            .<Person, Person> chunk(10)
            .reader(reader)
            .processor(processor)
            .writer(writer)
            .build();
    }

    @Bean
    protected Step step2(Tasklet tasklet) {
        return steps.get("step2")
            .tasklet(tasklet)
            .build();
    }

Are these methods defined elsewhere?

mminella:
You're right.  This has been updated.  Thanks!

PR #393, opened 2015-10-21
Update the tests to use the latest version of JRuby 1.x

wilkinsona:
Previously, the JRuby engine that was being used pulled in an old
version of JRuby. The engine is incompatible with more recent JRuby 1.x
releases which causes problems for the Spring IO Platform which uses
JRuby 1.7.x.

This commit updates the tests to use the latest version of JRuby 1.x
(1.7.22) which contains its own JSR 223 integration, i.e. there’s no
need for a separate script engine dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,7 +89,7 @@ allprojects {
 		xercesVersion = '2.8.1'
 		xmlunitVersion = '1.5'
 		xstreamVersion = '1.4.7'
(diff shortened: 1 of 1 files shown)


mminella:
LGTM.  Merged as 680eab20950fdfed27d7589d8f1824fdb0168ef8

PR #394, opened 2015-10-21
Make FlatFileItemReaderTests Spring Framework 4.2 compatible

wilkinsona:
Previously, FlatFileItemReaderTests made a precise assertion that
involved the toString output from ByteArrayResource. This output
has changed between Spring Framework 4.0.x and Spring Framework 4.2.x.
As a result, the test fails when run as part of the Spring IO Platform
build which forces the version of Spring Framework to 4.2.x.

This commit updates the failing test so that it passes against both
Spring Framework 4.0.x and Spring Framework 4.2.x by relaxing the
assertion a little so that it is not dependent on the precise output
of ByteArrayResource's toString method.


mminella:
LGTM Merged as 2f552b74493884a601d936515a29aa27853bf8ca.

PR #395, opened 2015-10-23
Update SmokeTests to be compatible with Spring Integration 4.2

wilkinsona:
Spring Integration 4.2 has fixed a bug where integration components
were being started too early. SmokeTests was relying on this bug to
start the test class which is also a @MessageEndpoint.

This commit updates the test to use a nested inner-class for the
endpoint and a bean declaration in the XML configuration. This allows
the test to pass against both Spring Integration 4.0 and 4.2.


garyrussell:
LGTM

mminella:
Merged as 2dfca3297873d59b3113d4596f3eab4c562d86b4.

PR #398, opened 2015-11-04
partitionJob.xml should demonstrate asynch process

guilhermerama:
The README file on samples describes the partitionJob as an example of the asynch process feature, so I changed the class of the taskExecutor to SimpleAsyncTaskExecutor


mminella:
Merged as c7ef097c251383b37c25baa9e88587b7831d4de6

PR #406, opened 2016-03-16
Don't use hamcrest-all.

stefanbirkner:
The library is considered to be used by people who download their dependencies manually. If you use a dependency-manager then you should use hamcrest-library. It has a dependency on hamcrest-core and does not need exclusion of hamcrest-core for other dependencies.


mminella:
Thanks for the update!  Merged.

PR #439, opened 2016-10-15
BATCH-2396 Guard Serializable check message

marschall:
The exception message for the Serializable check in
DefaultExecutionContextSerializer is always generated. For large
contexts this can quickly put a log of pressure on the allocator and
the heap.

This commit contains the following changes:
- generate the exception message for non-serializable values in
  DefaultExecutionContextSerializer only when needed

Issue: BATCH-2396


mminella:
LGTM.  Merged as f78aa275c045cf35168f7a9f081e7b3709f338ae.  Thanks for the contribution!

PR #443, opened 2016-12-20
Fix incorrect behaviour of RepositoryItemReader#jumpToItem() on restart

akapps:
`RepositoryItemReader` reads the latest already processed item when initialized with "jumpToItem" :
  - tests correction (confusion between indices and values)
  - jumpToItem(..) reimplemented to load the expected value on the next "read()" call

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -455,6 +455,7 @@ project('spring-batch-infrastructure-tests') {
 		testCompile("org.springframework:spring-oxm:$springVersion") {
 			exclude group: 'commons-lang', module: 'commons-lang'
 		}
+		testCompile "org.springframework.data:spring-data-jpa:$springDataJpaVersion"
 		testCompile "org.springframework:spring-jdbc:$springVersion"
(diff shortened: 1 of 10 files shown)


akapps:
Well, I focused on my problem and the change I provided is not quite satisfying _as is_, as it results in different behaviors whether the requested item is the first of a new page, or somewhere else in the page.

I should write some integration tests that highlight the Hibernate Sessions issue as well, and provide a more coherent solution.
I guess we could only prepare the cursor on `jumpToItem(..)`, and actually request the page only when `doRead()` is called.

akapps:
I completed and reorganized my PR.
A first commit corrects the unit tests with the expected behavior (see [BATCH-2526](https://jira.spring.io/browse/BATCH-2526])). I also added integration tests using entities with a lazy-loaded association. Tests are red with the current reader's implementation.
The 2nd commit adapts the reader.

benas:
LGTM, rebased and merged as cbdb251441968197aca805cdda4e634ab8542c39. Thank you for your contribution!

PR #445, opened 2017-01-04
BATCH-2565 : Test case to reproduce infinite loop and proposed update…

yge45:
… on FaultTolerantChunkProcessor

### DIFF
--- a/spring-batch-core-tests/src/main/resources/META-INF/batch/infiniteLoopOnRetryJob.xml
+++ b/spring-batch-core-tests/src/main/resources/META-INF/batch/infiniteLoopOnRetryJob.xml
@@ -1,0 +1,84 @@
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:energie="http://energie.aramice.fr/schema/web-config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:batch="http://www.springframework.org/schema/batch" xmlns:util="http://www.springframework.org/schema/util"
+	xsi:schemaLocation="
+		http://www.springframework.org/schema/beans	http://www.springframework.org/schema/beans/spring-beans.xsd
+		http://www.springframework.org/schema/batch http://www.springframework.org/schema/batch/spring-batch.xsd
+		http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd">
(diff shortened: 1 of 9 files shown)


pivotal-issuemaster:
@yge45 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=445)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=445) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yge45 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=445)!

yge45:
See https://jira.spring.io/browse/BATCH-2565 for detail on the problem

benas:
Hi @yge45 ,

As explained in JIRA, the code changes you suggest will actually fix the case but what will happen is:

1. cache the value the first time when the processor succeeds (not part of your PR)
2. remove the cached value the second time when scanning (undo step 1)
3. detect that through skip count and prevent scanning the (single skipped item) chunk

This is by design (of the transactional property of the processor) what we **don't** want to do. Either we cache the value and reuse it without recalling the processor, or we don't cache any value and retry calling the processor according to the retry policy you provide.

The issue is actually you are using a non idempotent processor with a fault tolerant step (which is not what the documentation suggests). I'm closing this issue. Thank you for your PR.

Kind regards
Mahmoud

yge45:
Hi Mahmoud

Thanks for the return. OK for your answer and your comment on the code changes I suggest (I must admit I have not understand all code in FaultTolerantChunkProcessor - too complicated for me).

I think we still have a problem. We should not go in a infinite loop in this case !!! From what I remember some items are not processed after first write error --> run state is KO and can't be restarted.
Shouldn't SB fail on getting an exception on an item processed a 2nd time (first time with success). This "NonIdempotentProcessingItemFailure" would prevent from getting in this infinite loop ?

To me (and I think it's the mots common use case ???) it's really hard to implement idempotent processor. In most case we can do it from business point of view but we can't be sure that a technical exception (SGBD or WS access) won't be thrown.   

Kind regards
Yoann

PR #446, opened 2017-01-06
Added AbstractCursorItemReader.connectionAutoCommit

gdarmont:
This PR is a follow up of discussion in PR https://github.com/spring-projects/spring-boot/pull/7866

The use case was to use a JdbcCursorItemReader with PostgreSQL (nothing fancy here).
From https://jdbc.postgresql.org/documentation/94/query.html, section "Getting results based on a cursor", the connection must not be in auto commit mode.
But JdbcCursorItemReader does not allow to explicity set connection mode to autoCommit = false. 

As discussed in the Spring Boot PR, configuring the JDBC pool to force all connection to autoCommit=false is not an optimal way to workaround the problem.

So here's a little PR that allows configuration of autocommit mode for the connection used by a CursorItemReader.

I validated the use of a Boolean wrapper against https://github.com/spring-projects/spring-framework/blob/master/spring-jdbc/src/main/java/org/springframework/jdbc/datasource/SingleConnectionDataSource.java#L65, to keep Spring philosophy.

Hope this helps,
Guillaume

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/AbstractCursorItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/AbstractCursorItemReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2013 the original author or authors.
+ * Copyright 2006-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -134,6 +134,9 @@ implements InitializingBean {
 
 	private boolean useSharedExtendedConnection = false;
 
+	private Boolean connectionAutoCommit;
+
+	private Boolean initialConnectionAutoCommit;
 
 	public AbstractCursorItemReader() {
 		super();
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/AbstractCursorItemReader.java
```diff
@@ -424,6 +444,9 @@ protected void initializeConnection() {
 			else {
 				this.con = dataSource.getConnection();
 			}
+			if (this.connectionAutoCommit != null && this.con.getAutoCommit() != this.connectionAutoCommit) {
+				this.con.setAutoCommit(this.connectionAutoCommit);
+			}
```
mminella: If we allow this, we need to set it back to what it was once we're done with it.

gdarmont: Thanks for the review.
I added the necessary code and amended the commit.

gdarmont:
Hi Spring Batch Team,

Do you think there is additional work to to on this PR ? I'd be happy to help.

Guillaume

mminella:
Rebased and merged as dca726af16b7145bb2950d1709c8eaaf77ede7a4.  Also created Jira issue BATCH-2661 for this.  Thanks!

PR #449, opened 2017-01-23
Override the version of Mockito used for Spring IO tests

wilkinsona:
Spring Boot 2.0 (and therefore Spring IO Platform Cairo) have upgraded
to Mockito 2.x which is not backwards compatible with 1.x. This breaks
a number of Spring Batch's tests when they're run with Mockito 2.x.

As Mockito is a test-only dependency, its version can be safely
overridden when running Spring IO compatibility tests without having
an adverse affect on compatibility for users. This commit overrides
the Platform's dependency management so that the Spring IO tests are
run using Mockito 1.x.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -241,6 +241,9 @@ configure(mainProjects) {
 				imports {
 					mavenBom "io.spring.platform:platform-bom:${platformVersion}"
 				}
+				dependencies {
+					dependency "org.mockito:mockito-core:$mockitoVersion"
+				}
 			}
 		}
 	}


mminella:
For 4.0, I'd rather just upgrade the version instead of override the version for the tests unless you have a reason not to...

wilkinsona:
Nope, that's fine with me. It'll be more work, but it's a better solution.

mminella:
I did the upgrade via commit 54a90beed7973875152b28e283a656e1f4f1759b so I'm closing this PR.  Thanks for pointing it out!

wilkinsona:
Great stuff. Thanks for the quick turnaround

PR #452, opened 2017-02-16
BATCH-2455 unifie listener ordering

Danny02:
Why:
Listeners should be called in normal order before some event
(i.e. read, write) and after that event in reverse order.

Side effects:
The ordering of Chunk and ItemReader Listeners is changed and will
change some behaviors of users. This change is still neccesary because
it is not only the correct and obviouse way, but also prevents new
users to hack around this problem.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/listener/CompositeChunkListener.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/listener/CompositeChunkListener.java
@@ -49,35 +49,40 @@ public class CompositeChunkListener implements ChunkListener {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


mminella:
Looks good to me.  I merged it onto master instead of the 3.x branch since it is a breaking change. 
 It will be available in Spring Batch 4.0.  Thanks for the contribution!

pivotal-issuemaster:
@Danny02 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=452)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=452) to manually synchronize the status of this Pull Request.

PR #453, opened 2017-02-18
BATCH-2394 JOIN Query with alias in SqlPagingQueryProviderFactoryBean not working in DB2

ssharma:
Front porting fix of https://github.com/spring-projects/spring-batch/pull/21 into 3.0.x line. Modified the fix to be compatible with 3.0.x code

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/SqlPagingQueryUtils.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/SqlPagingQueryUtils.java
@@ -246,7 +246,7 @@ public class SqlPagingQueryUtils {
 	 * @return a String that can be appended to an ORDER BY clause.
 	 */
 	public static String buildSortClause(AbstractSqlPagingQueryProvider provider) {
-		return buildSortClause(provider.getSortKeys());
+		return buildSortClause(provider.getSortKeysWithoutAliases());
 	}
 	
 	/**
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@ssharma Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=453)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=453) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ssharma Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=453)!

benas:
Hi @ssharma,

Thank you for this PR. I tested the changes and they don't fix the issue.

Here is the repo I used to reproduce the problem: https://github.com/benas/sandbox/tree/master/batch2394.

The example uses two tables `Person` and `Address` both having a column `name` (this is to reproduce the same issue as reported in [BATCH-2394](https://jira.spring.io/browse/BATCH-2394)). When I use a `sortKey` with an alias (like `p.name` to remove ambiguity between `Person.name` and `Address.name`), the change in the PR will remove the alias from the "ORDER BY" clause (as I was expecting) but the error is still happening.

It looks like the problem is specific to DB2 because the same query is working fine with MySQL.

Can you take a look?

Kr,
Mahmoud

benas:
Closing this PR as it does not fix the issue (see my previous comment). Thank you for your effort anyway!

PR #456, opened 2017-03-23
Fix typo with ExecutionContext class name

adchilds:
Fixes a typo in the Common Batch Patterns section of the reference.


mminella:
Thanks for pointing this out.  We fixed it in the new documentation.

PR #462, opened 2017-05-06
Align Dependency Versions with Spring

mdeinum:
Align dependencies of libraries to the versions as managed
by spring. The libraries concerned are
- Castor
- Hibernate
- Hibernate Validator
- Groovy
- SLF4J

For other libraries (like Jackson, and javax.mail) there isn't a final 
version yet and hence aren't updated just yet.


mminella:
Thanks for the update.  Merged as d512fef644858bd1917e1f549c50fc088368da55

PR #465, opened 2017-05-08
Added CompositeItemWriterBuilder.

cppwfs:
resolves BATCH-2597

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/CompositeItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/CompositeItemWriterBuilder.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.item.support.builder;
+
+import java.util.List;
+
+import org.springframework.batch.item.ItemWriter;
+import org.springframework.batch.item.support.CompositeItemWriter;
+import org.springframework.util.Assert;
+
+/**
+ * Creates a fully qualified CompositeItemWriter.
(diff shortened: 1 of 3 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/CompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+	 *
+	 * @param delegates the list of delegates to use. The delegates list must not be null
+	 * nor be empty.
+	 * @return this instance for method chaining.
+	 * 
+	 * @see CompositeItemWriter#setDelegates(List)
+	 */
+	public CompositeItemWriterBuilder delegates(List<ItemWriter<? super T>> delegates) {
+		Assert.notNull(delegates, "A list of delegates is required.");
+		Assert.notEmpty(delegates, "The delegates list must have one or more delegates.");
```
mminella: I've been doing all asserts in the build method to prevent the duplication you have here (for the null check).  If the user passes in null, that's ok as long as they don't call build until they call it with something else...

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/CompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+	 * item writer delegates associated with the CompositeItemWriter.
+	 *
+	 * @param ignoreItemStream if false the delegates' open, close, or update methods will
+	 * be called when the corresponding methods on the CompositeItemWriter are called. If
+	 * true the delegates' open, close, nor update methods will not be called (default is false).
+	 * @return this instance for method chaining.
+	 * 
+	 * @see CompositeItemWriter#setIgnoreItemStream(boolean)
+	 */
+	public CompositeItemWriterBuilder ignoreItemStream(boolean ignoreItemStream) {
```
mminella: Missing generic on return.  I'll fix in merge

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/CompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+	 * The list of item writers to use as delegates. Items are written to each of the
+	 * delegates.
+	 *
+	 * @param delegates the list of delegates to use. The delegates list must not be null
+	 * nor be empty.
+	 * @return this instance for method chaining.
+	 * 
+	 * @see CompositeItemWriter#setDelegates(List)
+	 */
+	public CompositeItemWriterBuilder delegates(List<ItemWriter<? super T>> delegates) {
```
mminella: Same thing here on the generic.

mminella:
Merged with minor polish as 75566ba8109511c68bec904f4f585682237dd2fd

PR #466, opened 2017-05-09
Adds builder for CompositeItemStream.

cppwfs:
resolves #BATCH-2598


cppwfs:
Since this is used internally by batch a builder is not necessary.

PR #467, opened 2017-05-09
Add Builder for ClassifierCompositeItemWriter

cppwfs:
resolves BATCH-2599

Added null check for setClassifier in ClassifierCompositeItemWriter.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
@@ -41,7 +41,7 @@ public class ClassifierCompositeItemWriterBuilder<T> {
 	 * @return this instance for method chaining
 	 * @see org.springframework.batch.item.support.ClassifierCompositeItemWriter#setClassifier(Classifier)
 	 */
-	public ClassifierCompositeItemWriterBuilder classifier(Classifier<T, ItemWriter<? super T>> classifier) {
+	public ClassifierCompositeItemWriterBuilder<T> classifier(Classifier<T, ItemWriter<? super T>> classifier) {
 		Assert.notNull(classifier, "A classifier is required.");
 		this.classifier = classifier;
 
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+	/**
+	 * Establish the classifier to be used for the selection of which {@link ItemWriter}
+	 * to use.
+	 *
+	 * @param classifier the classifier to set
+	 * @return this instance for method chaining
+	 * @see org.springframework.batch.item.support.ClassifierCompositeItemWriter#setClassifier(Classifier)
+	 */
+	public ClassifierCompositeItemWriterBuilder classifier(Classifier<T, ItemWriter<? super T>> classifier) {
```
mminella: Missing generic on return

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Creates a fully qualified ClassifierCompositeItemWriter.
+ *
+ * @author Glenn Renfro
+ *
+ * @since 4.0
+ */
+public class ClassifierCompositeItemWriterBuilder<T> {
+
+	private Classifier<T, ItemWriter<? super T>> classifier = new ClassifierSupport<T, ItemWriter<? super T>>(null);
```
mminella: Why are we providing a non-valid default?

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Establish the classifier to be used for the selection of which {@link ItemWriter}
+	 * to use.
+	 *
+	 * @param classifier the classifier to set
+	 * @return this instance for method chaining
+	 * @see org.springframework.batch.item.support.ClassifierCompositeItemWriter#setClassifier(Classifier)
+	 */
+	public ClassifierCompositeItemWriterBuilder<T> classifier(Classifier<T, ItemWriter<? super T>> classifier) {
+		Assert.notNull(classifier, "A classifier is required.");
```
mminella: Move this assert into the build.  As it sits right now, you can end up with an invalid writer (one with the invalid default) by just doing:
```
new ClassifierCompositeItemWriter().build();
```
A test case for this condition should also be added.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
cppwfs: Good point

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/builder/ClassifierCompositeItemWriterBuilder.java
cppwfs: done

mminella:
LGTM.  Squashed and merged as 70a1a83a9a7a43fdd7e2a3a199f7c989a825d18a

PR #470, opened 2017-05-10
Add builder for JmsItemReader

cppwfs:
resolves BATCH-2602

### DIFF
--- a/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
+++ b/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.item.jms.builder;
+
+import java.util.Date;
+
+import javax.jms.Message;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.springframework.batch.item.jms.JmsItemReader;
+import org.springframework.jms.core.JmsOperations;
+import org.springframework.jms.core.JmsTemplate;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Glenn Renfro
+ */
+public class JmsItemReaderBuilderTests {
+
+	private JmsOperations defaultJmsTemplate;
+
+	@Before
+	public void setupJmsTemplate() {
+		this.defaultJmsTemplate = mock(JmsOperations.class);
+		when(this.defaultJmsTemplate.receiveAndConvert()).thenReturn("foo");
+	}
+
+	@Test
+	public void testNoItemType() {
+		JmsItemReader<String> itemReader = new JmsItemReaderBuilder<String>().jmsTemplate(this.defaultJmsTemplate)
+				.build();
+		assertEquals("foo", itemReader.read());
+	}
+
+	@Test
+	public void testSetItemType() {
+		JmsItemReader<String> itemReader = new JmsItemReaderBuilder<String>().jmsTemplate(this.defaultJmsTemplate)
+				.itemType(String.class).build();
+
+		assertEquals("foo", itemReader.read());
+	}
+
+	@Test
+	public void testSetItemSubclassType() {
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
```diff
@@ -0,0 +1,127 @@
...
+
+	@Test
+	public void testNoItemType() {
+		JmsItemReader<String> itemReader = new JmsItemReaderBuilder<String>().jmsTemplate(this.defaultJmsTemplate)
+				.build();
+		assertEquals("foo", itemReader.read());
+	}
+
+	@Test
+	public void testSetItemType() {
```
mminella: Does this test actually validate the condition?  Since the behavior is the same as the previous test, I'd assume that you'd need to set a type other than String to validate that this worked...

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
```diff
@@ -0,0 +1,127 @@
...
+			fail("Expected IllegalStateException");
+		}
+		catch (IllegalStateException e) {
+			// expected
+			assertTrue(e.getMessage().indexOf("wrong type") >= 0);
+		}
+	}
+
+	@Test
+	public void testNextMessage() {
```
mminella: What is the point of this test?

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
```diff
@@ -0,0 +1,127 @@
...
+		Message message = mock(Message.class);
+		when(jmsTemplate.receive()).thenReturn(message);
+
+		JmsItemReader<Message> itemReader = new JmsItemReaderBuilder<Message>().jmsTemplate(jmsTemplate)
+				.itemType(Message.class).build();
+		assertEquals(message, itemReader.read());
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void testTemplateWithNoDefaultDestination() {
```
mminella: We don't need this test for the builder since it's not testing the builder's functionality.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
```diff
@@ -0,0 +1,127 @@
...
+		JmsTemplate jmsTemplate = new JmsTemplate();
+		jmsTemplate.setReceiveTimeout(100L);
+		new JmsItemReaderBuilder<String>().jmsTemplate(jmsTemplate).itemType(String.class).build();
+
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void testTemplateWithNoTimeout() {
```
mminella: We don't need this test for the builder since it's not testing the builder's functionality.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
cppwfs: To send a message versus a string.  I can rename it

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
cppwfs: Its not.  This test sets the itemType where the previous test did not.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
cppwfs: But shouldn't the test still be here to verify the existing behavior is the same?

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
cppwfs: But shouldn't the test still be here to verify the existing behavior is the same?

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
mminella: But the behavior is the same, isn't it?  If I comment out line 77 in the builder, does this test still pass?  To be a valid test, it shouldn't.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
mminella: We aren't testing the behavior of the reader itself, we're testing the behavior of the builder.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/jms/builder/JmsItemReaderBuilderTests.java
mminella: We aren't testing the behavior of the reader itself, we're testing the behavior of the builder.

mminella:
LGTM.  Squashed and merged as 3850e4c2fe61a691d914bd35e62f036bb26e47c5.

PR #475, opened 2017-05-15
Add Builder for RepositoryItemReader

cppwfs:
resolves BATCH-2606

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.item.data.builder;
+
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.batch.item.ExecutionContext;
+import org.springframework.batch.item.data.RepositoryItemReader;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * A builder implementation for the {@link RepositoryItemReader}.
+ *
+ * @author Glenn Renfro
+ * @since 4.0
+ * @see RepositoryItemReader
+ */
+
+public class RepositoryItemReaderBuilder<T> {
+
+	private PagingAndSortingRepository<?, ?> repository;
+
+	private Map<String, Sort.Direction> sorts;
+
+	private List<?> arguments;
+
+	private int pageSize = 10;
+
+	private String methodName;
+
+	private int currentItemCount;
+
+	private int maxItemCount;
+
+	private boolean saveState = true;
+
+	private String name;
+
+	/**
+	 * Arguments to be passed to the data providing method.
+	 *
+	 * @param arguments list of method arguments to be passed to the repository.
+	 * @return The current instance of the builder.
+	 * @see RepositoryItemReader#setArguments(List)
+	 */
+	public RepositoryItemReaderBuilder<T> arguments(List<?> arguments) {
+		this.arguments = arguments;
+
+		return this;
+	}
+
+	/**
+	 * Provides ordering of the results so that order is maintained between paged queries.
+	 *
+	 * @param sorts the fields to sort by and the directions.
+	 * @return The current instance of the builder.
+	 * @see RepositoryItemReader#setSort(Map)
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+	 *
+	 * @param methodName name of the method to invoke.
+	 * @return The current instance of the builder.
+	 * @see RepositoryItemReader#setMethodName(String)
+	 */
+	public RepositoryItemReaderBuilder<T> methodName(String methodName) {
+		this.methodName = methodName;
+
+		return this;
+	}
```
mminella: Is there a way to add a type safe way of doing this?  I've had complaints in the past that users have avoided using this reader because they'd rather write their own and get type safety between the reader and the repository.  I think we need to explore a bit to see if there's an easy way to add a type safe way to configure this.  Maybe something like:
```
fooRepositoryBuilder.name("foo")
    .repository(myFooRepository)
    .method(myFooRepository.findAll())  // I'm thinking about how when works in Mockito.  Also, I'm not sure how messy this would get with arguments but worth a look
    .build();
```
For the record, I'm not advocating replacing this method.  This would be either you configure it via the string or you configure it this other way, both options would be there.

cppwfs: Okeedokee.  I'll take a shot it ;-)

cppwfs: PR is updated allowing the user to pass in a method.  But, after updating,  I had second thoughts as to whether my solution meets the criteria of type safety.  Please take a peak though.

cppwfs:
Yeah we need to talk about this then.  I'll remove the commit.  Do we want to open a separate story.  That feature seems to exceed the purpose of this story

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
```diff
@@ -0,0 +1,214 @@
...
+	 *
+	 * @param repositoryReference of the used to get a repository and type-safe method for
+	 * use by the reader.
+	 * @return The current instance of the builder.
+	 * @see RepositoryItemReader#setMethodName(String)
+	 * @see RepositoryItemReader#setRepository(PagingAndSortingRepository)
+	 *
+	 */
+	public RepositoryItemReaderBuilder<T> repository(RepositoryMethodReference repositoryReference) {
+		Assert.notNull(repositoryReference, "repositoryReference must not be null.");
```
mminella: For these builders the Asserts should be in the #build() method.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
```diff
@@ -0,0 +1,214 @@
...
+		private RepositoryMethodIterceptor repositoryInvocationHandler;
+
+		private PagingAndSortingRepository<?, ?> repository;
+
+		public RepositoryMethodReference(PagingAndSortingRepository<?, ?> repository) {
+			this.repository = repository;
+			this.repositoryInvocationHandler = new RepositoryMethodIterceptor();
+		}
+
+		// T is a proxy of the object passed in in the constructor
```
mminella: We should be clear here in the javadoc that the returned proxy prevents actual method execution and should be used beyond this specific use case.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
```diff
@@ -0,0 +1,214 @@
...
+		reader.setSort(this.sorts);
+		reader.setName(this.name);
+		return reader;
+	}
+
+	/**
+	 * Establishes a proxy that will capture a the Repository and the associated
+	 * methodName that will be used by the reader.
+	 * @param <T> The type of repository that will be used by the reader.
+	 */
```
mminella: We should note that this method can only be used with non-final classes.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilder.java
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+	}
+
+	private static class RepositoryMethodIterceptor implements MethodInterceptor {
+		private String methodName;
+
+		@Override
+		public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable {
+			this.methodName = method.getName();
```
mminella: Can't we save off the arguments here as well so they don't even need to configure the arguments call?

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/data/builder/RepositoryItemReaderBuilderTests.java
```diff
@@ -0,0 +1,241 @@
...
+	private Map<String, Sort.Direction> sorts;
+
+	private List<String> testResult;
+
+	private ArgumentCaptor<PageRequest> pageRequestContainer;
+
+	@Before
+	public void setUp() throws Exception {
+		MockitoAnnotations.initMocks(this);
+		this.sorts = new HashMap<String, Sort.Direction>();
```
mminella: Spring Batch is java 8 now, please be sure to use diamond operators where appropriate.

mminella:
LGTM.  Squashed, polished, and merged.

PR #477, opened 2017-05-16
Added Builder for RepositoryItemWriter

cppwfs:
resolves BATCH-2609

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemWriterBuilder.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.item.data.builder;
+
+import org.apache.shiro.util.Assert;
+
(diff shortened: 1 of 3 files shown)


mminella:
Can we get this PR updated to provide the same type safe method configuration we did on the reader?

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemWriterBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.item.data.builder;
+
+import java.lang.reflect.Method;
+
+import org.apache.shiro.util.Assert;
```
mminella: Wrong import

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/builder/RepositoryItemWriterBuilder.java
```diff
@@ -0,0 +1,162 @@
...
+
+	}
+
+	private static class RepositoryMethodIterceptor implements MethodInterceptor {
+		private String methodName;
+
+		@Override
+		public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable {
+			this.methodName = method.getName();
+			return null;
```
mminella: We need to add a test where the repository returns a primitive (`int` for example).  I'm wondering if this will break.  If it does, we need to document that limitation.

cppwfs: Hmm... Not sure I understand.  (Argue with me) 
Java does not support generics with a primitive type.
For example will not compile.:
```
RepositoryItemWriter<int> writer = new RepositoryItemWriterBuilder<int>().methodName("save")
	.repository(this.repository).build();
```

mminella: Do Spring Data Repositories support primitives?  If they do, we'll need to be able to handle them (or document that we don't support them).

cppwfs:
Closing this PR  and opening https://github.com/spring-projects/spring-batch/pull/489.

PR #480, opened 2017-05-17
Added builder for MongoItemReader

cppwfs:


spring-batch-infrastructure/src/test/java/org/springframework/batch/item/data/builder/MongoItemReaderBuilderTests.java
```diff
@@ -0,0 +1,166 @@
...
+
+		assertNull("reader should not return result", reader.read());
+
+		Query query = this.queryContainer.getValue();
+		assertEquals("{ $natural : 1}", query.getHint());
+	}
+
+	@Test
+	public void testCollection() throws Exception {
+		MongoItemReader<String> reader = getBasicBuilder().parameterValues(new ArrayList<Object>() {
```
mminella: Small nit, `Collections.asSingletonList("foo")` is probably a better way to do this instead of creating an anonymous inner class with a static initializer (I used to do this but have moved away from it).

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/data/builder/MongoItemReaderBuilderTests.java
```diff
@@ -0,0 +1,166 @@
...
+		Query query = this.queryContainer.getValue();
+		assertEquals("{ \"name\" : \"foo\" }", query.getQueryObject().toJson());
+		assertEquals("{ \"name\" : -1 }", query.getSortObject().toJson());
+		assertEquals("collection", collectionContainer.getValue());
+	}
+
+	@Test
+	public void testNullTemplate() {
+		validateExceptionMessage(new MongoItemReaderBuilder<String>().targetType(String.class).query("{ }")
+				.sorts(this.sortOptions).name("mongoReaderTest").pageSize(50), "template is required.");
```
mminella: Formatting for all of the builder expressions in this class.  Will fix on merge.

mminella:
Merged as fa977333febfdaa0ae9b03b8f393c3a0f0a55101.

PR #488, opened 2017-06-01
spike for migrating documentation from docbook to asciidoctor

cppwfs:
Resolves BATCH-2620

First pass at migrating Reference Docs from Docbook to AsciiDoctor.

1) html based docs look good
2) pdf based docs look good, but links in index do not work.  Need to discuss
3) Formatting of raw adocs needs to be cleaned up (kept format from migration tool). 
4) To build ./gradlew asciidoctor  (results put in build/asciidoctor directory)


cppwfs:
In some segments of the document it refers to ` will be wired together via an XML configuration file`.  Do we still  want to maintain the "XML" statement?

mminella:
Until we update the docs to be "java config first", yes it should stay the same.

mminella:
LGTM.  Merged as 67853f22de6a509e0efbfe2232f02a1e5d346489

PR #491, opened 2017-06-22
Align XML and Java based configuration

mdeinum:
This commit aligns the XML and Java based validations.

When using XML to configure a chunk oriented step both an ItemReader and
ItemWriter are requied. However when using Java based configuration the
ItemWriter is optional if an ItemProcessor is present.

Having no ItemWriter and only an ItemProcessor lead to strange results
in one of our batch jobs, which was accidentily configured without an
ItemProcessor.

Related: BATCH-1520
Fixes: BATCH-2624


benas:
@mdeinum Could you please rebase your PR on the latest master branch and update the year to 2018 in the license header?

Could you also add a test that expects an `IllegalStateException` when the writer is null? This test is necessary to validate that the item writer is mandatory.

Thanks in advance.

benas:
#593 adds the missing test **on top** of this PR. All the credit goes to @mdeinum for the fix 👍. 

This can be closed if #593 is merged.

benas:
LGTM. rebased and merged as 3ccc4ff430421d32cbdcf7a196cf03a83b164676 .

PR #495, opened 2017-07-12
BATCH-2630 - Infinite loop on job restart when step was repeated

cameronhurd:
Bug fix:  use step execution ID as tie breaker when start time is identical

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java
@@ -229,7 +229,9 @@ public class SimpleJobRepository implements JobRepository {
 			if (latest == null) {
 				latest = stepExecution;
 			}
-			if (latest.getStartTime().getTime() < stepExecution.getStartTime().getTime()) {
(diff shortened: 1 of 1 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java
```diff
@@ -229,7 +229,9 @@ public StepExecution getLastStepExecution(JobInstance jobInstance, String stepNa
 			if (latest == null) {
 				latest = stepExecution;
 			}
-			if (latest.getStartTime().getTime() < stepExecution.getStartTime().getTime()) {
+			// Ordered by step execution ID so less than or equal makes last step execution ID 
+			// the tie breaker if start time is identical
+			if (latest.getStartTime().getTime() <= stepExecution.getStartTime().getTime()) {
```
mminella: This doesn't actually make the guarantee we need.  The `StepExecutionDao` interface doesn't guarantee order and could be re-implemented for another reason, leaving off the `order by`.  This really should check the `id` to be safe.

cameronhurd: Valid point... fixed

PR #498, opened 2017-07-23
Fixed typos in unit testing docs.

TaekSoonJang:
# Obvious Fix

Source code in [10.2](https://docs.spring.io/spring-batch/trunk/reference/html/testing.html#endToEndTesting) does not work for two reasons:

- jobLauncherTestUtils.launchJob().getStatus() does not return JobExecution.
- jobExecution.getExitStatus() does not return String.

Fixed the codes to compile and pass the test.


mminella:
Thanks for the fix!  Rebased and merged as 110b95ee1e205ef06a39ccc0853e462ec67412db.

PR #501, opened 2017-08-04
Updated the document to contain real Java Config

cppwfs:
Also updated the javascript to dynamically count the segments that need be updated vs. hardcoded


spring-batch-docs/asciidoc/domain.adoc
```diff
@@ -70,6 +115,44 @@ functionality on top of `Job`, however the batch
 namespace abstracts away the need to instantiate it directly. Instead, the
 `<job>` tag can be used:
```
mminella: The text needs to be updated as well since it talks about the `<job` tag and yet we show java config.

spring-batch-docs/asciidoc/domain.adoc
```diff
@@ -70,6 +115,44 @@ functionality on top of `Job`, however the batch
...
+<div id="java_section_1">
++++
+[source, java]
+----
+jobBuilderFactory.get("footballJob")
+    .flow(playerload())
+    .next(gameLoad())
+    .next(playerSummarization())
+    .end()
+    .build();
```
mminella: This should be the whole bean factory method:
```
@Bean
public Job footballJob() {
    return this.jobBuilderFactory.get("footballJob")
                     .start(playerLoad())
                     .next(gameLoad())
                     .next(playerSummarization())
                     .end()
                     .build();
}

File: spring-batch-docs/asciidoc/domain.adoc [hunk 3]
```diff
@@ -609,11 +706,50 @@ during the course of execution `StepExecution` and
 `JobExecution` implementations are persisted by
 passing them to the repository:
```
mminella: This description needs to be updated for this.  The XML snip shows how to define a new `JobRepository`.  The java snip shows how to obtain an instance that has already been configured.  They do two different things.

spring-batch-docs/asciidoc/domain.adoc
```diff
@@ -665,7 +801,7 @@ not be written out. More details about the `ItemProcessor` interface can be
 found in <<readersAndWriters.adoc#readersAndWriters,Readers And Writers>>.
 
 
-=== Batch Namespace
+=== Batch Namespace When Using XML Configuration
```
mminella: I'm wondering if we should just hide this when the java config toggle is flipped...

cppwfs: I like the idea.  But the header is not rendered properly when placed in the ifdef block.  It loses its formatting :-(

cppwfs:
Also take a peak at the generated PDF's and tell me what you think.

spring-batch-docs/asciidoc/domain.adoc
mminella: I mean the whole section (not just the header.

cppwfs:
Understood.  When the user selects XML, the section is rendered but the header is not rendered properly.  When Java is selected nothing shows.    On Monday I'll revisit,  but for now I was leaving it as is.

mminella:
With regards to the generated PDF, I think that illustrates even more strongly that each section that will contain java vs XML config will need to be able to stand on it's own.  Seeing just the extra code block looks out of place.  I think we need to be sure that the text associated with the code block works as well.

cppwfs:
This PR was a first pass at what we could do with a slider control.   So closing this until solution can be figured out on tag management.

PR #505, opened 2017-08-25
Fix for BATCH-2636

kingbbode:
Fix for BATCH-2636 : In the DefaultBatchConfigurer, autowiring the setDataSource Method makes the dataSource injection of the public construct meaningless.

The null check avoids injected DataSource due to AutowiredAnnotationBeanPostProcessor.


mminella:
Thanks for the PR.  This is addressed via e5fe6a5de62e251e26152b5319196dd1997925cd

PR #506, opened 2017-08-29
Remove first set of JavaDoc warnings

cppwfs:


mminella:
Polished and committed.

PR #508, opened 2017-09-14
Add an `ItemProcessor` that delegates to a `java.util.function.Function`

mminella:
This commit adds a new `ItemProcessor` that delegates to a
`java.util.function.Function`.  It also enables users to configure a
`Function` as an `ItemProcessor` via the builders (so they don't need to
worry about the `ItemProcessor` wrapper in the first place).

Resolves BATCH-2641


mminella:
Merged

PR #510, opened 2017-09-25
JSR-352 Doc Updates

mminella:
This commit refactors the JSR-352 documention to utilize the new java vs
xml toggle.

### DIFF
--- a/spring-batch-docs/asciidoc/jsr-352.adoc
+++ b/spring-batch-docs/asciidoc/jsr-352.adoc
@@ -6,6 +6,8 @@
 
 == JSR-352 Support
 
+include::toggle.adoc[]
+
 As of Spring Batch 3.0 support for JSR-352 has been fully implemented. This section is not a replacement for
         the spec itself and instead, intends to explain how the JSR-352 specific concepts apply to Spring Batch.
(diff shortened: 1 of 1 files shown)


spring-batch-docs/asciidoc/jsr-352.adoc
```diff
@@ -204,7 +206,28 @@ To use Spring dependency injection within a JSR-352 based batch job consists of
...
+       }
+}
+
+
+<?xml version="1.0" encoding="UTF-8"?>
+<job id="fooJob" xmlns="http://xmlns.jcp.org/xml/ns/javaee" version="1.0">
+    <step id="step1" >
+        <batchlet ref="fooBatchlet" />
+    </step>
+</job>
```
cppwfs: We probably want to have a ifdef block for html with what is in the PR.  Then add  a ifdef block for pdf that will enumerate that they can use either java config or XML to configure the fooBatchlet.   I can do this on merge if you agree.

mminella: Let me update...

cppwfs:
LGTM!

cppwfs:
Merged Squashed.

PR #513, opened 2017-10-02
Editing pass

Buzzardo:
I did my "English teacher" thing and made sure that we don't imply the existing of objects called itemreaders and itemreaders.


mminella:
LGTM.  Squashed and merged as 83a468efa65cdca6651bcbd53c30f84f9ffc820c

PR #529, opened 2017-10-09
Editing pass for glossary.adoc

Buzzardo:
I fixed sentence errors. Very simple document (as it should be ), so it can't go wrong in very many ways.

(I don't know why the last line disappeared in Github. It's present in my local adoc file, and I didn't remove it. Also, when I add another line and save, Atom removes the extra line. I don't know why I sometimes get no blank line at the end.)


mminella:
LGTM.  Merged as 25d3b3070472159caac8fc1513bdb69c21778aa8

PR #532, opened 2017-10-09
90-character formatting

Buzzardo:
Other than a couple of image references, I fixed all the lines so that they would be no longer than 90 characters and removed unneeded blank lines.

I did find a few other things (mostly periods missing from list items), and I fixed them.


mminella:
LGTM. Rebased and merged as d0238c969951a75b490768c97c0ddf11bb95f71f

PR #533, opened 2017-10-09
90-character formatting

Buzzardo:
I adjusted all the lines so that they would be no longer than 90 characters. I also removed unneeded blank lines.


cppwfs:
LGTM

cppwfs:
Rebased Squashed Merged.

PR #535, opened 2017-10-13
90-character pass for repeat.adoc

Buzzardo:
I removed all the extraneous characters and line breaks from non-code lines and fitted all the non-code lines to 90 characters.


cppwfs:
Rebased, polished, squashed merged.

PR #536, opened 2017-10-13
90-character pass for scalability.adoc

Buzzardo:
I removed extraneous characters from non-code lines and arranged the non-code lines to fit into 90 characters.


mminella:
LGTM.  Merged as 0f989d9c9f1bc737d3ce84f01be5c19196d8b232

PR #537, opened 2017-10-16
90-character pass for domain.adoc

Buzzardo:
I removed extraneous characters from non-code lines and made each code line be as close to 90 characters as possible. I caught a few items I missed in the editing pass (such as missing hyphen and code ticks around some code words.


mminella:
LGTM.  Merged as c83d9b4557c2fa0d8dd16f57c7bbd2be32ab7470

PR #538, opened 2017-10-18
First draft of a new topic about Spring Boot and Spring Batch

Buzzardo:
With information that I've gotten from Michael, I've built a new topic to steer readers around the worst pitfalls that come with using Spring Batch inside of a Spring Boot application.

I expect a few rounds of modifications, so please hold off on merging this commit.

### DIFF
--- a/spring-batch-docs/asciidoc/springBootSpringBatch.adoc
+++ b/spring-batch-docs/asciidoc/springBootSpringBatch.adoc
@@ -1,0 +1,61 @@
+== Using Spring Batch with Spring Boot
+
+A number of things work differently when you use Spring Batch from within a Spring Boot
+application. This chapter describes the following differences:
+
+* <<runningCommandLine>>
+* <<storingJobRepository>>
+
+[[runningCommandLine]]
+=== Running from the Command Line
+
+Running Spring Batch with Spring Boot from the command line differs from running Spring
+Batch by itself from the command line. Spring Boot uses a class called
+`JobLaunchingCommandLineRunner`. Spring Batch uses a class called `CommandLineJobRunner`.
+They work a bit differently. However, for the most part, the differences do not cause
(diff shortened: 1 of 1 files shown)


benas:
I was reading the spring boot reference documentation and noticed that there are different sections that show the integration with other projects: [Spring Integration](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-integration), [Spring Session](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-session), etc.

So I think this PR is better suited to be part of boot's documentation. What do you think?

Buzzardo:
The idea is to help Batch users avoid a problem. Because the audience is Batch users, it makes more sense for this topic to be in the Batch guide.

benas:
> Because the audience is Batch users,

For me the audience is rather Boot users, according to the title of the section `Using Spring Batch with Spring Boot`. Not all batch users are boot users. I remember giving a talk a few months ago and 80% of the audience was non-boot users (probably because for many enterprises old batch jobs (at least those older than boot) are less likely to be changed or modernized to boot once they work well in production).

> The idea is to help Batch users avoid a problem

I don't see those as problems, they are boot specific details (how boot passes parameters, how boot auto-configures a datasource, etc) that a boot user should be aware of when developing a Spring Batch job. So IMO those should go in a "Spring Batch" section in boot docs (like sections for other projects mentioned above).

Buzzardo:
I recall that @mminella requested this addition, so I guess you should take it up with him. Perhaps things have changed in the intervening 14 months.

snicoll:
FTR, I'd welcome a better coverage of Spring Batch in the Spring Boot reference guide. Spring Boot users are likely to look there first, especially given the nature of the document in this PR.

benas:
I agree with @snicoll.

@Buzzardo please open this PR against boot's docs. I can help reviewing it if necessary.

PR #539, opened 2017-10-18
90-character pass for step.adoc

Buzzardo:
I removed extraneous characters in non-code lines and adjusted the lengths of non-code lines to be as close to but less than 90 characters as possible.

I did find and fix a number of issues that I missed the first time.


mminella:
I'm going to close this without merging.  This is the file I'm currently working on with the java config.  You'll want to do this change once I'm done.

PR #540, opened 2017-10-19
90-character pass for common-patterns.adoc

Buzzardo:
I removed all the extraneous white space from non-code lines and broke the non-code lines as close to 90 characters as possible. I caught one keyword that wasn't formatted and fixed that.


cppwfs:
LGTM

cppwfs:
Rebased Merged

PR #558, opened 2017-12-28
PullRequest for issue BATCH-2666

ThoNill:
In the package-info.java of the package org.springframework.batch.item.xmlpathreader
is a description of the structure of the packages und for the use of the api. 

The reason for this project was the need to read a complicated XML structure into a 
set of objects in a ItemReader. I generalized this to the task of extending the naming attribute of the annotation XmlElement to a path of XML elements.


mminella:
I'm closing this PR based on the comment left in the related Jira issue.  If you'd like to continue this path, let's come up with an architecture we can agree upon in Jira before opening up another PR.

PR #562, opened 2018-01-08
BATCH-2662: Fix syntax highlighting of Java example

benas:
This PR fixes the syntax highlighting of the Java example (which currently uses XML code colors) in section 1.1.1 of the documentation as reported in [BATCH-2662](https://jira.spring.io/browse/BATCH-2662)


cppwfs:
LGTM!

PR #564, opened 2018-01-08
Various changes to support a better document toggle

Buzzardo:
I got a simpler document toggle to work with the dynamic ToC, and I added cookies to remember the doc toggle setting for our readers. I started from scratch, with an eye specifically toward simplifying the HTML and JavaScript and doing more with CSS. I have tested it on Chrome, Firefox, Safari, and Opera. I'll try it on IE/Edge after this commit (when I can hit it from my Windows laptop).


mminella:
First let me say that I love the work.  The buttons are nice and I love that the ToC will be able to be consistent across the Spring portfolio projects.  Great work!  That being said, my only issue is with the both option.  We currently have verbiage for when both are displayed for the PDF use case (since we can't toggle there).  I don't want the both use case to be forgotten so I'd like to see either this PR remove the both button for now, or fix the verbiage for the both use case as part of this PR.  If we remove the both button now, we can add it back once we do fix the verbiage for that use case.

mminella:
Squashed, rebased, and merged as 0d9f4ce6f143bab4efd15b17d5bdb4b032c767cd.

PR #565, opened 2018-01-15
BATCH-2663: re-initialize the filter count when scanning a chunk

benas:
This PR is a possible fix for [BATCH-2663](https://jira.spring.io/browse/BATCH-2663).

Currently, the filter count of the step contribution is applied for each item of a scanned chunk. For example, with a chunk of 30, if 10 items are filtered and one item is skipped during write, then the filter count is equal to 210 (10 + 20 * 10).

The idea of the fix is to re-initialize the filter count when scanning the chunk.

Note that as of version 4.0.0.RELEASE, the filter count is applied for each retry attempt. This PR does **not** alter this behaviour as it is out of scope the current issue.

### DIFF
--- a/spring-batch-core-tests/src/test/java/org/springframework/batch/core/test/step/FaultTolerantStepIntegrationTests.java
+++ b/spring-batch-core-tests/src/test/java/org/springframework/batch/core/test/step/FaultTolerantStepIntegrationTests.java
@@ -1,0 +1,169 @@
+package org.springframework.batch.core.test.step;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.Step;
+import org.springframework.batch.core.StepExecution;
+import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
+import org.springframework.batch.core.repository.JobRepository;
+import org.springframework.batch.core.step.builder.FaultTolerantStepBuilder;
+import org.springframework.batch.core.step.skip.SkipLimitExceededException;
+import org.springframework.batch.core.step.skip.SkipPolicy;
+import org.springframework.batch.item.ItemProcessor;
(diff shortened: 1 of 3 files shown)


spring-batch-core-tests/src/test/java/org/springframework/batch/core/test/step/FaultTolerantStepIntegrationTests.java
```diff
@@ -0,0 +1,169 @@
+package org.springframework.batch.core.test.step;
```
mminella: While not as big of a deal on tests, I recommend having your IDE generate the license at the top of any java file you create.

spring-batch-core/src/main/java/org/springframework/batch/core/step/item/FaultTolerantChunkProcessor.java
```diff
@@ -149,6 +149,12 @@ protected void initializeUserData(Chunk<I> inputs) {
 			inputs.setUserData(data);
 			data.setOutputs(new Chunk<O>());
 		}
+		else {
```
mminella: Along the same lines as the license comment above, when we work on files with a license in the header, please be sure to update the date in the license to the current year.  Also, don't be afraid to add your author tag as well.

mminella:
Rebased and merged as c78701a871110b12e1f59026234baee956295f2e.

spring-batch-core-tests/src/test/java/org/springframework/batch/core/test/step/FaultTolerantStepIntegrationTests.java
benas: ok, I've updated my IDE for that.

spring-batch-core/src/main/java/org/springframework/batch/core/step/item/FaultTolerantChunkProcessor.java
benas: Sure! I will do it next time. Thanks

PR #567, opened 2018-01-24
BATCH-2397: fix parameters handling in MethodInvokingTaskletAdapter

benas:
This PR contains a fix for [BATCH-2397](https://jira.spring.io/browse/BATCH-2397).

When a tasklet is declared with xml using the shortcut version, the `MethodInvokingTaskletAdapter` that is created automatically does not address passing parameters expected by `Tasklet#execute`. This is incorrect according to the documentation of the schema attribute "method":

```
The method suggested should have the same arguments as Tasklet.execute (or a subset), 
and have a compatible return type (boolean, void or RepeatStatus).
```

In this PR, I have created a unit test for each case of this specification (see methods of the [TestTasklet](https://github.com/spring-projects/spring-batch/compare/master...benas:BATCH-2397?expand=1#diff-a3bfb08a898a89b0689823c92af2f839R143) class) and an integration test for the shortcut version.

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/MethodInvokingTaskletAdapterTest.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/MethodInvokingTaskletAdapterTest.java
@@ -1,0 +1,258 @@
+/*
+ * Copyright 2006-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.batch.core.step.tasklet;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.StepExecution;
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.batch.repeat.RepeatStatus;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author Mahmoud Ben Hassine
+ */
+public class MethodInvokingTaskletAdapterTest {
+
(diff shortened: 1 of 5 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/adapter/AbstractMethodInvokingDelegator.java
```diff
@@ -45,7 +45,7 @@
 
 	private String targetMethod;
 
-	private Object[] arguments;
+	protected Object[] arguments;
```
mminella: Why would we make this field protected instead of adding a getter?

spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/MethodInvokingTaskletAdapterTest.java
```diff
@@ -0,0 +1,239 @@
...
+		
+		public StepContribution getStepContribution() {
+			return stepContribution;
+		}
+		
+		public ChunkContext getChunkContext() {
+			return chunkContext;
+		}
+	}
+}
```
mminella: Please add a new line at the end of the file

spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/MethodInvokingTaskletAdapterTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.StepExecution;
+import org.springframework.batch.core.scope.context.ChunkContext;
+import org.springframework.batch.repeat.RepeatStatus;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+public class MethodInvokingTaskletAdapterTest {
```
mminella: Thanks for adding the explicit test for this class.  The tests we had for this class were integration tests through the `TaskletParserAdapterTests`.

spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/MethodInvokingTaskletAdapterTest.java
benas: Done.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/adapter/AbstractMethodInvokingDelegator.java
benas: Using a getter is better indeed. PR updated. Thanks for the review!

mminella:
LGTM.  Rebased and merged as 79f3a67883712413d7f3340b926c5b3e30f4960f

PR #568, opened 2018-01-25
BATCH-2465: add note about the importance to have a unique key constraint on the sortKey

benas:
This PR is for [BATCH-2465](https://jira.spring.io/browse/BATCH-2465).

It updates the reference documentation and the Javadoc to mention explicitly the importance to have a unique key constraint on the `sortKey` when using paging item readers.

I deliberately used "It is important" and not "It is required" (or mandatory) because in the end this is not required but in that case there is no guarantee to not lose data between executions.

Please let me know if the wording should be updated.


mminella:
LGTM.  Rebased and merged as 54f3a5c837551a2f3b619e14c6a737c25f698187

PR #571, opened 2018-01-30
BATCH-2564: make AutomaticJobRegistrar implement SmartLifeCycle

benas:
This PR fixes [BATCH-2564](https://jira.spring.io/browse/BATCH-2564).

The `AutomaticJobRegistrar` is started on `ContextRefreshedEvent` event. This is sometimes too late to register jobs especially when other life cycle components using jobs are started before this registrar.

This commit changes the` AutomaticJobRegistrar` to implement `SmartLifeCycle` and makes its autoStartup and phase properties configurable.

It should be noted that the `onApplicationEvent` method has been removed even if it is a public API. This method is not intended to be used by client code and even if it was, its usage is considered wrong anyway.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/AutomaticJobRegistrar.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/AutomaticJobRegistrar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2007 the original author or authors.
+ * Copyright 2006-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/AutomaticJobRegistrar.java
```diff
@@ -54,6 +53,10 @@
 
 	private volatile boolean running = false;
 
+	private int phase;
```
garyrussell: You should probably default to an early `phase`, such as `Integration.MIN_VALUE + 1000` (leaving room for other components to start earlier, but starting well before components with the default phase (0).

benas: Thank you for the review Gary! Indeed, a negative value is better as a default phase. I will update the PR accordingly.

benas: I have updated the PR to set the default phase to `Integer.MIN_VALUE + 1000`.

mminella:
LGTM.  Rebased and merged as 4dc406ef6083d6fdeb4ceac7fb9c4828c28b0a92.

PR #572, opened 2018-01-30
Made the "Both" option make sense

Buzzardo:
I edited the leader paragraphs above each code block that was flagged as being XML or Java content, such that the paragraphs make sense when both the XML and the Java blocks are present. I turned the Both button back on. I also caught a bunch of other editing things.

### DIFF
--- a/spring-batch-docs/asciidoc/common-patterns.adoc
+++ b/spring-batch-docs/asciidoc/common-patterns.adoc
@@ -50,8 +50,10 @@ public class ItemFailureLoggerListener extends ItemListenerSupport {
 }
 ----
 
-Having implemented this listener, it must be registered with a step, as shown in the
-following example:
+Having implemented this listener, it must be registered with a step.
+
+[role="xmlContent"]
+The following example shows how to register a listener with a step in XML:
 
 .XML Configuration
 [source, xml, role="xmlContent"]
@@ -65,6 +67,9 @@ following example:
 </listeners>
 </step>
 ----
+
+[role="javaContent"]
+The following example shows how to register a listener with a step Java:
 
 .Java Configuration
 [source, java, role="javaContent"]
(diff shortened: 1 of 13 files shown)


benas:
I tested this PR and noticed that the "Both" option will introduce some repetitive sentences making the documentation not coherent:

screenshot 2018-12-13 at 12 42 36

Moreover, we need to make sure every section having a role corresponds to its snippet and test it with the "Both" option to avoid inconsistencies like the following one:

screenshot 2018-12-13 at 12 40 11

To be honest, I don't see the added value of this "Both" option, at least with the current design. May be we can update the style to show snippets side by side, but still.

@Buzzardo @mminella What do you think?

Buzzardo:
Those are errors we need to correct. I'll make a new PR for them.

Buzzardo:
Do you want an issue for this or is it too simple for that?

benas:
No need for an issue or a new PR. You can fix those in this PR so we merge a coherent change set at once.

Buzzardo:
Hi, Mahmoud,

That was tricky. I had to rephrase a bit to make it make sense and still keep the key information. I also fixed a couple grammatical errors that I caught while I was there.

Please check this commit closely, as I had to resolve a bunch of conflicts due to the age of the original commit. I THINK I got everything right, but please do check.

Thanks much.

benas:
@Buzzardo Thank you for all these updates! I did a quick check and it looks ok now. Let's iterate on this. Could you please rebase the PR on the latest master branch? Thank you upfront.

Buzzardo:
@benas I have rebased and cleared up all the conflicts.

benas:
Rebased, squashed and merged as efdce562b0e71695221f6b165ab5979a9b6c3e66. I did another check and noticed a couple of typos and rendering issues like:

Screenshot 2020-05-20 at 11 49 45

Screenshot 2020-05-20 at 12 06 00

which I fixed in ffd39087a84335d15641e687f5fd328906bb3b63.

@Buzzardo Thank you very much for your PR!

PR #573, opened 2018-01-30
BATCH-2673: fix the PDF version of the reference documentation

benas:
Currently, the PDF version of the reference documentation is correctly generated (in `build/asciidoc/pdf/index-single.pdf`) but not included in the distribution (instead, it is the empty `toggle.pdf` file that is included).

This PR fixes this issue and renames the `index-single.pdf` to `spring-batch-reference.pdf` (like the [current](https://docs.spring.io/spring-batch/trunk/reference/pdf/spring-batch-reference.pdf) file name)


mminella:
LGTM.  Rebased and merged as 7a33b1a5d5658a3a781dd59e8b05398ad819efc4

PR #575, opened 2018-01-31
BATCH-2491: use LinkedHashSet instead of HashSet for command line options

benas:
This PR fixes [BATCH-2491](https://jira.spring.io/browse/BATCH-2491).

Currently, options passed on the command line are collected in a `HashSet`. This does not keep options order as they are passed in the command line.

This PR changes the "opts" variable type to `LinkedHashSet`. Note there is no test for this change as "opts" is a local variable to the main method.


mminella:
LGTM.  Rebased and merged as daf0ecfc9fe9cc66c5ca5f00a94f8803715bedb4.

PR #576, opened 2018-02-01
BATCH-2445: make annotation based chunk listeners available to non-fault tolerant steps

benas:
This PR fixes [BATCH-2445](https://jira.spring.io/browse/BATCH-2445).

Currently, annotation based chunk listeners are not registered when using a non-fault tolerant step builder. This PR moves the code of chunk listener annotations handling to the `AbstractTaskletStepBuilder` so that other tasklet step builders can use it.


mminella:
Merged as 1ed5a33bd108e285ace077b7b09982f4ab20c809

PR #578, opened 2018-02-06
BATCH-2679: Add "Back to index" link in the sidebar of the documentation

benas:
This PR resolves [BATCH-2679](https://jira.spring.io/browse/BATCH-2679).

It adds a "Back to index" link to the sidebar of the reference documentation.
The snippet was [friendly copied](https://github.com/spring-projects/spring-framework/blob/master/src/docs/asciidoc/docinfo-footer.html#L37-L44) from Spring Framework's docs 😄 

@Buzzardo Hi Jay, I wanted to add you as a reviewer for this PR but was not able find you in the list. What do you think of this change?

### DIFF
--- a/spring-batch-docs/asciidoc/docinfo-footer.html
+++ b/spring-batch-docs/asciidoc/docinfo-footer.html
@@ -36,4 +36,14 @@
     window.addEventListener('resize', handleTocOnResize);
     window.addEventListener('tocRefresh', handleTocRefresh);
     handleTocOnResize();
+
+    var link = document.createElement("a");
+    link.setAttribute("href", "index.html");
+    link.innerHTML = "<i class=\"fa fa-chevron-left\"></i>&nbsp;&nbsp;Back to index";
+    var p = document.createElement("p");
+    p.setAttribute("id", "backToIndex");
+    p.appendChild(link);
+    var toc = document.getElementById('toc');
+    toc.insertBefore(p, toctitle);
+
 </script>


Buzzardo:
I would like the "< Back to index" link to be above the "Table of Contents" heading. To that end, I changed the code in my local instance (so that I could test my idea) to the following:

    var link = document.createElement("a");
    link.setAttribute("href", "index.html");
    link.innerHTML = "<i class=\"fa fa-chevron-left\"></i>&nbsp;&nbsp;Back to index";
    var p = document.createElement("p");
    p.setAttribute("id", "backToIndex");
    p.appendChild(link);
    var toc = document.getElementById('toc')
    toc.insertBefore(p, toctitle);

That also simplifies the code, which I always like.

If we all agree that my change makes sense, I'll open a ticket with the Framework team to move the link in their documents, too.

mminella:
Where are we at with this?  The proposal by @Buzzardo puts the link in a different place.  If everyone is ok with it, can we please get the PR updated?

benas:
It's ok for me to put the link above the "Table of Contents" as suggested by Jay.

@mminella Is this ok for you too? If yes, I'll update the PR accordingly.

mminella:
Yes please.

benas:
I have updated the PR with the snippet from Jay.
Tested locally and works fine.

mminella:
Thanks for the updates.  Rebased and merged as 619a85f2f2b652d9839af1f3cf07e20d805a9cde.

PR #579, opened 2018-02-12
BATCH-2401: fix CPU intensive loop while polling partitions completion

benas:
This PR resolves [BATCH-2401](https://jira.spring.io/browse/BATCH-2401).

Currently, the `JsrPartitionHandler` is polling partitions completion continuously. This causes a high CPU usage.

This PR makes the polling thread sleep for a configurable amount of time in order to decrease CPU usage during the polling period.


mminella:
LGTM.  Rebased and merged as e11b838452e02cf793365dcef436c506dff5c2de

PR #581, opened 2018-02-16
BATCH-2537: add support for @Primary annotated data sources

benas:
This PR resolves [BATCH-2537](https://jira.spring.io/browse/BATCH-2537).

Currently, the data source configuration in `AbstractBatchConfiguration` is defined as follows:

1. If there is no data source defined in the context, a map based job repository is used
2. If there is exactly one data source defined, it is used
3. If there are 2+ data sources,  an `IllegalStateException` is thrown **even if one of the data sources is annotated with `@Primary` (which should be the one to use)**.

This PR makes it possible to use the data source annotated with `@Primary` when multiple data sources are defined in the context. The new configuration is defined as follows:

1. If there is no data source defined in the context, a map based job repository is used
2. If there is exactly one data source defined, it is used
```diff
--3. If there are 2+ data sources,  an `IllegalStateException` is thrown **even if one of the data sources is annotated with `@Primary` (which should be the one to use)**.
++3.1 If there are 2+ data sources and one of them is annotated with `@Primary`,  this one is used.
++3.2 If there are 2+ data sources and none of them is annotated with `@Primary`,  a `UnsatisfiedDependencyException` is thrown (except when one them is named "dataSource", in which case it will be used (due to auto wiring by name)).
```

When a `UnsatisfiedDependencyException` is thrown, it is clearly said that Spring is not able to autowire a data source due to the presence of multiple candidates. In this case, the user may add `@Primary` on one of the data sources or name one of them "dataSource". This PR clearly documents this detail in the javadoc of `@EnableBatchProcessing`.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2013 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


mminella:
LGTM.  Merged as 7a9a2a9c50c0e1f7b103dd8e81e283e3503a85bb

PR #584, opened 2018-02-23
BATCH-2213: unwrap exceptions thrown from annotated step listeners

benas:
This PR resolves [BATCH-2213](https://jira.spring.io/browse/BATCH-2213).

Currently, when a method of an annotated step listener throws an exception, the exception is wrapped in a `InvocationTargetException` (by the reflection API) which in turn is wrapped in a `IllegalArgumentException` (by Spring Batch). This requires the user to unwrap the original exception from the `StepListenerFailedException`. This behavior is not consistent with interface based listeners where the original exception is the root cause of `StepListenerFailedException`s.

This PR unwraps the original exception and makes it the root cause of `StepListenerFailedException`.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/listener/MulticasterBatchListener.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/listener/MulticasterBatchListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2013 the original author or authors.
+ * Copyright 2006-2018 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


mminella:
Per our conversation, we're putting this PR on hold for 4.1.

benas:
Michael,

Can you please review this PR? I think we can now proceed on this issue since we are working v4.1.

Kr
Mahmoud

mminella:
Rebased and merged as 61bec2805dda06b25f1dd5c3030066754cdf7b94

PR #586, opened 2018-02-28
BATCH-2587: Deprecate the use of the Neo4jOperations API

benas:
This PR resolves [BATCH-2587](https://jira.spring.io/browse/BATCH-2587).


mminella:
This may not be as simple as I thought.  I thought we already had the session based APIs in here.  We can't depricate something without pointing users to the new API...

benas:
yes, I wish we had `forRemoval` attribute in the `@Deprecated` earlier than Java 9 ..

benas:
I updated the PR to document the fact that these APIs are for removal in v4.x.

mminella:
Rebased and merged as 4b889f9b90e7cc4c3812aad0c88a3816634bede5

PR #588, opened 2018-03-11
Allow override factory properties in user-defined configurer subclass

andreysaksonov:
E.g. to set custom serializer. It was a pain to workaround https://jira.spring.io/browse/BATCH-2575 in spring-batch 3.x because neither factories nor dataSource/transactionManager are not exposed to subclass of default configurer - so the only solution is to create new implementation of BatchConfigurer and copy-paste most of the DefaultBatchConfigurer code there just to call one setter on factory.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/DefaultBatchConfigurer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/DefaultBatchConfigurer.java
@@ -125,18 +125,28 @@ public class DefaultBatchConfigurer implements BatchConfigurer {
 		return jobLauncher;
 	}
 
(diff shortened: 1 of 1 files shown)


mminella:
I think this takes the wrong approach.  We really should enable sub classes better access to the `DataSource` and `TransactionManager` involved instead of exposing factory beans like this.   There is a pull request that already does that in which we are currently looking at bringing in (in some form) for 4.1: https://github.com/spring-projects/spring-batch/pull/415

andreysaksonov:
@mminella The idea was to use exposed methods first in subclass (super.createFooFactory) first, then apply additional configurations on the factory (call .setSomeProperty on factory), then DefaultConfigurer will use customized factories to create actual instance of repository/explorer. Exposure of transactionManager/dataSource will require to copy-paste again creation of factories/beans in subclass.

PR #593, opened 2018-04-05
BATCH-2624: Chunk oriented step builders must validate that an ItemWriter is provided

benas:
This PR rebases #491 on the master branch and adds a test on the mandatory item writer.

I took the opportunity to also add a test on the mandatory item reader.


benas:
Thanks for the review. I updated the PR.

PR #597, opened 2018-04-16
BATCH-2699: Move derby's generated files under "build" directory

benas:
This PR resolves [BATCH-2699](https://jira.spring.io/browse/BATCH-2699).

To test the fix:

1. `git checkout master`
2. `./gradlew clean build`
3. `git checkout 3.0.x`
4. `./gradlew clean build` --> fails

Apply the patch in this PR then repeat steps 1 to 4. Now the build should work fine.

### DIFF
--- a/spring-batch-core-tests/src/test/java/test/jdbc/datasource/DerbyDataSourceFactoryBean.java
+++ b/spring-batch-core-tests/src/test/java/test/jdbc/datasource/DerbyDataSourceFactoryBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009-2014 the original author or authors.
+ * Copyright 2009-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spring-batch-core-tests/src/test/java/test/jdbc/datasource/DerbyDataSourceFactoryBean.java
```diff
@@ -38,7 +38,7 @@ protected DataSource createInstance() throws Exception {
 		System.setProperty("derby.storage.pageCacheSize", "100");
 
 		final EmbeddedDataSource ds = new EmbeddedDataSource();
-		ds.setDatabaseName("derbydb");
+		ds.setDatabaseName("build/derbydb");
```
benas: Even though this is a setter for the database "name", without this change the `derbydb` folder will be created under the root directory of the module and not under `build` directory.

spring-batch-infrastructure/src/test/java/test/jdbc/datasource/DerbyDataSourceFactoryBean.java
```diff
@@ -42,7 +42,7 @@ protected DataSource createInstance() throws Exception {
 		System.setProperty("derby.storage.pageCacheSize", "100");
 
 		final EmbeddedDataSource ds = new EmbeddedDataSource();
-		ds.setDatabaseName("derbydb");
+		ds.setDatabaseName("build/derbydb");
```
benas: Even though this is a setter for the database "name", without this change the `derbydb` folder will be created under the root directory of the module and not under `build` directory.

mminella:
LGTM.  Rebased and merged as 2913c8de66945295085547966a2b9d20c1c4b039.

PR #598, opened 2018-04-17
Fix typos in JdbcPagingItemReaderBuilder

drumonii:
Fix some spelling typos in javadoc, typo in an `IllegalArgumentException` message, and typo with an accidental field usage.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcPagingItemReaderBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcPagingItemReaderBuilder.java
@@ -331,7 +331,7 @@ public class JdbcPagingItemReaderBuilder<T> {
 	private PagingQueryProvider determineQueryProvider(DataSource dataSource) {
 
 		try {
-			DatabaseType databaseType = DatabaseType.fromMetaData(this.dataSource);
+			DatabaseType databaseType = DatabaseType.fromMetaData(dataSource);
 
 			AbstractSqlPagingQueryProvider provider;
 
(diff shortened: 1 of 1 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcPagingItemReaderBuilder.java
```diff
@@ -331,7 +331,7 @@
 	private PagingQueryProvider determineQueryProvider(DataSource dataSource) {
 
 		try {
-			DatabaseType databaseType = DatabaseType.fromMetaData(this.dataSource);
+			DatabaseType databaseType = DatabaseType.fromMetaData(dataSource);
```
mminella: The `this` should remain.  Using `this` is part of the Spring Framework coding standards: https://github.com/spring-projects/spring-framework/wiki/Code-Style#field-and-method-references

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcPagingItemReaderBuilder.java
```diff
@@ -331,7 +331,7 @@
 	private PagingQueryProvider determineQueryProvider(DataSource dataSource) {
```
mminella: This method could probably be static...

drumonii: If I set that method to static then I'd need to pass in `selectClause`, `fromClause`, `whereClause`, `groupClause`, `sortKeys`, as they are referenced using `this`. But I guess at that point, might as well use `this.dataSource` without a parameter and keep it non static. Let me know...

mminella: You're right.  That's what I get from reading just the diff.  Looks good to me. Thanks for the updates!

benas:
LGTM. rebased and merged as a8ecb4fdcff643f4e71b9e11cb5df2a7452c5544.

Thank you @drumonii for your PR!

PR #599, opened 2018-04-18
BATCH-2686: add builders for master/worker beans in remote chunking

benas:
This PR resolves [BATCH-2686](https://jira.spring.io/browse/BATCH-2686).


spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+	private static final String OUTPUT_CHANNEL_DEFAULT_NAME = "replies";
+	private static final String SERVICE_ACTIVATOR_METHOD_NAME = "handleChunk";
+
+	private ItemProcessor<I, O> itemProcessor;
+	private ItemWriter<O> itemWriter;
+	private GenericApplicationContext applicationContext;
+	private String inputChannelName = INPUT_CHANNEL_DEFAULT_NAME;
+	private String outputChannelName = OUTPUT_CHANNEL_DEFAULT_NAME;
+
+	public RemoteChunkingSlaveStepBuilder<I, O> itemProcessor(ItemProcessor<I, O> itemProcessor) {
```
mminella: Method level javadoc

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+		this.itemProcessor = itemProcessor;
+		return this;
+	}
+
+	public RemoteChunkingSlaveStepBuilder<I, O> itemWriter(ItemWriter<O> itemWriter) {
```
mminella: Method level javadoc

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+		this.itemWriter = itemWriter;
+		return this;
+	}
+
+	public RemoteChunkingSlaveStepBuilder<I, O> inputChannelName(String inputChannelName) {
```
mminella: Method level javadoc

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+		this.inputChannelName = inputChannelName;
+		return this;
+	}
+
+	public RemoteChunkingSlaveStepBuilder<I, O> outputChannelName(String outputChannelName) {
```
mminella: Method level javadoc

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+		this.outputChannelName = outputChannelName;
+		return this;
+	}
+
+	public RemoteChunkingSlaveStepBuilder<I, O> applicationContext(GenericApplicationContext applicationContext) {
```
mminella: Method level javadoc

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingSlaveStepBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+ *     {@link ChunkProcessorChunkHandler} as a service activator</li>
+ * </ul>
+ *
+ * @param <I> type of input items
+ * @param <O> type of output items
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+public class RemoteChunkingSlaveStepBuilder<I, O> {
```
mminella: Can we call this a worker step instead of a slave step?

spring-batch-samples/src/main/java/org/springframework/batch/sample/remotechunking/WorkerConfiguration.java
```diff
@@ -0,0 +1,168 @@
...
+	public IntegrationFlow outgoingReplies(AmqpTemplate rabbitTemplate) {
+		return IntegrationFlows
+				.from("replies")
+				.handle(Amqp.outboundAdapter(rabbitTemplate).routingKey("replies"))
+				.get();
+	}
+
+	@Bean
+	public IntegrationFlow integrationFlow(DirectChannel requests, DirectChannel replies) {
+		return workerFlowBuilder
```
mminella: `this.`

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
```diff
@@ -0,0 +1,172 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = {
+		JobRunnerConfiguration.class,
+		RemoteChunkingJobFunctionalTests.MasterConfiguration.class})
+public class RemoteChunkingJobFunctionalTests {
+
+	@Autowired
+	private JobLauncherTestUtils jobLauncherTestUtils;
+
+	@Ignore("This test is ignored because a worker JVM should be running to process chunks")
```
mminella: Why can't we launch the other JVM?

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -948,6 +948,77 @@ when it receives chunks from the master.
 For more information, see the section of the "Scalability" chapter on
 link:$$http://docs.spring.io/spring-batch/reference/html/scalability.html#remoteChunking$$[Remote Chunking].
 
+Starting from version 4.1, Spring Batch Integration provides the `@EnableBatchIntegration`
```
mminella: I'd love to see your diagram included in the reference documentation.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+		super(new StepBuilder(stepName));
+	}
+
+	/**
+	 * Set the application context in which required beans will be registered.
+	 *
+	 * @param applicationContext the application context to set.
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemoteChunkingMasterStepBuilder<I, O> applicationContext(GenericApplicationContext applicationContext) {
```
mminella: I _think_ this should be a `BeanDefinitionRegistry`: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/support/BeanDefinitionRegistry.html#registerBeanDefinition-java.lang.String-org.springframework.beans.factory.config.BeanDefinition-

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+		return this;
+	}
+
+	/**
+	 * Set the input channel on which replies from workers will be received.
+	 *
+	 * @param inputChannel the input channel
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemoteChunkingMasterStepBuilder<I, O> inputChannel(PollableChannel inputChannel) {
```
mminella: Any of these setters that map to something else, I try to put a `@see` on as well.  For this example, I'd do `@see ChunkMessageChannelItemWriter#setReplyChannel()`.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+		return this;
+	}
+
+	/**
+	 * Set the {@link MessagingTemplate} to use to send data to workers.
+	 *
+	 * @param messagingTemplate the messaging template to use
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemoteChunkingMasterStepBuilder<I, O> messagingTemplate(MessagingTemplate messagingTemplate) {
```
mminella: We need to be clear here that we are going to override the default destination for this instance in case a user accidentally shares it.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+	 */
+	public TaskletStep build() {
+
+		Assert.notNull(applicationContext, "A GenericApplicationContext must be provided");
+
+		// configure messaging template
+		if (messagingTemplate == null) {
+			messagingTemplate = new MessagingTemplate();
+		}
+		messagingTemplate.setDefaultChannel(outputChannel);
```
mminella: `this.messagingTemplate`

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderFactory.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Convenient factory for a {@link RemoteChunkingMasterStepBuilder} which sets
+ * the {@link JobRepository}, {@link PlatformTransactionManager} and
+ * {@link GenericApplicationContext} automatically.
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+public class RemoteChunkingMasterStepBuilderFactory {
```
mminella: Is there a reason to expose this or should we make this an inner class to the `BatchIntegrationConfiguration`?

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+ */
+package org.springframework.batch.integration.chunk;
+
+import org.springframework.batch.core.ChunkListener;
+import org.springframework.batch.core.ItemProcessListener;
+import org.springframework.batch.core.ItemReadListener;
+import org.springframework.batch.core.ItemWriteListener;
+import org.springframework.batch.core.SkipListener;
+import org.springframework.batch.core.StepExecutionListener;
+import org.springframework.batch.core.jsr.RetryListener;
```
mminella: Wrong import I believe...

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+
+	private PlatformTransactionManager transactionManager;
+
+	private GenericApplicationContext applicationContext;
+
+	@Autowired
+	public BatchIntegrationConfiguration(
+			JobRepository jobRepository,
+			PlatformTransactionManager transactionManager,
+			GenericApplicationContext applicationContext) {
```
mminella: See previous comment about `BeanDefinitionRegistry`.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -0,0 +1,86 @@
...
+ *       	 workerBuilder
+ *       		.itemProcessor(itemProcessor())
+ *       		.itemWriter(itemWriter())
+ *       		.inputChannel(incomingRequestsFromMaster())
+ *       		.outputChannel(outgoingRepliesToMaster())
+ *       		.build();
+ * 	}
+ *
+ * }
+ * </pre>
```
mminella: Very nicely done javadoc

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+
+import java.util.Arrays;
+
+/**
+ * @author Mahmoud Ben Hassine
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = {RemoteChunkingMasterStepBuilderTest.BatchConfiguration.class})
+public class RemoteChunkingMasterStepBuilderTest {
```
mminella: I hate to do this, but we'll need coverage for the dummy setters to make sure they are still there and that they don't modify the results...

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	/**
+	 * Build a master {@link TaskletStep}.
+	 *
+	 * @return the configured master step
+	 */
+	public TaskletStep build() {
+
+		Assert.notNull(applicationContext, "A GenericApplicationContext must be provided");
```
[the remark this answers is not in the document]
benas: fixed

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+ */
+public class RemoteChunkingMasterStepBuilder<I, O> extends FaultTolerantStepBuilder<I, O> {
+
+	private MessagingTemplate messagingTemplate;
+	private GenericApplicationContext applicationContext;
+	private PollableChannel inputChannel;
+	private DirectChannel outputChannel;
+
+	private final int DEFAULT_MAX_WAIT_TIMEOUTS = 40;
+	private static final long DEFAULT_THROTTLE_LIMIT = 6;
```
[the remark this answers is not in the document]
benas: Taken from `ChunkMessageChannelItemWriter`

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingWorkerFlowBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+ * </ul>
+ *
+ * @param <I> type of input items
+ * @param <O> type of output items
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+
+public class RemoteChunkingWorkerFlowBuilder<I, O> {
```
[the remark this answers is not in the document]
benas: fixed

spring-batch-samples/src/main/java/org/springframework/batch/sample/remotechunking/WorkerConfiguration.java
benas: fixed

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: I agree! fixed

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: fixed

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: I added this clarification.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: done

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderFactory.java
benas: I used the same approach as `StepBuilderFactory` and `JobBuilderFactory` which are public. It could be an inner class of `BatchIntegrationConfiguration`, but in this case, when we autowire the beans, we would write something like:

```java
@Autowired
private BatchIntegrationConfiguration.RemoteChunkingMasterStepBuilderFactory factory;
```

IMO, the user has not to know about the "internal" `BatchIntegrationConfiguration` (even if it's public) just like `AbstractBatchConfiguration` and its subclasses.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
benas: done

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
benas: fixed.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: This is used in overridden method:

```java
@Override
public RemoteChunkingMasterStepBuilder<I, O> listener(RetryListener listener) {
	super.listener(listener);
	return this;
}
```

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
benas: We can launch the other JVM but the issue is when to stop it? Since on the worker side we don't have a job/step, I could not use a listener for example to stop the JVM. It's technically possible to stop the JVM once all chunks received from the master are processed, but this technical code will "pollute" the example and may disturb the user from the main purpose of the sample.

I wrote a unit test on the master side to be consistent with other samples. But AFAIK, this is the first sample requiring two JVMs (the amqp sample does not wait for the other jvm, it just sends some messages to a queue and dies). Since it's a sample, it's probably better to write a main method on the master side too and remove this unit test. What do you think?

spring-batch-docs/asciidoc/spring-batch-integration.adoc
benas: I added the diagram and updated the documentation accordingly

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
mminella: I'd use `@Before` and `@After` to manage the worker.  That keeps the management of the other JVMs separate from the test code.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
mminella: You can remove that method.  It's only used internally for our JSR implementation.  JSR-352 does not support remote chunking or java configuration.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
mminella: The code here says otherwise...

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: ok, I removed the method.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
benas: fixed, my mistake, sorry for that.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
mminella: We probably should note at the class level (?) that if a `MessagingTemplate` is not provided, we will create one that uses the default values.  I can add this on merge if you agree.

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
mminella: Where?

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
```diff
@@ -0,0 +1,182 @@
...
+
+	@Autowired
+	private JobLauncherTestUtils jobLauncherTestUtils;
+
+	private AnnotationConfigApplicationContext workerApplicationContext;
+
+	@Before
+	public void setUp() {
+		workerApplicationContext = new AnnotationConfigApplicationContext(WorkerConfiguration.class);
+		System.out.println("Started a worker..");
```
mminella: Should remove system.outs...

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	@After
+	public void tearDown() {
+		System.out.println("Stopping the worker..");
```
mminella: Again the system.outs

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * <p>The master step of the tested job will read data and send chunks to the worker
+ * (started in {@link RemoteChunkingJobFunctionalTests#setUp()}) for processing and writing.</p>
+ */
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = {
+		JobRunnerConfiguration.class,
+		RemoteChunkingJobFunctionalTests.MasterConfiguration.class})
+public class RemoteChunkingJobFunctionalTests {
```
mminella: Stupid question, but since this is all within the same JVM now...does it make sense to use a broker that can be bootstrapped within the test?  Something like ActiveMQ (I believe we use that elsewhere in tests)?

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingWorkerBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Create an {@link IntegrationFlow} with a {@link ChunkProcessorChunkHandler}
+	 * configured as a service activator listening to the input channel and replying
+	 * on the output channel.
+	 *
+	 * @return the integration flow
+	 */
+	public IntegrationFlow build() {
+
+		Assert.notNull(itemWriter, "An ItemWriter must be provided");
```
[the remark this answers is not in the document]
benas: fixed in 10271e1f11dec35df413346763c6e5cb0469deef

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
benas: Makes sense. I updated the sample to use an embedded broker. See f21dbd550b70f9a0deb043608e3ee407c4c0ec43

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -0,0 +1,86 @@
...
+package org.springframework.batch.integration.config.annotation;
+
+import org.springframework.batch.integration.chunk.RemoteChunkingWorkerBuilder;
+import org.springframework.context.annotation.Import;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
```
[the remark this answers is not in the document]
benas: fixed in 8faecf4fdc16ae95da43d3fe8ed601b142b0fec5

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
benas: I thought that's fine for the sample (not production code). Removed in 83d43d60c578f475ab08d7262edee51cfce0d1a4

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
benas: fixed in 83d43d60c578f475ab08d7262edee51cfce0d1a4

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.support.GenericApplicationContext;
+import org.springframework.integration.channel.DirectChannel;
+import org.springframework.integration.channel.QueueChannel;
+import org.springframework.messaging.PollableChannel;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.transaction.PlatformTransactionManager;
+
+import java.util.Arrays;
```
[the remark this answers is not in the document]
benas: fixed in 8faecf4fdc16ae95da43d3fe8ed601b142b0fec5

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemoteChunkingJobFunctionalTests.java
```diff
@@ -0,0 +1,182 @@
...
+import org.springframework.integration.channel.DirectChannel;
+import org.springframework.integration.channel.QueueChannel;
+import org.springframework.integration.config.EnableIntegration;
+import org.springframework.integration.dsl.IntegrationFlow;
+import org.springframework.integration.dsl.IntegrationFlows;
+import org.springframework.messaging.PollableChannel;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import java.util.Arrays;
```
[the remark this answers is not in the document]
benas: fixed in 8faecf4fdc16ae95da43d3fe8ed601b142b0fec5

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
```diff
@@ -0,0 +1,439 @@
...
+import org.springframework.integration.core.MessagingTemplate;
+import org.springframework.messaging.PollableChannel;
+import org.springframework.retry.RetryPolicy;
+import org.springframework.retry.backoff.BackOffPolicy;
+import org.springframework.retry.policy.RetryContextCache;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.interceptor.TransactionAttribute;
+import org.springframework.util.Assert;
+
+import java.util.function.Function;
```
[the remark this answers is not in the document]
benas: fixed in 8faecf4fdc16ae95da43d3fe8ed601b142b0fec5

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+	public BatchIntegrationConfiguration(
+			JobRepository jobRepository,
+			PlatformTransactionManager transactionManager) {
+		this.jobRepository = jobRepository;
+		this.transactionManager = transactionManager;
+	}
+
+	@Bean
+	public RemoteChunkingMasterStepBuilderFactory remoteChunkingMasterStepBuilderFactory() {
+		return new RemoteChunkingMasterStepBuilderFactory(jobRepository, transactionManager, beanDefinitionRegistry);
```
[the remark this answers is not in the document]
benas: I added "this" keyword where I thought is appropriate in 0b6d8122b45cc7d29a41328590631b902bd44a9b. But looks like you meant use the keyword systematically for every class level field. I don't see we use this convention in `StepBuilderFactory` and `JobBuilderFactory` for example. But I fixed this in 8fbc847148b7c3281fc07a6d7e75981e652e6b4c

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+
+/**
+ * Base configuration class for Spring Batch Integration factory beans.
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+@Configuration
+public class BatchIntegrationConfiguration implements ApplicationContextAware {
```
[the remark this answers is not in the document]
benas: In hindsight, we don't really need to register the `RemoteChunkHandlerFactoryBean` as a bean in the application context or bean definition registry. It is only required to have a `ChunkProcessorChunkHandler` which is already a SI message endpoint acting as a service activator (through annotations). So I used the `RemoteChunkHandlerFactoryBean` to create a `ChunkProcessorChunkHandler` and removed the dependency to the `BeanDefinitionRegistry`. Less is more!

Fixed in 94bd4d161e39b91b6d5c73c57800599853f4a1ea.

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
benas: I thought you were talking about setters of the current class: `maxWaitTimeout`, `throttleLimit` (not the ones overridden). So by fixed I meant here: ba2534ede3123df0562f98e22a5e60c41a8558e0

Now for "dummy" setters, I added a test to cover them here: 61737be99e214e81cb0ce992c6c582d3b888681a

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilder.java
benas: I agree. I added a clarification at the class level about this detail. See 0a7eda3a776917d9c89860d220bfc546f2e3370a

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderFactory.java
```diff
@@ -0,0 +1,60 @@
...
+	 * Creates a {@link RemoteChunkingMasterStepBuilder} and initializes its job
+	 * repository and transaction manager.
+	 * 
+	 * @param name the name of the step
+	 * @return a {@link RemoteChunkingMasterStepBuilder}
+	 */
+	public <I, O> RemoteChunkingMasterStepBuilder<I, O> get(String name) {
+		return new RemoteChunkingMasterStepBuilder<I, O>(name)
+				.repository(this.jobRepository)
+				.transactionManager(this.transactionManager);
```
artembilan: Looking to this peace of code, I would say that I don't need this `factory`, the simple `RemoteChunkingMasterStepBuilder` is enough for me, since I really can inject any possible `jobRepository` and `transactionManager`. 

However I guess I can override them since you return for me that  `RemoteChunkingMasterStepBuilder`.
But again: I don't see too match benefit of this factory.

Just IMHO.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -0,0 +1,86 @@
...
+ * <pre class="code">
+ * &#064;Configuration
+ * &#064;EnableBatchIntegration
+ * public class AppConfig {
+ *
+ * 	&#064;Autowired
+ * 	private RemoteChunkingMasterStepBuilderFactory masterBuilderFactory;
+ *
+ * 	&#064;Autowired
+ * 	private RemoteChunkingWorkerBuilder workerBuilder;
```
artembilan: Well, indeed, looking to this sample, it really would feel better to have `RemoteChunkingMasterStepBuilder` and `RemoteChunkingWorkerBuilder` - transparent and intuitive. 
Otherwise if I would inject `RemoteChunkingMasterStepBuilderFactory`, I would expect similar for the worker.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -935,19 +935,103 @@ public ChunkProcessorChunkHandler chunkProcessorChunkHandler() {
...
+		@Autowired
+		private RemoteChunkingWorkerBuilder workerBuilder;
+
+		@Bean
+		public IntegrationFlow workerFlow(DirectChannel requests, DirectChannel replies){
+			return this.workerBuilder
+				.itemProcessor(itemProcessor())
+				.itemWriter(itemWriter())
+				.inputChannel(requests) // requests received from the master
+				.outputChannel(replies) // replies sent to the master
```
artembilan: I think you need to add explain somewhere that all these channels must be configured as well.
We are in Spring Batch: that might not be so obvious for end-user what to do for Spring Integration side.

benas: A picture is worth thousand words: https://github.com/benas/spring-batch/blob/c6ffc5da5aec4c75fea066d7c2293c1b26182826/spring-batch-docs/asciidoc/images/remote-chunking-config.png

This is also explained in the docs.

spring-batch-integration/src/main/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderFactory.java
benas: Yes, in the beginning, we had only the builders. But then we thought it is better to add this factory to automatically set the job repository and the transaction manager (just like the `StepBuilderFactory`) and free the user from do it. But you are right, there is a factory on the master side but not on the worker side..

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
benas: You are right, there is a factory on the master side but not on the worker side.. Because the worker side does not know anything about SB artifacts (step, job repository, etc) except a dumb item processor and writer.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
artembilan: Why does end-user have to provide `writer`, `reader` and `processor`?
I thought in this master-worker pattern everything is done by the `RemoteChunkProcessor`...

Thanks

artembilan: OK. Never mind I see: `itemReader` is a source of data for this job - can be DB, file or any thing else.
The `itemProcessor` on the worker side is indeed the part we handle an item.

I can understand that `itemWriter` is for sinking data (e.g. DB, file etc.), but I can't understand where is the correlation with the `replies`? Do we send on our own after successful sinking?

benas: Well, this is the contract between the user and SB to provide at least where to read data from and where to write it (as well as the business logic of how to optionally process it). So I think this is the minimum minimorum of what the user has to provide. As a first step now, the user has also to provide SI channels and Adapters.

* For channels, we tried to create them automatically in the beginning but we faced some lifecycle issues and we decided to leave them for the user as a first step.
* For adapters, we can't take them form the user because we don't know upfront which middleware (s)he is using (jms, amqp, etc)

benas: > Do we send on our own after successful sinking?

The worker use the `ItemWriter` to write data directly to the sink and the `ChunkProcessorChunkHandler` to send back instances of `ChunkResponse` (see return type of `handkeChunk` method) to the master.

benas:
Hi @mminella ,

This PR has been rebased on the latest master and updated with correct example of remote chunking.

Kr,
Mahmoud

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -0,0 +1,88 @@
...
+ *     <li>{@link RemoteChunkingWorkerBuilder}: used to create the integration
+ *     flow on the worker side.</li>
+ * </ul>
+ *
+ * For example:
+ *
+ * <pre class="code">
+ * &#064;Configuration
+ * &#064;EnableBatchIntegration
+ * public class RemoteChunkingAppConfig {
```
mminella: We should add the `@EnableBatchProcessing` here since this by itself would fail (no transaction manager or job repository), correct?

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
```diff
@@ -0,0 +1,264 @@
...
+				.exceptionHandler(new DefaultExceptionHandler())
+				.stepOperations(new RepeatTemplate())
+				.chunkOperations(new RepeatTemplate())
+				.backOffPolicy(new NoBackOffPolicy())
+				.stream(new ItemStreamSupport() {})
+				.keyGenerator(Object::hashCode)
+				.build();
+
+		// then
+		Assert.assertNotNull(taskletStep);
```
mminella: Just because the step was created doesn't mean that all the values passed above were set correctly.  We need to assert each one.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
benas: correct.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -944,6 +938,88 @@ when it receives chunks from the master.
...
+
+There is no need anymore to explicitly configure the `SimpleChunkProcessor`
+and the `ChunkProcessorChunkHandler` (Those can still be explicitly configured if required).
+
+The following example shows how to use these APIs:
+
+[source, java]
+----
+@EnableBatchIntegration
+public class RemoteChunkingJobConfiguration {
```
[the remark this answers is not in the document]
benas: indeed.

spring-batch-integration/src/test/java/org/springframework/batch/integration/chunk/RemoteChunkingMasterStepBuilderTest.java
benas: I agree. I thought the goal was just to make sure these setters are called somewhere so that they are covered (in case we remove on of them accidentally). I was too pragmatic there 😄

mminella: I'm almost done refactoring the test for this and will merge once I'm done.

PR #603, opened 2018-05-10
Add remote chunking sample

benas:
This PR resolves [BATCH-2721](https://jira.spring.io/browse/BATCH-2721). It adds a self contained remote chunking sample that uses an embedded JMS broker (for simplicity). Even though the broker is embedded, communication between the master and workers is still done through JMS queues and Spring Integration channels and messages are sent over the wire through a TCP port.

The sample in this PR will not only enrich the current collection of samples, but also serve as a basis to show the added value of the feature we are trying to implement in [BATCH-2686](https://jira.spring.io/browse/BATCH-2686).

This PR also fixes a few typos in the documentation section about remote chunking.


mminella:
This PR leaves a `activemq-data` directory under `spring-batch-samples` that doesn't get cleaned up on a `./gradlew clean`.  I think that directory needs to be configured to be under a `build` directory.

mminella:
Rebased and merged as 026bbe056b147edf63b5239e1c980410f5059524.

PR #609, opened 2018-05-25
BATCH-2294: Don't replace my PlatformTransactionManager with a new one

elreydetodo:
For some reason DefaultBatchConfigurer decides that you can't possibly
already have a PlatformTransactionManager and makes a new one for you.
Because of the way Spring's bean factory works, this new one is
sometimes created after the one you _actually_ wanted, and it _replaces_
the real one in your application with a lame one that is useless.

This patch changes the implementation of DefaultBatchConfigurer so it
tries to autowire an existing bean and only makes a new one if one
doesn't exist already.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/DefaultBatchConfigurer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/DefaultBatchConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -48,7 +48,7 @@ public class DefaultBatchConfigurer implements BatchConfigurer {
 
 	/**
 	 * Sets the dataSource.  If the {@link DataSource} has been set once, all future
-	 * values are passed are ignored (to prevent {@code}@Autowired{@code} from overwriting
+	 * values passed in are ignored (to prevent {@code}@Autowired{@code} from overwriting
 	 * the value).
 	 *
 	 * @param dataSource
@@ -58,9 +58,19 @@ public class DefaultBatchConfigurer implements BatchConfigurer {
 		if(this.dataSource == null) {
 			this.dataSource = dataSource;
 		}
+	}
 
+	/**
+	 * Sets the transactionManager. If the {@link PlatformTransactionManager} has been set
+	 * once, all future values passed in are ignored (to prevent {@code}@Autowired{@code}
+	 * from overwriting the value).
+	 *
+	 * @param transactionManager
+	 */
+	@Autowired(required = false)
+	public void setTransactionManager(PlatformTransactionManager transactionManager) {
 		if(this.transactionManager == null) {
-			this.transactionManager = new DataSourceTransactionManager(this.dataSource);
+			this.transactionManager = transactionManager;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


elreydetodo:
This PR tries to fix the same issue as #415, which has merge conflicts. I took what seems like the simpler route.

I was unsure how to create a test for this situation. I'm open to suggestions. I'm also happy to make adjustments to the way I fixed it if anyone has a better suggestion.

spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/DefaultBatchConfigurer.java
```diff
@@ -93,12 +103,18 @@ public JobExplorer getJobExplorer() {
 	@PostConstruct
 	public void initialize() {
 		try {
-			if(dataSource == null) {
-				logger.warn("No datasource was provided...using a Map based JobRepository");
-
-				if(this.transactionManager == null) {
+			if(this.transactionManager == null) {
```
elreydetodo: This formatting seems inconsistent with the [Spring Framework Code Style](https://github.com/spring-projects/spring-framework/wiki/Code-Style). There should be a space between "if" and the paren. I mimicked what was already in the file because I felt it would be weird to mix the style within a single file. I'm happy to fix the formatting if you'd like me to, but I want to make sure the approach I took is acceptable before I go to that effort.

elreydetodo:
@mminella can you take a look at this PR? It's doing the same thing as #415, but it doesn't have a merge conflict and it doesn't add that getter you didn't like.

mminella:
@elreydetodo We actually reviewed these PRs internally yesterday.  We liked the structure of the if statements in the init method better in #415 but we liked your logging additions.  If you want to apply the changes he made to your PR and add a test, we'd be happy to merge it.  cc: @benas

benas:
> I was unsure how to create a test for this situation.

I wrote a test suite here: https://github.com/benas/spring-batch/commit/b6fa592ef1115502fd71d04c7125b26f635b1c6c

While testing this PR, I found a couple of issues:

*  When I specify a data source but not a transaction manager, I was expecting a `DataSourceTransactionManager` to be used but it is not the case ([this test](https://github.com/benas/spring-batch/blob/b6fa592ef1115502fd71d04c7125b26f635b1c6c/spring-batch-core/src/test/java/org/springframework/batch/core/configuration/annotation/TransactionManagerConfigurationWithBatchConfigurerTests.java#L57) is failing)
* When I don't specify a data source and no transaction manager, I'm expecting a `ResourcelessTransactionManager` to be used but this is not the case ([this test](https://github.com/benas/spring-batch/blob/b6fa592ef1115502fd71d04c7125b26f635b1c6c/spring-batch-core/src/test/java/org/springframework/batch/core/configuration/annotation/TransactionManagerConfigurationWithBatchConfigurerTests.java#L37) is failing)

Those tests are failing because of the following sequence:

1. `SimpleBatchConfiguration` is loaded, the 5 `@Bean` annotated method are invoked and beans are added to the application context. One of them is the transaction manager which is a lazy proxy with null reference.
2. The setter `DefaultBatchConfigurer#setTransactionManager` is invoked and gets injected the transaction manager bean created in 1)
3. The `DefaultBatchConfigurer#initialize` is invoked where the condition `if(this.transactionManager == null)` is never true.

As mentioned in #415, I think we should not try to autowire the transaction manager. If the user wants a custom transaction manager, he has to provide a custom `BatchConfigurer` (subclassing `DefaultBatchConfigurer` and overriding the getter should be sufficient). Note that this is possible if we use the getter in `initialize` and `createJobRepository` methods (this approach is tested here: https://github.com/benas/spring-batch/commit/b6fa592ef1115502fd71d04c7125b26f635b1c6c, and all tests are passing (except the ones that are ignored because we don't want to address them)).

mminella:
Based on the comments @benas made as well as the PR he submitted to address this issue, I'm closing this PR.  #634 should address the functionality discussed in this issue.  Thank you for the contribution!

PR #612, opened 2018-06-08
Add remote partitioning samples

benas:
This PR fixes the code examples in the remote partitioning section of the documentation and adds two remote partitioning samples:

* One that shows how to poll the job repository to monitor workers progress
* And another one that shows how to aggregate replies sent by workers to notify work completion

The sample deliberately uses a simple worker step (just a tasklet that prints to the console the name of the partition it is processing) in order to focus on how to configure a remote partitioning job.

An embedded JMS broker and an embedded database are used for simplicity.

This PR resolves [BATCH-2730](https://jira.spring.io/browse/BATCH-2730) and serves as a base for [BATCH-2687](https://jira.spring.io/browse/BATCH-2687).

### DIFF
--- a/spring-batch-samples/README.md
+++ b/spring-batch-samples/README.md
@@ -638,12 +638,24 @@ happens inside the main step transaction).
 The purpose of this sample is to show multi-threaded step execution
 using the `PartitionHandler` SPI.  The example uses a
 `TaskExecutorPartitionHandler` to spread the work of reading
-some files acrosss multiple threads, with one `Step` execution
+some files across multiple threads, with one `Step` execution
 per thread.  The key components are the `PartitionStep` and the
 `MultiResourcePartitioner` which is responsible for dividing up
 the work.  Notice that the readers and writers in the `Step`
 that is being partitioned are step-scoped, so that their state does
 not get shared across threads of execution.
+
+### [Remote Partitioning Sample](id:remotePartitioning)
+
+This sample shows how to configure a remote partitioning job. The master step
+uses a `MessageChannelPartitionHandler` to send partitions to and receive 
+replies from workers. Two examples are shown:
+
+* A master step that polls the job repository to see if all workers have finished
+their work
+* A master step that aggregates replies from workers to notify work completion
+
+The sample uses an embedded JMS broker and an embedded database for simplicity. 
 
 ### [Remote Chunking Sample](id:remoteChunking)
 
--- a/spring-batch-samples/README.md
+++ b/spring-batch-samples/README.md
@@ -655,7 +655,9 @@ replies from workers. Two examples are shown:
 their work
 * A master step that aggregates replies from workers to notify work completion
 
-The sample uses an embedded JMS broker and an embedded database for simplicity. 
+The sample uses an embedded JMS broker and an embedded database for simplicity
+but any option supported via Spring Integration for communication is technically
+acceptable. 
 
 ### [Remote Chunking Sample](id:remoteChunking)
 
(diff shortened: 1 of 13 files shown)


spring-batch-samples/src/test/java/org/springframework/batch/sample/RemotePartitioningJobWithRepositoryPollingFunctionalTests.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.test.context.ContextConfiguration;
+
+/**
+ * The master step of the job under test will create 3 partitions for workers
+ * to process.
+ *
+ * @author Mahmoud Ben Hassine
+ */
+@ContextConfiguration(classes = {JobRunnerConfiguration.class, MasterConfiguration.class})
+public class RemotePartitioningJobWithRepositoryPollingFunctionalTests extends RemotePartitioningJobFunctionalTests {
```
mminella: Unless I'm mistaken, the only difference between the two functional tests is the imports (and the names).  Can we look at either using JUnit's parameterized tests or at a minimum, putting everything except how the `ApplicationContext` is created in the parent and just have the subclasses provide their `ApplicationContext` with the correct bits (so the parent does the test itself as well).

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemotePartitioningJobWithMessageAggregationFunctionalTests.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.test.context.ContextConfiguration;
+
+/**
+ * The master step of the job under test will create 3 partitions for workers
+ * to process.
+ *
+ * @author Mahmoud Ben Hassine
+ */
+@ContextConfiguration(classes = {JobRunnerConfiguration.class, MasterConfiguration.class})
+public class RemotePartitioningJobWithMessageAggregationFunctionalTests extends RemotePartitioningJobFunctionalTests {
```
mminella: Unless I'm mistaken, the only difference between the two functional tests is the imports (and the names). Can we look at either using JUnit's parameterized tests or at a minimum, putting everything except how the ApplicationContext is created in the parent and just have the subclasses provide their ApplicationContext with the correct bits (so the parent does the test itself as well).

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemotePartitioningJobWithRepositoryPollingFunctionalTests.java
benas: Good catch! I updated the PR accordingly

spring-batch-samples/src/test/java/org/springframework/batch/sample/RemotePartitioningJobWithMessageAggregationFunctionalTests.java
benas: Good catch! I updated the PR accordingly

spring-batch-samples/README.md
```diff
@@ -638,13 +638,25 @@ happens inside the main step transaction).
...
+
+This sample shows how to configure a remote partitioning job. The master step
+uses a `MessageChannelPartitionHandler` to send partitions to and receive 
+replies from workers. Two examples are shown:
+
+* A master step that polls the job repository to see if all workers have finished
+their work
+* A master step that aggregates replies from workers to notify work completion
+
+The sample uses an embedded JMS broker and an embedded database for simplicity.
```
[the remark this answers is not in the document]
benas: That's better. I updated the README.

mminella:
Rebased, squashed, and merged as 5730f380e28a822681b6685a69695ee54a0edec6.

PR #617, opened 2018-06-26
Add a validator implementation based on the Bean Validation API

benas:
This PR resolves [BATCH-2690](https://jira.spring.io/browse/BATCH-2690).

### DIFF
--- a/spring-batch-docs/asciidoc/readersAndWriters.adoc
+++ b/spring-batch-docs/asciidoc/readersAndWriters.adoc
@@ -2659,6 +2659,45 @@ public SpringValidator validator() {
 }
 ----
 
+You can also use the `BeanValidatingItemProcessor` to validate items annotated with
+the Bean Validation API (JSR-303) annotations. For example, given the following type `Person`:
+
+[source, java]
+----
+class Person {
+
+    @NotEmpty
+    private String name;
+
+    public Person(String name) {
+     this.name = name;
+    }
+
+    public String getName() {
+     return name;
+    }
+
+    public void setName(String name) {
+     this.name = name;
+    }
+
+}
+----
+
+you can validate items by declaring a `BeanValidatingItemProcessor` bean in your
+application context and register it as a processor in your chunk-oriented step:
+
+[source, java]
+----
+@Bean
+public BeanValidatingItemProcessor<Person> beanValidatingItemProcessor() throws Exception {
+    BeanValidatingItemProcessor<Person> beanValidatingItemProcessor = new BeanValidatingItemProcessor<>();
+    beanValidatingItemProcessor.setFilter(true);
+
+    return beanValidatingItemProcessor;
+}
+----
+
 [[process-indicator]]
 === Preventing State Persistence
 
--- a/spring-batch-docs/asciidoc/whatsnew.adoc
+++ b/spring-batch-docs/asciidoc/whatsnew.adoc
@@ -11,6 +11,7 @@ The Spring Batch 4.1 release adds the following features:
 * A new `@SpringBatchTest` annotation to simplify testing batch components
 * A new `@EnableBatchIntegration` annotation to simplify remote chunking configuration
 * A new `JsonItemReader` and `JsonFileItemWriter` to support the JSON format
+* Add support for validating items with the Bean Validation API
 
 [[whatsNewTesting]]
 === `@SpringBatchTest` Annotation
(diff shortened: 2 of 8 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/validator/BeanValidatingItemProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Create a new instance of {@link BeanValidatingItemProcessor}.
+	 * @param localValidatorFactoryBean used to configure the Bean Validation validator
+	 * @throws Exception If unable to configure the Bean Validation infrastructure
+	 */
+	public BeanValidatingItemProcessor(LocalValidatorFactoryBean localValidatorFactoryBean) throws Exception {
+		super();
+		Assert.notNull(localValidatorFactoryBean, "localValidatorFactoryBean must not be null");
+		localValidatorFactoryBean.afterPropertiesSet();
```
mminella: We shouldn't be calling this here.  That is the responsibility of the container since it's being injected.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/validator/BeanValidatingItemProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+		Validator validator = localValidatorFactoryBean.getValidator();
+		SpringValidatorAdapter springValidatorAdapter = new SpringValidatorAdapter(validator);
+		SpringValidator<T> springValidator = new SpringValidator<>();
+		springValidator.setValidator(springValidatorAdapter);
+		springValidator.afterPropertiesSet();
```
mminella: Since the `ValidatingItemProcessor` implements `IntializingBean`, this `afterPropertiesSet` should be called in an override of the parent's method to honor the lifecycle.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/validator/BeanValidatingItemProcessor.java
benas: What if the `LocalValidatorFactoryBean` is not injected? When I call the default constructor of `BeanValidatingItemProcessor`, the new instance of `LocalValidatorFactoryBean` is not managed by Spring and the `afterPropertiesSet` is not called. This `afterPropertiesSet` has a complete logic to configure the bean validation infrastructure with many fallbacks. BTW, if I don't call it, the tests fail with `java.lang.IllegalArgumentException: No target ValidatorFactory set` because one of the fallbacks is not configured anymore.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/validator/BeanValidatingItemProcessor.java
benas: fixed in d2a00f5

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/validator/BeanValidatingItemProcessor.java
mminella: The `InitializingBean#afterPropertiesSet()` method is a method that should be called by the container.  Saying it may not be called due to someone else doing the wrong thing means we'd never be able to rely on that being called and everywhere a bean gets injected we'd need to call it ourselves.  Unless we are creating the bean within the method so it cannot be called by the container (as we are just below this), we must assume that the method is being called for us.

benas:
@mminella I updated the PR as discussed and I added documentation in the `What's new` section as well as in the reference documentation.

spring-batch-docs/asciidoc/readersAndWriters.adoc
```diff
@@ -2630,6 +2630,46 @@ public SpringValidator validator() {
...
+you can validate items by declaring a `BeanValidatingItemProcessor` bean in your
+application context and register it as a processor in your chunk-oriented step:
+
+[source, java]
+----
+@Bean
+public BeanValidatingItemProcessor<Person> beanValidatingItemProcessor() throws Exception {
+    BeanValidatingItemProcessor<Person> beanValidatingItemProcessor = new BeanValidatingItemProcessor<>();
+    beanValidatingItemProcessor.setFilter(true);
+    beanValidatingItemProcessor.afterPropertiesSet();
```
mminella: In this case, you wouldn't need to call `afterPropertiesSet` since the container will do it for you.  I can fix this on merge.

spring-batch-docs/asciidoc/whatsnew.adoc
```diff
@@ -158,3 +159,48 @@ To add other libraries, you can implement the `JsonObjectReader` interface.
...
+you can validate items by declaring a `BeanValidatingItemProcessor` bean in your
+application context and register it as a processor in your chunk-oriented step:
+
+[source, java]
+----
+@Bean
+public BeanValidatingItemProcessor<Person> beanValidatingItemProcessor() throws Exception {
+	BeanValidatingItemProcessor<Person> beanValidatingItemProcessor = new BeanValidatingItemProcessor<>();
+	beanValidatingItemProcessor.setFilter(true);
+	beanValidatingItemProcessor.afterPropertiesSet();
```
mminella: See note from above.

benas:
rebased on latest master (and resolved conflicts) + removed the `afterPropertiesSet` calls.

PR #618, opened 2018-07-05
Add builders to simplify remote partitioning setup

benas:
This PR resolves [BATCH-2687](https://jira.spring.io/browse/BATCH-2687)


spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+		Assert.notNull(this.inputChannel, "An InputChannel must be provided");
+		Assert.notNull(this.jobExplorer, "A JobExplorer must be provided");
+
+		if (this.stepLocator == null) {
+			BeanFactoryStepLocator beanFactoryStepLocator = new BeanFactoryStepLocator();
+			beanFactoryStepLocator.setBeanFactory(this.beanFactory);
+			this.stepLocator = beanFactoryStepLocator;
+		}
+		if (this.outputChannel == null) {
+			this.outputChannel = new NullChannel();
```
mminella: A debug log message here indicating that the null channel has been configured and the master must poll for results may be a good idea.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Create a new {@link RemotePartitioningWorkerStepBuilderFactory}.
+	 * @param jobRepository the job repository to use
+	 * @param jobExplorer the job explorer to use
+	 * @param transactionManager the transaction manager to use
+	 */
+	public RemotePartitioningWorkerStepBuilderFactory(JobRepository jobRepository,
+														JobExplorer jobExplorer,
+														PlatformTransactionManager transactionManager) {
```
mminella: Odd indentation

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+public class RemotePartitioningWorkerStepBuilderFactory implements BeanFactoryAware {
+
+	private BeanFactory beanFactory;
+	private JobExplorer jobExplorer;
+	private JobRepository jobRepository;
+	private PlatformTransactionManager transactionManager;
```
mminella: Any reason the three variables above can't be `final`?

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+		this.stepLocator = stepLocator;
+		return this;
+	}
+
+	/**
+	 * Set the bean factory.
+	 * @param beanFactory the bean factory
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemotePartitioningWorkerStepBuilder beanFactory(BeanFactory beanFactory) {
```
mminella: Why are the required fields (jobExplorer, etc) not configured via the constructor?

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+	private static final long DEFAULT_TIMEOUT = 3600000L;
+
+	private MessagingTemplate messagingTemplate;
+	private MessageChannel inputChannel;
+	private MessageChannel outputChannel;
+	private JobExplorer jobExplorer;
+	private BeanFactory beanFactory;
+	private long pollInterval = DEFAULT_POLL_INTERVAL;
+	private long timeout = DEFAULT_TIMEOUT;
+
```
mminella: Why are the required values that are provided by the factory (`JobExplorer`, etc) not configured via the constructor?

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+	@Test
+	public void inputChannelMustNotBeNull() {
+		// given
+		this.expectedException.expect(IllegalArgumentException.class);
+		this.expectedException.expectMessage("inputChannel must not be null");
+
+		// when
+		Step step = new RemotePartitioningMasterStepBuilder("step")
+				.inputChannel(null)
+				.build();
```
mminella: Based on the code you shouldn't need to call `build` on all of these null check tests.

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+	@Test
+	public void timeoutMustNotBeGreaterThanZero() {
+		// given
+		this.expectedException.expect(IllegalArgumentException.class);
+		this.expectedException.expectMessage("The timeout must be greater than zero");
+
+		// when
+		Step step = new RemotePartitioningMasterStepBuilder("step")
+				.timeout(-1L)
```
mminella: Is there the option for no timeout?

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+		this.expectedException.expectMessage("When configuring a master step " +
+				"for remote partitioning using the RemotePartitioningMasterStepBuilder, " +
+				"the partition handler will be automatically set to an instance " +
+				"of MessageChannelPartitionHandler. The partition handler must " +
+				"not be provided in this case.");
+
+		// when
+		Step step = new RemotePartitioningMasterStepBuilder("step")
+				.partitionHandler(partitionHandler)
+				.build();
```
mminella: Shouldn't need to call build

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+		// then
+		// expected exception
+	}
+
+	@Test
+	public void testMasterStepCreation_whenPollingRepository() {
```
mminella: Why camel case and _?  I'd rather just keep the camel case only.

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderTest.java
```diff
@@ -0,0 +1,151 @@
...
+	public void inputChannelMustNotBeNull() {
+		// given
+		this.expectedException.expect(IllegalArgumentException.class);
+		this.expectedException.expectMessage("inputChannel must not be null");
+
+		// when
+		Step step = new RemotePartitioningWorkerStepBuilder("step")
+				.inputChannel(null)
+				.tasklet(this.tasklet)
+				.build();
```
mminella: See previous notes on the build calls

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,90 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
 
 You can find a complete example of a remote partitioning job
 link:$$https://github.com/spring-projects/spring-batch/tree/master/spring-batch-samples#remote-partitioning-sample$$[here].
+
+Starting from version 4.1, Spring Batch Integration introduces the `@EnableBatchIntegration`
+annotation that can be used to simplify a remote partitioning setup. This annotation provides
+two beans that can be autowired in the application context:
```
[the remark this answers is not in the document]
benas: updated

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,90 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
 
 You can find a complete example of a remote partitioning job
 link:$$https://github.com/spring-projects/spring-batch/tree/master/spring-batch-samples#remote-partitioning-sample$$[here].
+
+Starting from version 4.1, Spring Batch Integration introduces the `@EnableBatchIntegration`
```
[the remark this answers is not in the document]
benas: updated

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,90 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
...
+        @Autowired
+        private RemotePartitioningWorkerStepBuilderFactory workerStepBuilderFactory;
+
+        @Bean
+        public Step workerStep() {
+                 return this.workerStepBuilderFactory
+                    .get("workerStep")
+                    .inputChannel(incomingRequestsFromMaster())
+                    .outputChannel(outgoingRepliesToMaster())
+                    .tasklet(tasklet()) // or chunk(CHUNK_SIZE)
```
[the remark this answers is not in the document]
benas: I agree. changed to chunk option.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -23,24 +23,33 @@
...
+ *     setting up the job repository and transaction manager.</li>
  *     <li>{@link RemoteChunkingWorkerBuilder}: used to create the integration
- *     flow on the worker side.</li>
+ *     flow on the worker side of a remote chunking setup.</li>
+ *     <li>{@link RemotePartitioningMasterStepBuilderFactory}: used to create
+ *     a master step of a remote partitioning setup by automatically setting up
+ *     the job repository, job explorer, bean factory and transaction manager.</li>
+ *     <li>{@link RemotePartitioningWorkerStepBuilderFactory}: used to create
+ *     a worker step of a remote partitioning setup by automatically setting up
+ *     the job repository, job explorer, bean factory and transaction manager.</li>
```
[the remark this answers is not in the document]
benas: Correct, they don't *setup* any of these beans but they *do set* them. I updated the javadoc by removing the word `up`.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -75,15 +84,57 @@
...
+ *       		.build();
+ * 	}
+ *
+ * 	&#064;Bean
+ * 	public Step workerStep() {
+ *       	 return this.workerStepBuilderFactory
+ *       		.get("workerStep")
+ *       		.inputChannel(incomingRequestsFromMaster())
+ *       		.outputChannel(outgoingRepliesToMaster())
+ *       		.tasklet(tasklet()) // or chunk(CHUNK_SIZE)
```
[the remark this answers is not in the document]
benas: I agree. changed to chunk option.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+/**
+ * Builder for a master step in a remote partitioning setup. This builder creates and
+ * sets a {@link MessageChannelPartitionHandler} on the master step.
+ *
+ * <p>If no <code>messagingTemplate</code> is provided through
+ * {@link RemotePartitioningMasterStepBuilder#messagingTemplate(MessagingTemplate)},
+ * this builder will create one. The <code>outputChannel</code> set with
+ * {@link RemotePartitioningMasterStepBuilder#outputChannel(MessageChannel)} will be
+ * used as a default channel of the messaging template and will override any default
+ * channel already set on the messaging template.
```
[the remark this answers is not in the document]
benas: emphasize with a `strong` tag

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+		Assert.notNull(outputChannel, "outputChannel must not be null");
+		this.outputChannel = outputChannel;
+		return this;
+	}
+
+	/**
+	 * Set the {@link MessagingTemplate} to use to send data to workers.
+	 * <p><strong>The default destination of the messaging template will be
+	 * overridden by the output channel provided through
+	 * {@link RemotePartitioningMasterStepBuilder#outputChannel(MessageChannel)}</strong>.</p>
```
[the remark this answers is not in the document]
benas: I'm not sure I understand this point. Let's discuss it on Monday.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemotePartitioningMasterStepBuilder pollInterval(long pollInterval) {
+		Assert.isTrue(pollInterval > 0, "The poll interval must be greater than zero");
+		this.pollInterval = pollInterval;
+		return this;
+	}
+
+	/**
+	 * Set the timeout. Defaults to 1 hour.
```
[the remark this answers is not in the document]
benas: Changed to `-1L` (as in `org.springframework.batch.integration.partition.MessageChannelPartitionHandler#setTimeout` where it will endup)

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+			IntegrationFlowContext integrationFlowContext = this.beanFactory.getBean(IntegrationFlowContext.class);
+			integrationFlowContext.registration(standardIntegrationFlow)
+					.autoStartup(false)
+					.register();
+		}
+
+		try {
+			partitionHandler.afterPropertiesSet();
+			super.partitionHandler(partitionHandler);
+		} catch (Exception e) {
```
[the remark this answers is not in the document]
benas: fixed.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+	/**
+	 * Create a new {@link RemotePartitioningMasterStepBuilderFactory}.
+	 * @param jobRepository the job repository to use
+	 * @param jobExplorer the job explorer to use
+	 * @param transactionManager the transaction manager to use
+	 */
+	public RemotePartitioningMasterStepBuilderFactory(JobRepository jobRepository,
+														JobExplorer jobExplorer,
+														PlatformTransactionManager transactionManager) {
```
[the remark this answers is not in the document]
benas: fixed.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
+ */
+public class RemotePartitioningMasterStepBuilderFactory implements BeanFactoryAware {
+
+	private BeanFactory beanFactory;
+	private JobExplorer jobExplorer;
+	private JobRepository jobRepository;
+	private PlatformTransactionManager transactionManager;
```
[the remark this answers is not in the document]
benas: no reason. Marked as final.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+ * <ul>
+ *     <li>listens to {@link StepExecutionRequest}s coming from the master
+ *     on the input channel</li>
+ *     <li>invokes the {@link StepExecutionRequestHandler} to execute the worker
+ *     step for each incoming request. The worker step is located using the provided
+ *     {@link StepLocator}. If no {@link StepLocator} is provided, a {@link BeanFactoryStepLocator}
+ *     configured with the current {@link BeanFactory} will be used
+ *     <li>replies to the master on the output channel (when the master step is
+ *     configured to aggregate replies from workers). If no output channel
+ *     is provided, a {@link NullChannel} will be used</li>
```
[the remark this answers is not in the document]
benas: Added the clarification.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
benas: Good idea. Added a debug message.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
benas: fixed.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
benas: no reason. Marked as final.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
benas: To be consistent with other builders.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
benas: To be consistent with other builders.

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
benas: Correct (from the developer's point of view). But I believe tests should be written from the user's point of view (black box). In this case, trying to create a step using the builder.

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
benas: See previous comment on the build calls

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
benas: fixed.

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderTest.java
benas: See previous comment on the build calls

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
benas: Changed the default to `-1L` which means no timeout (like in `org.springframework.batch.integration.partition.MessageChannelPartitionHandler#setTimeout`)

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,92 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
...
 You can find a complete example of a remote partitioning job
 link:$$https://github.com/spring-projects/spring-batch/tree/master/spring-batch-samples#remote-partitioning-sample$$[here].
+
+The `@EnableBatchIntegration` annotation that can be used to simplify a remote
+ partitioning setup. This annotation provides two beans useful for remote partitioning:
+
+* `RemotePartitioningMasterStepBuilderFactory`: used to configure the master step
+* `RemotePartitioningWorkerStepBuilderFactory`: used to configure the worker step
+
+These APIs will take care of configuring a number of components as described in the following diagram:
```
artembilan: In fixes for Spring Integration docs I've seen that @Buzzardo changed all our future tenses to the present simple.
WDYT?

spring-batch-docs/asciidoc/whatsnew.adoc
```diff
@@ -124,6 +124,55 @@ that uses these new APIs in the
 link:$$https://github.com/spring-projects/spring-batch/tree/master/spring-batch-samples#remote-chunking-sample$$[samples module]
 as well as more details in the <<spring-batch-integration.adoc#remote-chunking,Spring Batch Integration>> chapter.
 
+Just like the remote chunking configuration simplification, this version also
+introduces new APIs to simplify a remote partitioning setup:
+`RemotePartitioningMasterStepBuilder` and `RemotePartitioningWorkerStepBuilder`.
+Those can be autowired in your configuration class if the
+`@EnableBatchIntegration` is present as shown in the following example:
+
+[source, java]
```
artembilan: I don't find it is pretty useful to mention the whole sample in the `What's New`: I might just not be interested in the future you are describing. So, the simple link to the target chapter with details is really enough.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
```diff
@@ -56,4 +63,16 @@ public RemoteChunkingMasterStepBuilderFactory remoteChunkingMasterStepBuilderFac
...
 	}
 
+	@Bean
+	public RemotePartitioningMasterStepBuilderFactory remotePartitioningMasterStepBuilderFactory() {
+		return new RemotePartitioningMasterStepBuilderFactory(this.jobRepository,
+				this.jobExplorer, this.transactionManager);
+	}
+
+	@Bean
+	public RemotePartitioningWorkerStepBuilderFactory remotePartitioningWorkerStepBuilderFactory() {
```
artembilan: Didn't you think about separating concerns for high-level representation?
I mean I might be interested only in `master` or only in `worker` in my current microservice. Therefore extra beans are just noise for me. 

Am I missing anything ?

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -75,15 +84,60 @@
...
+ * 	private RemotePartitioningWorkerStepBuilderFactory workerStepBuilderFactory;
+ *
+ * 	&#064;Bean
+ * 	public Step masterStep() {
+ *       	 return this.masterStepBuilderFactory
+ *       		.get("masterStep")
+ *       		.partitioner("workerStep", partitioner())
+ *       		.gridSize(GRID_SIZE)
+ *       		.outputChannel(outgoingRequestsToWorkers())
+ *       		.inputChannel(incomingRepliesFromWorkers())
```
artembilan: Sounds like would be better to name them as `request` and `reply` channels. 
I don't tell that this `master/worker` approach reminds me [Scatter-Gather](https://docs.spring.io/spring-integration/docs/5.0.6.RELEASE/reference/html/messaging-routing-chapter.html#scatter-gather) with an appropriate channel names 😄

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -75,15 +84,60 @@
...
+ *       		.build();
+ * 	}
+ *
+ * 	&#064;Bean
+ * 	public Step workerStep() {
+ *       	 return this.workerStepBuilderFactory
+ *       		.get("workerStep")
+ *       		.inputChannel(incomingRequestsFromMaster())
+ *       		.outputChannel(outgoingRepliesToMaster())
```
artembilan: The same `request/reply` pair is applied here as well.
This fully reflect `inbound/outbound` gateway abstraction in Spring Integration, I believe.
Or I just miss something here...

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.PollableChannel;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.util.Assert;
+
+/**
+ * Builder for a master step in a remote partitioning setup. This builder creates and
+ * sets a {@link MessageChannelPartitionHandler} on the master step.
+ *
+ * <p>If no <code>messagingTemplate</code> is provided through
```
artembilan: Actually `<code>` (or any other HTML-like tags) is deprecated in JavaDocs. Would be better to use `{@code }` and so on.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+ * {@link RemotePartitioningMasterStepBuilder#messagingTemplate(MessagingTemplate)},
+ * this builder will create one. The <code>outputChannel</code> set with
+ * {@link RemotePartitioningMasterStepBuilder#outputChannel(MessageChannel)} will be
+ * used as a default channel of the messaging template and will <strong>override any default
+ * channel already set on the messaging template</strong>.
+ *
+ * @since 4.1
+ * @author Mahmoud Ben Hassine
```
artembilan: I somehow believe that `@since` has to go after `@author`, but that's already our code style...
Although it really would be better to align all of them. May be via: https://github.com/spring-io/spring-javaformat/

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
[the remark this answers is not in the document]
artembilan: Well, if they don't supply each other, we can consider to reject the second if the first has been provided.
I mean make `outputChannel` and `messagingTemplate` as mutually exclusive.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+		this.timeout = timeout;
+		return this;
+	}
+
+	/**
+	 * Set the bean factory.
+	 * @param beanFactory the bean factory to use
+	 * @return this builder instance for fluent chaining
+	 */
+	public RemotePartitioningMasterStepBuilder beanFactory(BeanFactory beanFactory) {
```
artembilan: I think this simply is propagated from the `RemotePartitioningMasterStepBuilderFactory`, so should not be `public`. Isn't it?

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+		return super.build();
+	}
+
+	private boolean isPolling() {
+		return this.inputChannel == null;
+	}
+
+	@Override
+	public RemotePartitioningMasterStepBuilder repository(JobRepository jobRepository) {
```
artembilan: Yeah... In Spring Integration Java DSL to have an inheritance for builders and to avoid such a duplication we use a concretization via generic argument:
```
public class AggregatorSpec extends CorrelationHandlerSpec<AggregatorSpec, AggregatingMessageHandler> {
``` 
for example.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
```diff
@@ -0,0 +1,258 @@
...
+	private MessageChannel outputChannel;
+	private JobExplorer jobExplorer;
+	private StepLocator stepLocator;
+	private BeanFactory beanFactory;
+
+	/**
+	 * Initialize a step builder for a step with the given name.
+	 * @param name the name of the step
+	 */
+	public RemotePartitioningWorkerStepBuilder(String name) {
```
artembilan: Are we really allowed to create instances of this class directly, but not only via an appropriate `factory`?

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
```diff
@@ -0,0 +1,76 @@
...
+	 * repository, job explorer, bean factory and transaction manager.
+	 * @param name the name of the step
+	 * @return a {@link RemotePartitioningWorkerStepBuilder}
+	 */
+	public RemotePartitioningWorkerStepBuilder get(String name) {
+		return new RemotePartitioningWorkerStepBuilder(name)
+				.repository(this.jobRepository)
+				.jobExplorer(this.jobExplorer)
+				.beanFactory(this.beanFactory)
+				.transactionManager(this.transactionManager);
```
artembilan: I somehow think that all these options should not be `public` in target builder.
Or are you intended to allow to override factory-based injections?

spring-batch-integration/src/test/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilderTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.springframework.test.util.ReflectionTestUtils.getField;
+
+/**
+ * @author Mahmoud Ben Hassine
+ */
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {RemotePartitioningMasterStepBuilderTest.BatchConfiguration.class})
+public class RemotePartitioningMasterStepBuilderTest {
```
artembilan: Our pattern is always `Tests`.
Also consider to use `@DirtiesContext` if your application context uses some active components, e.g. pollers.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,92 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
 
 You can find a complete example of a remote partitioning job
 link:$$https://github.com/spring-projects/spring-batch/tree/master/spring-batch-samples#remote-partitioning-sample$$[here].
+
+The `@EnableBatchIntegration` annotation that can be used to simplify a remote
+ partitioning setup. This annotation provides two beans useful for remote partitioning:
```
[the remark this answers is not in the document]
benas: I agree, but I tried to be consistent will the current docs that wrap lines.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
benas: Fixed. Thanks!

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
benas: If you are interested only in the master or worker, then you can inject only beans related to the master or worker setup. Some people prefer declaring all beans in the same config class and use profiles to activate/deactivate some beans.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
benas: Well if you ask me `WDYT?`, I need to tell you what I think right? 😜 But I've got the point, fair enough.

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
benas: Yes you are allowed to use the builder directly, but you have to set the properties (that the factory sets) manually.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
artembilan: Right, I can inject, but my application context is still polluted with all of them.
The problem that you don't give me any preferences. There is only one annotation to add on my `@Configuration` and both infrastructure are going to be available in my application context.

Not sure what you mean about profiles, but this is already out of my control since we are talking here about imported configuration from the annotation without any profiling.

That might not be so bad since this factories are not big objects at all. I might be a bit confused eventually when I have to provide some infrastructure bean which is required by the `master`, when I am only in the `worker` or vise versa.

Just my feeling with the hope you get some food to think in the future.
The current solution might be just OK as is. Just be prepared for some similar feedback from the community! 😄

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
benas: Yes it is intended to make them overridable.

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/BatchIntegrationConfiguration.java
benas: I see your point. What do you suggest to separate the concerns? Two annotations? An additional attribute in the current annotation? Any other idea is welcome!

For the record, `@EnableBatchProcessing` adds 9 beans to the application context and I might not be interested in all of them. We took the same approach and we thought it's probably a good idea to centralize the 4 beans related to integration under the `@EnableBatchIntegration` config.

artembilan: That's fine with me. I just want to warn you about potential requests from the community.
Or even better - document the situation since we are already aware of the concern.
That's all, nothing more. Even `@EnableIntegration` provides a lot of beans. My only concern here that there is a clear separation of `master` and `worker` responsibility, so that might be a bit confusing here, unlike in the Integration when all the stuff is enabled by default, but that doesn't say that it is only for `client` or `server` or similar separation of responsibility.

Let's stop from here and go on with your current solution!

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilder.java
artembilan: OK. So, if we can instantiate builders directly without factory, then my previous comment about removing `public` modifier is wrong and end-user should be able to provide all the necessary options for the builder.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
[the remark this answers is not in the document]
mminella: We migrated to this format before the sentence per line rule was decided on.  I'd want @Buzzardo to rule on weather or not it's worth going through again.  I wouldn't hold this PR back for that though.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,92 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
...
+
+        @Autowired
+        private RemotePartitioningMasterStepBuilderFactory masterStepBuilderFactory;
+
+        @Bean
+        public Step masterStep() {
+                 return this.masterStepBuilderFactory
+                    .get("masterStep")
+                    .partitioner("workerStep", partitioner())
+                    .gridSize(GRID_SIZE)
```
mminella: I don't see the `GRID_SIZE` variable defined.  Probably should just hard code a value for the docs.  I can fix on merge if you're ok with it.

spring-batch-docs/asciidoc/spring-batch-integration.adoc
```diff
@@ -1261,3 +1261,92 @@ You must also ensure that the partition `handler` attribute maps to the `partiti
...
+        @Autowired
+        private RemotePartitioningWorkerStepBuilderFactory workerStepBuilderFactory;
+
+        @Bean
+        public Step workerStep() {
+                 return this.workerStepBuilderFactory
+                    .get("workerStep")
+                    .inputChannel(incomingRequestsFromMaster())
+                    .outputChannel(outgoingRepliesToMaster())
+                    .chunk(CHUNK_SIZE)
```
mminella: Same note as above with GRID_SIZE...

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -75,15 +84,60 @@
...
+ *
+ * 	&#064;Autowired
+ * 	private RemotePartitioningWorkerStepBuilderFactory workerStepBuilderFactory;
+ *
+ * 	&#064;Bean
+ * 	public Step masterStep() {
+ *       	 return this.masterStepBuilderFactory
+ *       		.get("masterStep")
+ *       		.partitioner("workerStep", partitioner())
+ *       		.gridSize(GRID_SIZE)
```
mminella: See previous notes about constant definitions

spring-batch-integration/src/main/java/org/springframework/batch/integration/config/annotation/EnableBatchIntegration.java
```diff
@@ -75,15 +84,60 @@
...
+ *       		.build();
+ * 	}
+ *
+ * 	&#064;Bean
+ * 	public Step workerStep() {
+ *       	 return this.workerStepBuilderFactory
+ *       		.get("workerStep")
+ *       		.inputChannel(incomingRequestsFromMaster())
+ *       		.outputChannel(outgoingRepliesToMaster())
+ *       		.chunk(CHUNK_SIZE)
```
mminella: See previous notes about constant definitions

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningMasterStepBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+		this.timeout = timeout;
+		return this;
+	}
+
+	/**
+	 * Set the bean factory.
+	 * @param beanFactory the bean factory to use
+	 * @return this builder instance for fluent chaining
+	 */
+	RemotePartitioningMasterStepBuilder beanFactory(BeanFactory beanFactory) {
```
mminella: Is there a reason this is package private?  On a fluent builder this feels odd...

spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/RemotePartitioningWorkerStepBuilderFactory.java
mminella: If we're providing a factory bean for this, we probably should make users use it for a consistency thing...

mminella:
Rebased, squashed, and merged as 39676a45c48313dfae1905f6f9769a9c6ab32e90

PR #623, opened 2018-07-17
Add JSR-305 annotations to public APIs (integration module)

benas:


spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/MessageChannelPartitionHandler.java
```diff
@@ -213,7 +213,7 @@ public void setReplyChannel(PollableChannel replyChannel) {
 		final Set<StepExecution> split = stepExecutionSplitter.split(masterStepExecution, gridSize);
 
 		if(CollectionUtils.isEmpty(split)) {
-			return null;
+			return split;
```
mminella: This caused a test to fail.  `MessageChannelPartitionHandlerTests#testNoPartitions()`

mminella:
Fixed, rebased, and merged

PR #625, opened 2018-08-08
[BATCH-2741] Fix issue - job parameter overwrites

jojoldu:
Hello?
I recently updated SpringBoot 2.0 with Spring Batch 4.0.
But I think there is a bug.
If the job fails, I keep getting the value of the failed Job Parameter even if I change the value of the Job Parameter with the same key next time.
In 1.5.x, I confirmed that it did not happen, and when I checked the code, it showed a bug in the code and sent a PR

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/JobParametersBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/JobParametersBuilder.java
@@ -19,7 +19,13 @@ package org.springframework.batch.core;
 import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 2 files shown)


benas:
As explained in [BATCH-2711](https://jira.spring.io/browse/BATCH-2711), the method `JobParametersBuilder#getNextJobParameters` should not deal with restartability anymore after it has been moved to Spring Batch in order to be consistent with `CommandLineJobRunner` and `JobOperator`. So the fix in this PR cannot be applied since the method will *still* deal with restartability. The issue is deeper than just fixing the `merge` method as in this PR, it should be fixed in both Spring Batch and Spring Boot.

Thank you @jojoldu for your contribution anyway!

PR #631, opened 2018-08-22
Add convenience methods in FlatFileItemWriterBuilder to generate delimited/fixedLength files

benas:
This PR resolves [BATCH-2696](https://jira.spring.io/browse/BATCH-2696).

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
@@ -18,6 +18,7 @@ package org.springframework.batch.item.file.builder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 
 import org.springframework.batch.item.file.FlatFileFooterCallback;
 import org.springframework.batch.item.file.FlatFileHeaderCallback;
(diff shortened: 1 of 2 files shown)


benas:
Added test in 4da50ba42df0da1d5b2f810cb6c33531c7500ea1 to cover generating a delimited file with the default delimiter.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
```diff
@@ -234,14 +249,245 @@
...
+		 * is set via {@link DelimitedBuilder#fieldExtractor(FieldExtractor)}.
+		 *
+		 * @param names names of each field
+		 * @return The parent {@link FlatFileItemWriterBuilder}
+		 * @see BeanWrapperFieldExtractor#setNames(String[])
+		 */
+		public FlatFileItemWriterBuilder<T> names(String[] names) {
+			this.names.addAll(Arrays.asList(names));
+			return this.parent;
+		}
```
mminella: It would be nice to have a List option as well...

benas: I took the same approach as the `FlatFileItemReaderBuilder` to be consistent.

We can add this for both reader and writer builders in another step.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
```diff
@@ -234,14 +249,245 @@
...
+		 * is set via {@link FormattedBuilder#fieldExtractor(FieldExtractor)}.
+		 *
+		 * @param names names of each field
+		 * @return The parent {@link FlatFileItemWriterBuilder}
+		 * @see BeanWrapperFieldExtractor#setNames(String[])
+		 */
+		public FlatFileItemWriterBuilder<T> names(String[] names) {
+			this.names.addAll(Arrays.asList(names));
+			return this.parent;
+		}
```
mminella: It would be nice to have a vargs option as well...

PR #634, opened 2018-08-31
Allow the transaction manager to be overridden in DefaultBatchConfigurer

benas:
Before this commit, it was not possible to override the transaction manager by subclassing `DefaultBatchConfigurer` and overriding the `getTransactionManager` method.

This commit uses the `getTransactionManager` method in the `initialize` method in order to take into account the transaction manager provided by the user.

Resolves [BATCH-2294](https://jira.spring.io/browse/BATCH-2294)

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/configuration/annotation/TransactionManagerConfigurationWithoutBatchConfigurerTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/configuration/annotation/TransactionManagerConfigurationWithoutBatchConfigurerTests.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


spring-batch-core/src/test/java/org/springframework/batch/core/configuration/annotation/TransactionManagerConfigurationWithoutBatchConfigurerTests.java
```diff
@@ -0,0 +1,147 @@
...
+import org.springframework.transaction.PlatformTransactionManager;
+
+/**
+ * @author Mahmoud Ben Hassine
+ */
+public class TransactionManagerConfigurationWithoutBatchConfigurerTests extends TransactionManagerConfigurationTests {
+
+	@Test
+	public void testConfigurationWithNoDataSourceAndNoTransactionManager() throws Exception {
+		AnnotationConfigApplicationContext applicationContext = new AnnotationConfigApplicationContext(BatchConfigurationWithNoDataSourceAndNoTransactionManager.class);
```
mminella: Nit pick...no reason to code to the implementation in these methods

benas: I agree. Fixed.

mminella:
LGTM.  Rebased and merged as a68d70043ae6c390a6a877cd435ccafec6cd9d62.

nightswimmings:
Isn't this solution problematic as long as SimpleBatchConfiguration keeps the 
``` 	
@Bean
	public PlatformTransactionManager transactionManager() throws Exception {
```
part?
Why not something like in the analogous SimpleTaskAutoConfiguration with a @ConditionalOnMissingBean?

benas:
`@ConditionalOnMissingBean` is from Spring Boot. SCT can depend on boot, but we can't depend on boot in Spring Batch.

PR #636, opened 2018-09-06
Fix JsonFileItemWriter for multiple write's

gd-estrepetov:
Resolves BATCH-2749

### DIFF
--- a/spring-batch-infrastructure-tests/src/test/java/org/springframework/batch/item/json/JsonFileItemWriterFunctionalTests.java
+++ b/spring-batch-infrastructure-tests/src/test/java/org/springframework/batch/item/json/JsonFileItemWriterFunctionalTests.java
@@ -59,6 +59,8 @@ public abstract class JsonFileItemWriterFunctionalTests {
 	private ExecutionContext executionContext;
 	private Trade trade1 = new Trade("123", 5, new BigDecimal("10.5"), "foo");
 	private Trade trade2 = new Trade("456", 10, new BigDecimal("20.5"), "bar");
+	private Trade trade3 = new Trade("789", 15, new BigDecimal("30.5"), "foobar");
+	private Trade trade4 = new Trade("987", 20, new BigDecimal("40.5"), "barfoo");
 
 	protected abstract JsonObjectMarshaller<Trade> getJsonObjectMarshaller();
 	protected abstract JsonObjectMarshaller<Trade> getJsonObjectMarshallerWithPrettyPrint();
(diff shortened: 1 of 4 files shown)


benas:
Thank you for reporting this issue and for opening a PR! 👍 

The problem is that the snippet that adds a separator if a previous chunk has been written is inside the loop. So I would not add a boolean (`alreadyAddedLineSeparator`) but instead we need to move the snippet outside the loop:

```diff
@Override
public String doWrite(List<? extends T> items) {
	StringBuilder lines = new StringBuilder();
+	if (!items.isEmpty() && state.getLinesWritten() > 0) {
+		lines.append(JSON_OBJECT_SEPARATOR).append(this.lineSeparator);
+	}
	Iterator<? extends T> iterator = items.iterator();
	while (iterator.hasNext()) {
-		if (iterator.hasNext() && state.getLinesWritten() > 0) {
-			lines.append(JSON_OBJECT_SEPARATOR).append(this.lineSeparator);
-		}
		T item = iterator.next();
		lines.append(' ').append(this.jsonObjectMarshaller.marshal(item));
		if (iterator.hasNext()) {
			lines.append(JSON_OBJECT_SEPARATOR).append(this.lineSeparator);
		}
	}
	return lines.toString();
}
```

There is another test that needs to be updated (in addition to the ones you added in this PR): `JsonSupportIntegrationTests`. For this one, just adding a line in the `trades.json` file is enough. For example:

```diff
[
 {"isin":"123","quantity":5,"price":10.5,"customer":"foo","id":1,"version":0},
 {"isin":"456","quantity":10,"price":20.5,"customer":"bar","id":2,"version":0},
- {"isin":"789","quantity":15,"price":30.5,"customer":"baz","id":3,"version":0}
+ {"isin":"789","quantity":15,"price":30.5,"customer":"baz","id":3,"version":0},
+ {"isin":"100","quantity":20,"price":35.5,"customer":"bat","id":4,"version":0}
]
```

Could you please update the PR with these changes? Thank you upfront!

gd-estrepetov:
Fixed your comments, by the way it should be at least 4 objects to reproduce the issue, if size of chunk is 2.

benas:
Thank you for updating the PR! LGTM.

In my previous comment, I meant the `spring-batch-samples/src/test/resources/org/springframework/batch/item/json/trades.json` file needs also to be updated with 4 items. I'll do it when rebasing this PR.

mminella:
LGTM.  Rebase, squashed, and merged as 7a3248cabf5eb59def5a217e6d6113c5617cdf70.  Thanks for the catch and the fix!

PR #641, opened 2018-09-16
Corrected the links and titles as per the Github markdown syntax

Ganapathi004:
Notes:
For the section Tasklet Job, there is no row in the table. For the IO sample job, there seems to be another (2nd ??) table and no section for it. In the second table, the rows delimited, jpa, multiRecordType and multiResource, there are no detailed sections.


mminella:
LGTM.  Rebased, squashed, and merged as 50625f50547106569f06b4f35e365603eb8b751b.  Thanks for the updates!

PR #643, opened 2018-09-26
 [BATCH-2760] Failed JobExecution due to unavailable TaskExecutor leaves End Time unpopulated

dimitrisli:
On instances when the `taskExecutor` is not picking up the submitted **Job Execution**, with this fix we are explicitly setting the **End Time** of the JobExecution in order not to give false positives via the `JobExplorer#findRunningJobExecutions()`. This last method is judging on running **Job Executions** based on **End Time** being null.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobLauncher.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobLauncher.java
@@ -165,6 +165,7 @@ public class SimpleJobLauncher implements JobLauncher, InitializingBean {
 		}
 		catch (TaskRejectedException e) {
 			jobExecution.upgradeStatus(BatchStatus.FAILED);
+			jobExecution.setStartTime(new Date(System.currentTimeMillis()));
 			jobExecution.setEndTime(new Date(System.currentTimeMillis()));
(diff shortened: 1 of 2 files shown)


dimitrisli:
Setting `new Date(System.currentTimeMillis())` following the pattern used in `JobExecution#lastUpdated()` calls within the project as well as `JobExecution#createTime` field default.

dimitrisli:
Jira ticket: https://jira.spring.io/browse/BATCH-2760

dimitrisli:
I've added: 
1. setting the **Start Time** as well as I figured it's also not getting set
2. unit test for this fix mocking the `TaskRejectedException` getting thrown from the `taskExecutor`.

mminella:
This PR implies that the job ran when in fact it didn't.  Personally, I'd rather see us fix the logic in the places where we evaluate if a job is running to exclude those that haven't started in the first place.  We can continue the discussion for this in the Jira issue BATCH-2675.  I'm going to close this now until we determine a best course.  We can always re-open it if we agree that this is the right path.

dimitrisli:
@mminella Thanks for the review. I'm continuing the discussion in BATCH-2675 with an alternative approach. thanks!

dimitrisli:
As a continuation of the above discussion in the comment section of: https://jira.spring.io/browse/BATCH-2675, I've raised a new PR: https://github.com/spring-projects/spring-batch/pull/659

PR #661, opened 2018-10-25
Return null instead of 0 when a Long/Double job parameter is absent

benas:
This PR resolves [BATCH-2385](https://jira.spring.io/browse/BATCH-2385):

* It makes the `JobParameters#getLong` and `JobParameters#getDouble` methods return `null` instead of 0 when the provided key is absent (same behaviour as `JobParameters#getString` which returns `null` when the key is absent and not an empty String). Indeed, with the current implementation, asking for a parameter and getting 0 back makes it impossible to know if the parameter was actually set to the value 0 or if the parameter is absent.
* It deprecates `JobParameters#getLong(String key, long defaultValue)` in favor of a new equivalent method with a wrapper type for the default value `JobParameters#getLong(String key, Long defaultValue)` in order to make it possible to specify `null` as a default value.
* It adds `@Nullable` where appropriate

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2013 the original author or authors.
+ * Copyright 2006-2018 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


benas:
Rebased and merged as 9e02e935ce4a599e8c113d85f1927a16a1a7d5c8.

PR #662, opened 2018-10-26
BATCH-2765: Kafka ItemWriter

mouellet:
This commit adds an ItemWriter implementation for Apache Kafka using
a KafkaTemplate with default topic set. Extending KeyValueItemWriter,
the writer requires a Converter to map the Kafka ProducerRecord's key
from the item.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ allprojects {
 		springDataMongodbVersion = '2.2.0.M3'
 		springDataNeo4jVersion = '5.2.0.M3'
 		springIntegrationVersion = '5.2.0.M1'
-		springKafkaVersion = '2.1.10.RELEASE'
+		springKafkaVersion = '2.2.5.RELEASE'
 		springLdapVersion = '2.3.2.RELEASE'
 
 		activemqVersion = '5.15.9'
--- a/build.gradle
+++ b/build.gradle
@@ -56,6 +56,7 @@ allprojects {
 		springDataMongodbVersion = '2.2.0.M3'
 		springDataNeo4jVersion = '5.2.0.M3'
 		springIntegrationVersion = '5.2.0.M1'
+		springKafkaVersion = '2.1.10.RELEASE'
 		springLdapVersion = '2.3.2.RELEASE'
 
 		activemqVersion = '5.15.9'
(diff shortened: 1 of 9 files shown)


benas:
That's great! Thank you! We will consider this PR when working on [BATCH-2765](https://jira.spring.io/browse/BATCH-2765)

JKerwood:
The Kafka send is async and the future isn't handled.  If the Kafka send fails it will be unknown to the application.

mouellet:
> The Kafka send is async and the future isn't handled. If the Kafka send fails it will be unknown to the application.

The writer is using a `KafkaTemplate` that takes care of that. One could add a `ProducerListener` to that `KafkaTemplate` and maybe throw an exception on failure to stop the process..

@benas, is that sufficient or should the writer manage that itself?

benas:
> @benas, is that sufficient or should the writer manage that itself?

Since the writer uses the `KafkaTemplate`, it is not its concern to poll the future for completion, inspect the result when available and throw an exception in case of errors, and do that **for each item in the chunk** inside the `write` method.

> One could add a ProducerListener to that KafkaTemplate and maybe throw an exception on failure to stop the process..

That is an option, but is not the way to go IMO. Error handling should be consistent with other messaging writers (JMS, AMQP), aka using the appropriate transaction manager. In this case, it should be a [KafkaTransactionManager](https://docs.spring.io/spring-kafka/docs/2.2.5.RELEASE/reference/#using-kafkatransactionmanager). As mentioned in Spring Kafka's docs, both the transaction manager and the kafka template should be configured with the same transaction capable `ProducerFactory`. This will ensure any success/failure in the producer's send operation (behind the template) will be correctly committed/rolled back by the kafka transaction manager. Here is the relevent section from Spring Kafka's docs:

```
If a transaction is active, any KafkaTemplate operations performed
within the scope of the transaction use the transaction’s Producer. 
The manager commits or rolls back the transaction, depending on success or failure.
You must configure the KafkaTemplate to use the same ProducerFactory
as the transaction manager.
```

This PR LGTM. Rebased and merged as 06556dc97e17f9d1ef24c1245b4f84e25ab0cfa9.

Thank you @mouellet for your contribution!

PR #664, opened 2018-10-29
BATCH-2764: Add KafkaItemReader

mouellet:
Single-thread consumer with manually assigned list of partitions and offsets store outside of Kafka. Supports:
 * Retry/restart, topic-partition offsets saved in ExecutionContext.
 * Uses 'max.poll.records' for chunking/paging.
 * Remote chunking, manual topic-partition assignment.


benas:
@mouellet Thank you for updating this PR! FYI, kafka support has been planned for the next version 4.2. We will let you know once we start working on your PRs. Keep tuned.

mouellet:
@benas that's fine, this is not yet ready to review and I will updating it again soon! The Kafka Consumer is not trivial and I'd like the reader to cover a simple use cases of using batch processing over stream processing. I'll remove the WIP once I'm done!

benas:
Hi @mouellet ,

> I'll remove the WIP once I'm done!

Is this PR still in progress? We started our work on Kafka support and I wanted to know if you still have some work to do on the kafka item reader before I start reviewing/testing it.

Please rebase your changes on the latest master and upgrade `springKafkaVersion` to 2.2.5.RELEASE.

mouellet:
Hi @benas,

Yes there is still a little work to do, I've put that on hold because I needed some feedback on design... I've started using an [OffsetsProvider](spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/support/OffsetsProvider.java) strategy so that the Reader could be re-runnable and support:
- Seeking to beginning offsets [(BeginningOffsetsProvider)](spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/support/BeginningOffsetsProvider.java).
- Seeking to offsets based on timestamps ([TimestampOffsetsProvider](spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/support/TimestampOffsetsProvider.java)).
- Seeking to saved offsets in ExecutionContext after failed execution.
- Seeking to Kafka committed offsets.
- Seeking to provided or externally stored offsets.

What do you think of those requirements?

benas:
@mouellet Thank you for the updates. 

> What do you think of those requirements?

The only strategy we need for the moment is "Seeking to saved offsets in ExecutionContext after failed execution" to be consistent with other readers. So there is actually no need for the `OffsetsProvider` concept and its implementations.

Here are my thoughts after a first review of the PR:

* The reader should accept only one topic to read data from, otherwise we could read different types of records which is not compatible with the contract of an item reader. However, we can accept multiple partitions for the same topic. As a user, I should provide the config properties, a topic name and a list of partitions. So a possible signature of the constructor would be `KafkaItemReader(Properties consumerProperties, String topicName, Integer... partitions)`. There is no need to expose the `ConsumerFactory` API, we can create a kafka consumer in the constructor with `new KafkaConsumer<>(consumerProperties)`.
* The fact that the `jumpToItem` method is empty is a smell. With kafka, we don't need to use the `read.count` value and jump to a given item a restart scenario, we can use offsets for that. So the `KafkaItemReader` should not extend `AbstractItemCountingItemStreamItemReader`, it should rather extend `AbstractItemStreamItemReader`.
* We don't need the `assignTimeoutProvider` and the `AtomicBoolean assigned`. It is enough to have a configurable poll duration with a default value.

Does this make sense? If you agree, please the PR accordingly and I will review it asap.

Best regards,
Mahmoud

benas:
Hi @mouellet ,

We are planning to release Kafka support in v4.2.0.M2 which is planned for next Monday (May 13th, 2019). If you can't update the PR before this date, I can merge it and fix the couple of points discussed in my previous comment if you agree on them. Of course all credit goes to you for this great contribution! Wdyt?

In all cases, this will be a first iteration for Kafka support and we can improve things in the next milestones.

Best regards,
Mahmoud

mouellet:
Hi @benas, 

I'm really sorry I won't have time to update the PR for next Monday... But I completely agree with all your comments and looking forward to collaborate with you again in the future!

Thanks,
Mathieu

benas:
Rebased, squashed and merged as 77c9b74673ada1e485c563cf80d1f136d07bd334.

I updated the code as discussed previously in b19998a40ad02f368d64816193ae46c63fe1ad30.

@mouellet Thank you for your contribution!

benas:
@mouellet v4.2.0.M2 has been released and I took the opportunity to thank you in the [announcement blog post](https://spring.io/blog/2019/05/13/spring-batch-4-2-0-m2-available-now). FYI, Josh has made a Spring Tips installment about it [here](https://twitter.com/SpringTipsLive/status/1128523274521055233).

PR #665, opened 2018-10-30
[BATCH-2762] Introducing SynchronizedItemStreamWriter

dimitrisli:
Also introducing SynchronizedItemStreamWriterBuilder. Multi-threaded
Tests have been that additionally prove that currently ItemStreamWriter
is not thread-safe.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
@@ -28,6 +28,8 @@ import java.util.List;
  * {@link SynchronizedItemStreamWriter#write write()} method
  *
  * @author Dimitrios Liapis
+ *
+ * @param <T> type of object being written
  */
 public class SynchronizedItemStreamWriter<T> implements ItemStreamWriter<T>, InitializingBean {
 
@@ -62,6 +64,6 @@ public class SynchronizedItemStreamWriter<T> implements ItemStreamWriter<T>, Ini
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		Assert.notNull(this.delegate, "A delegate item reader is required");
+		Assert.notNull(this.delegate, "A delegate item writer is required");
 	}
 }
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


dimitrisli:
The multi-threaded tests are inspired by the construct in [SynchronizedItemStreamReaderTests](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/support/SynchronizedItemStreamReaderTests.java) but it follows slightly different approach:

- using and atomic integer to track access to the write method
- a CountDownLatch to get all the threads ready
- assertion within the write method (closely paired with the atomic integer above)

Also realised that the existing [SynchronizedItemStreamReaderTests](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/support/SynchronizedItemStreamReaderTests.java) is not actually properly working as the threads are not being started correctly - [they are just invoking directly the run() method](https://github.com/spring-projects/spring-batch/blob/547533fab072289c062916e51c589de36ea3dfe2/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/support/SynchronizedItemStreamReaderTests.java#L122). I can fix this on a separate PR.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	@Override
+	public void close() throws ItemStreamException {
+		this.delegate.close();
+	}
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		Assert.notNull(this.delegate, "A delegate item reader is required");
```
benas, line 65: The message here should be "A delegate item writer is required".

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/SynchronizedItemStreamWriter.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.util.Assert;
+
+import java.util.List;
+
+/**
+ * An {@link ItemStreamWriter} decorator with a synchronized
+ * {@link SynchronizedItemStreamWriter#write write()} method
+ *
+ * @author Dimitrios Liapis
+ */
```
benas, line 31: Please add Javadoc for `@param <T>`.

benas:
@dimitrisli Thank you for this PR! The synchronized item writer and its builder LGTM. However, when I see the tests, it feels like we are trying to test the `synchronized` modifier of Java. I would not bother testing that only one thread is calling the method at a time. I really trust the JVM here.

I think calling the synchronized writer and make sure the delegate was called is already ok. I know you said you took inspiration from `SynchronizedItemStreamReaderTests`, but those tests could be simplified as well. I will give you more details on that in your other PR #666.

I added a couple of minor inline comments, if you can address them the PR should be good to merge. Thank you upfront.

benas:
Thank you for the updates @dimitrisli !

>  However, when I see the tests, it feels like we are trying to test the synchronized modifier of Java. I would not bother testing that only one thread is calling the method at a time. I really trust the JVM here. I think calling the synchronized writer and make sure the delegate was called is already ok.

Do you agree on simplifying tests? I did not request a change for that (I can do it when merging your PR unless you want to push a second update), but at least I wanted to know if you agree on that.

dimitrisli:
@benas totally agree. I've removed the concurrency tooling and focused on the delegate pass-through methods calling

benas:
@dimitrisli Great! LGTM now. Rebased, squashed and merged as 117b7d5f2d321dfaa1b74e5c57d0276e8154b407. Please note that I applied some minor changes (update year in license headers, update javadoc, etc, see 9fbd3daa54ba4e7d131caa117cd04ef5ab1abb09).

PR #667, opened 2018-11-05
[BATCH-2768] Correcting javadoc typo PartitionStepBuilder#gridSize()

dimitrisli:


benas:
LGTM. rebased and merged as e707a571571d7a7fc7371604da0502c40257e7b0. I also added the author name to the modified file.

> To be honest I don't think I'm getting the in the first attempt part

It means that in a subsequent run, the grid size will not be necessarily the number of partitions that will be re-run as only failed partitions are retried. The grid size is a hint to the partitioner that is used only at partitioning time (the first attempt). I hope this is clearer now.

Thank you for the PR!

PR #668, opened 2018-11-08
[BATCH-DOC]fix wrong Java Configuration for In-Memory JobRepository

clfsoft:


pivotal-issuemaster:
@clfsoft Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=668)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=668) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@clfsoft Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=668)!

benas:
@dimitrisli Thank you for the follow up! The key sentence is: `For this reason, Spring batch provides an in-memory Map version of the job repository`. Moreover, the characteristics described in the paragraph are also those of the Map based job repository (`It also cannot guarantee that two job instances with the same parameters are launched simultaneously, and is not suitable for use in a multi-threaded Job, ...)` which is not the case when using the jdbc based `JobRepositoryFactoryBean` with an in-memory datasource. So the current docs are incorrect and should show the `MapJobRepositoryFactoryBean` instead.

> I think if you see the Java Configuration side-by-side with its XML counterpart it makes sense as it is currently.

When I set the XML toggle, here is the snippet shown:

```xml
<bean id="jobRepository"
  class="org.springframework.batch.core.repository.support.MapJobRepositoryFactoryBean">
    <property name="transactionManager" ref="transactionManager"/>
</bean>
```

So for me, it is the `MapJobRepositoryFactoryBean` that should be shown in the Java config snippet to be consistent.

benas:
LGTM. rebased and merged as f90198fef8e0a9dd1b5445c3c691cefcae515bbf.

Thank you @clfsoft for your contribution!

PR #671, opened 2018-11-28
Create a new method DefaultFieldSet#read in favor of DefaultFieldSet#readAndTrim

benas:
Before this commit, making the `DefaultFieldSet` to not trim values was possible by overriding the `readAndTrim` method to call `readRawString`, something like:

```
@Override
protected String readAndTrim(int index) {
	return super.readRawString(index);
}
```

This is confusing since we are calling a method that is supposed to trim values while it actually returns raw values.

This commit introduces a new method called `read` which trims values by default (to preserve the current behaviour as documented in `DefaultFieldSet`). This new method can be then overridden like:

```
@Override
protected String read(int index) {
	return super.readRawString(index);
}
```

to not trim values. This change is backward compatible and provides and better naming.

Resolves [BATCH-2404](https://jira.spring.io/browse/BATCH-2404)

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DefaultFieldSet.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DefaultFieldSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2007 the original author or authors.
+ * Copyright 2006-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


benas:
After discussing this PR with @mminella , the new name `read` is less explicit than `readAndTrim` and could be confusing. Extending `DefaultFieldSet` to override `readAndTrim` and returning the extension using a custom `FieldSetFactory` is a bit hacky (as shown in [this example](https://github.com/benas/sandbox/blob/master/batch2404/src/main/java/com/example/demo/MyJob.java#L72-L111)), so probably the method should not have been `protected` in the first place to avoid workarounds like those presented in [BATCH-2637](https://jira.spring.io/browse/BATCH-2637).

PR #673, opened 2018-12-11
Add assertion that a serializer was set in the JdbcExecutionContextDao

benas:
Resolves [BATCH-2779](https://jira.spring.io/browse/BATCH-2779)


FadeHub:
Does spring batch support multithreaded restart?

benas:
@FadeHub I don't understand why this question is asked here, but you can find the answer in the following SO threads:

* [Why does making ItemReader thread-safe leads us to loosing restartability?](https://stackoverflow.com/questions/19971973/spring-batch-why-does-making-itemreader-thread-safe-leads-us-to-loosing-restar)
* [Restarting a spring batch job that is run by multiple threads](https://stackoverflow.com/questions/33427208/restarting-a-spring-batch-job-that-is-run-by-multiple-threads)

benas:
rebased and merged as 7ca11f2f998e5e46b35ec7a56fcd0e62b572b667.

PR #676, opened 2019-01-11
BATCH-2783: Update method to truly return a unique job parameter through a UUID

saikatbhadra:
We noticed that integration tests for a batch job (with many tests) were failing intermittently.  After debugging, I found that JobLauncherTestUtils.launchJob was occasionally relaunching the same job instance. After looking at the code, I saw that a random number generator was being used which does not guarantee that the number generate would be unique. I believe a UUID would be more appropriate as strongly guarantees that an ID does not repeat (except in highly exceptional cases). This change strongly assures that the JobParameter is unique every time getUniqueJobParameters() is called.

### DIFF
--- a/spring-batch-test/src/main/java/org/springframework/batch/test/JobLauncherTestUtils.java
+++ b/spring-batch-test/src/main/java/org/springframework/batch/test/JobLauncherTestUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.batch.test;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -155,12 +156,12 @@ public class JobLauncherTestUtils {
 	}
 
 	/**
-	 * @return a new JobParameters object containing only a parameter for the
-	 * current timestamp, to ensure that the job instance will be unique.
+	 * @return a new JobParameters object containing only a UUID parameter,
+	 * to ensure that the job instance will be unique.
 	 */
 	public JobParameters getUniqueJobParameters() {
 		Map<String, JobParameter> parameters = new HashMap<>();
-		parameters.put("random", new JobParameter((long) (Math.random() * JOB_PARAMETER_MAXIMUM)));
+		parameters.put("random", new JobParameter(UUID.randomUUID().toString()));
 		return new JobParameters(parameters);
 	}
 
--- a/spring-batch-test/src/test/java/org/springframework/batch/test/JobLauncherTestUtilsTests.java
+++ b/spring-batch-test/src/test/java/org/springframework/batch/test/JobLauncherTestUtilsTests.java
@@ -17,11 +17,7 @@ package org.springframework.batch.test;
 
 import org.junit.Test;
 
-import org.springframework.batch.core.ExitStatus;
-import org.springframework.batch.core.Job;
-import org.springframework.batch.core.JobExecution;
-import org.springframework.batch.core.Step;
-import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.*;
 import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.batch.core.configuration.annotation.JobBuilderFactory;
 import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
@@ -34,7 +30,11 @@ import org.springframework.context.annotation.AnnotationConfigApplicationContext
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 
 /**
  * @author mminella
@@ -50,6 +50,18 @@ public class JobLauncherTestUtilsTests {
 		JobExecution execution = testUtils.launchStep("step1");
 
 		assertEquals(ExitStatus.COMPLETED, execution.getExitStatus());
+	}
+
+	@Test
+	public void getUniqueJobParameters_doesNotRepeatJobParameters() {
+		ApplicationContext context = new AnnotationConfigApplicationContext(TestJobConfiguration.class);
+		JobLauncherTestUtils testUtils = context.getBean(JobLauncherTestUtils.class);
+		Set<JobParameters> jobParametersSeen = new HashSet<>();
+		for (int i = 0; i < 10_000; i++) {
+			JobParameters jobParameters = testUtils.getUniqueJobParameters();
+			assertFalse(jobParametersSeen.contains(jobParameters));
+			jobParametersSeen.add(jobParameters);
+		}
 	}
 
 	@Configuration


saikatbhadra:
Any update or feedback on my pull request?

benas:
Thank you for your PR. We will let you know when we start working on it.

saikatbhadra:
Any update on this fix to a testing bug?

saikatbhadra:
Is there any update here? This PR is over a year old.

benas:
This is breaking change. We don't include breaking changes in patch releases, but we can decide to include them in minor releases if the breaking change is minor and if it does not make sense to make it wait until the next major release.

Spring Batch depends on other projects (Spring Framework, Data, Integration) and other projects depend on Spring Batch as well (Spring Boot). The release plan requires a lot of effort by the entire Spring Engineering team to coordinate releases across the portfolio.

So even if one might think the change in a PR is minor and wondering why it is not merged yet, please note that reviewing/merging PRs is not just a matter of clicking a button but a very careful process to select in which release a PR should go depending on many constraints and priorities.

Practically speaking, we aim at releasing a minor release each year with patch releases in between, but that's a goal and not a commitment. That's why we try to be completely transparent about what's coming next each time we announce a release (here is [an example](https://spring.io/blog/2019/12/04/spring-batch-4-0-4-4-1-3-and-4-2-1-available-now)) in order to give our users a chance to plan their upgrade process upfront. 

So yes, based on the current yearly basis minor release plan, a PR can wait a year (or more) depending on our constraints and priorities. We appreciate your effort and we will let you know when this PR will be reviewed as mentioned [previously](https://github.com/spring-projects/spring-batch/pull/676#issuecomment-458445396). Please watch this issue for any updates to the assigned "Milestone" field. Thank you for your patience.

spring-batch-test/src/main/java/org/springframework/batch/test/JobLauncherTestUtils.java
```diff
@@ -155,12 +156,12 @@ public JobExecution launchJob(JobParameters jobParameters) throws Exception {
...
 	/**
-	 * @return a new JobParameters object containing only a parameter for the
-	 * current timestamp, to ensure that the job instance will be unique.
+	 * @return a new JobParameters object containing only a UUID parameter,
+	 * to ensure that the job instance will be unique.
 	 */
 	public JobParameters getUniqueJobParameters() {
 		Map<String, JobParameter> parameters = new HashMap<>();
-		parameters.put("random", new JobParameter((long) (Math.random() * JOB_PARAMETER_MAXIMUM)));
+		parameters.put("random", new JobParameter(UUID.randomUUID().toString()));
```
mminella, line 164: While I understand that iterating over 1000000 numbers randomly may not provide enough entropy to prevent duplicates, I feel like we should be able to figure out a way to address this in a non-breaking change.  Even something like taking the current time (long) and adding a random number to it should be good enough...  I'm thinking of 

`parameters.put("random", new JobParameter(new Date().getTime() + (long) (Math.random() * JOB_PARAMETER_MAXIMUM)));`

While not super elegant, it prevents the potential of breaking thousands of tests in the community, most of which have not been impacted by this issue.

saikatbhadra: I understand the need to minimize breaking changes.  I like your suggestion. In addition to the suggest change I would push for a new method (something like **getRealUniqueJobParameters** but with a better name) for users who are using this method in their test framework and want a stronger assurance that the method returns unique parameters.

So basically I'm proposing:
1. Let's add a non-breaking change using your suggestion to add more entropy in the existing method
2. Let's create a new method that uses UUID (or something else) to make a stronger guarantee of actual uniqueness.
3. There should be a decision on which of these methods JobLauncherTestUtils.launchJob should call. 

To give context, we were getting almost daily job failures on our integration tests because we were using `jobLauncherTestUtils.launchJob` which calls this. We ended up writing our own version of this test utils which uses UUID generation to generate the job parameters.

saikatbhadra: I understand the need to minimize breaking changes. I like your suggestion. In addition to the suggest change I would push for a new method (something like getRealUniqueJobParameters but with a better name) for users who are using this method in their test framework and want a stronger assurance that the method returns unique parameters.

So basically I'm proposing:

Let's add a non-breaking change using your suggestion to add more entropy in the existing method
Let's create a new method that uses UUID (or something else) to make a stronger guarantee of actual uniqueness.
There should be a decision on which of these methods JobLauncherTestUtils.launchJob should call.
To give context, we were getting almost daily job failures on our integration tests because we were using jobLauncherTestUtils.launchJob which calls this. We ended up writing our own version of this test utils which uses UUID generation to generate the job parameters.

mminella: I like the idea of a new method.  We could even depricate the current one in favor of it.  I also agree that we'd need a better name than `getRealUniqueJobParameters` ;)  I'd also vote to leave the `JobLauncherTestUtils.launchJob` method alone until the old method is actually removed.

benas: I see no need to add a new method, we can fix this in a non-breaking way. However, I would not roll my own random number generation logic but delegate this to a class provided by the JDK. `UUID.randomUUID().timestamp()` would have worked (from its Javadoc its results in a "60 bit timestamp measured in 100-nanosecond units since midnight, October 15, 1582 UTC"), but unfortunately Java does not support the generation of time based UUIDs (`UUID.randomUUID()` generates a UUID of type 4 which is not time based, so this fails with `java.lang.UnsupportedOperationException: Not a time-based UUID`).

The other option is to use a `java.security.SecureRandom`:

```diff
- parameters.put("random", new JobParameter((long) (Math.random() * JOB_PARAMETER_MAXIMUM)));
+ parameters.put("random", new JobParameter(new SecureRandom().nextLong()));
```

`SecureRandom` is a cryptographically strong random number generator which addresses most of the limitations of pseudo random generators like `java.util.Random` used behind the scene by `Match.random()` in the current code. `SecureRandom` is actually what's used by the UUID class to generate random UUIDs.

There are a couple of advantages of using this approach:

- It is not a breaking change (the type of the `random` job parameter is still `long`)
- It delegates random number generation to class provided by the JDK

BTW, the test added in this PR passes with this approach even if I increase the upper bound to 1 million elements (instead of 10k).

saikatbhadra: I'm hesitating a bit here because we are talking about a domain that I have no expertise in (how to generate UUID like numbers in a long). To take a step back let's clarify the requirements:

1.  **To maintain backwards compatability** - this means we need to have a random JobParameters with type long to fit past behavior.
2. **To have a low likelihood of a repeat** - A repeated number could result in a job failure of and breaks the method's contract of generating a unique"job parameter. I believe my original proposed solution used UUID's which have around 1 in 50 billion chance of collision, but unfortunately is not a practical solution due to it's data type. We probably don't need that strong of a collision guarantee but we definitely something much better than the current random number.

@benas - how does your proposed solution compare with just using something like UUID.randomUUID().getMostSignificantBits()

Again, I don't have any expertise in this area but the overall objective in my mind is to "reduce the probability of collision of a generated long value" so I just want to take a step back and see if there are any more appropriate JDK methods.

benas:
A UUID is a 128 bits value, and the methods `getMostSignificantBits()` and `getLeastSignificantBits()` return the 64 bit values corresponding to the most/least significant bits. Those work as well and are backward compatible (the test passes with both of them). My thinking is:

* Using a long value with `Math.random()` or `java.util.Random()` is the weakest solution in terms of uniqueness and causes the issue.
* Using a String value with `UUID.randomUUID()` is the strongest one but is a breaking change.

`SecureRandom` is the middle ground which satisfies both requirements 1) and 2) you mentioned and more (aka delegating entropy generation to a class from the JDK). The problem with `getMostSignificantBits()` and `getLeastSignificantBits()` is that although they work, if we use `getMostSignificantBits` for example, one would ask why not `getLeastSignificantBits` and vice versa. So I prefer the secure random option to avoid such open questions (and BTW that's why tried `UUID.randomUUID().timestamp()` but it does not work as mentioned previously). 

What I'm trying to do is to figure out how to fix the issue in a backward compatible way by reusing something from the JDK (using an external library would be a bad decision). What triggered my comment here is the discussion about adding new method and deprecating the current one, which is not needed IMO.

And since you are talking about stepping back, let's step back and be pragmatic here: we are in the context of a utility class that generates a random number used as a job parameter. I don't see why one would generate a million job instances in a for-loop in production (and even in tests, I don't see a valid use case for that, at least for such a big amount of job instances). So I would not make things more complex than they should be for a theoretical edge case in a utility class. And if what Spring Batch provides OOTB is not enough, one could always add a String parameter `uuid` with `UUID.randomUUID().toString()` in addition to the current long `random`.

saikatbhadra:
>  we are in the context of a utility class that generates a random number used as a job parameter.

The method is named `getUniqueJobParameters()` not `getRandomJobParameters()` so I believe the expectation is not a random parameter but a unique one. I think earlier this method actually used a timestamp based solution which reflects that the overall goal is for it be unique rather than random.

I'm just trying to come up with a solution where this method works in an integration tests since it is a test utility function. The problem is when you use this utility function in a test suite, you will get job failures since the job parameters will repeat over time. It's not the probability of having a collision across one run, but rather the probability of collisions across multiple runs of a test suite (@mminella solution might be better in this scenario)

I'm happy to go with whatever solution Spring would like to go with since any solution is better than the status quo, just trying to make sure you understand the use case where this function breaks.

benas:
> The method is named getUniqueJobParameters() not getRandomJobParameters() so I believe the expectation is not a random parameter but a unique one

The [Javadoc](https://docs.spring.io/spring-batch/docs/4.2.x/api/org/springframework/batch/test/JobLauncherTestUtils.html#getUniqueJobParameters--) of `getUniqueJobParameters()` says:

```
return a new JobParameters object containing only a parameter for the current
timestamp, to ensure that the job instance will be unique.
```

while the parameter `random` is not a timestamp (this inconsistency should be fixed in this PR btw). The method name and its Javadoc were consistent until this commit: https://github.com/spring-projects/spring-batch/commit/00aab3eafa47215f5ba524c8f74e42b79964910c#diff-c13b6ac20dce141719ab6f093d2774d5L160

So your comment applies to this commit and not to my statement "we are in the context of a utility class that generates a random number used as a job parameter". I just gave facts about the **current** status of the code regardless of the method name (I believe the aforementioned commit should have updated the Javadoc (and/or the method name?) according to the code change). The current behaviour is generating a random number and using it as a job parameter. The issue you reported is that this number happens to be non unique over time which is exactly what I'm trying fix in a backward compatible way without ending up with two methods `getUniqueJobParameters()` and `getRealUniqueJobParameters()` as you suggest.

> It's not the probability of having a collision across one run, but rather the probability of collisions across multiple runs of a test suite (@mminella solution might be better in this scenario)

If the range of random values is the same, the same problem would happen wether you generate those values in a for loop (like in your test) or over time across multiple runs of a test suite. Just to be clear, there is nothing wrong with Michael's solution, but it just pushes the problem a bit further without solving it completely. BTW, the test fails with this solution. That's why I said I would not try to implement my own entropy generation logic but delegate this to a battle-tested class provided by the JDK instead (either `UUID` or `SecureRandom`).

> just trying to make sure you understand the use case where this function breaks

I guess you correctly expressed your expectation with the failing test, so I think I understand since I was able to make it pass.

saikatbhadra:
Since the original repo has been removed from my machine, I had to create a new PR with the suggested changes:  https://github.com/spring-projects/spring-batch/pull/3701

PR #682, opened 2019-02-04
Using isQuoteCharacter in DelimitedLineTokenizer instead of direct co…

LukaszCaputa:
…mparison of chars. In case the function is overloaded in children class that fix can guarantee proper code execution.


RomanHiden:
This affects me as well.

benas:
LGTM. Rebased and merged as 7d45f2c5673bbfbb67c9713317d1d8a3d5e59e99. Thank you for your contribution!

PR #686, opened 2019-02-22
Constructors with var args/Lists

drumonii:
Adds constructors that take a `List` or var args to classes that set a `List` field. 

I did a rudimentary search through the code to find `List<`. So if I'm missing a class, I can easily add the appropriate constructors.

I can create an issue for this if needed. I found it to be fairly trivial since it's only constructor updates. Hence, I didn't update the `@author`. I did update the copyright year.


hbourada:
I think is a good enhancement mostly when creating jobs with java configuration it can reduce code lines

drumonii:
@hbourada any particular classes you have in mind? I was keeping this PR's scope low with targeting core and support classes especially for, like you said, usefulness with java config.

PR #687, opened 2019-02-23
ListItemWriter's iteration add can be replaced with addAll

drumonii:
`ListItemWriter` iterates through the items and adds for each. This can be replaced with an `addAll`.


benas:
Thanks for the PR. Rebased and merged as 15dd8e280c6ec230e539fd17e805dfe300520519.

PR #688, opened 2019-03-02
Simplify BatchTestContextCustomizerFactory

sbrannen:
Since `@SpringBatchTest` is effectively a marker annotation from the perspective of `BatchTestContextCustomizerFactory`, there is no need to retrieve the merged, synthesized annotation.

Checking for the presence of `@SpringBatchTest` is sufficient.

### DIFF
--- a/spring-batch-test/src/main/java/org/springframework/batch/test/context/BatchTestContextCustomizerFactory.java
+++ b/spring-batch-test/src/main/java/org/springframework/batch/test/context/BatchTestContextCustomizerFactory.java
@@ -32,8 +32,7 @@ public class BatchTestContextCustomizerFactory implements ContextCustomizerFacto
 
 	@Override
 	public ContextCustomizer createContextCustomizer(Class<?> testClass, List<ContextConfigurationAttributes> configAttributes) {
-		SpringBatchTest springBatchTest = AnnotatedElementUtils.findMergedAnnotation(testClass, SpringBatchTest.class);
-		if (springBatchTest != null) {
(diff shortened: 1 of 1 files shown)


benas:
Hi Sam,

Thank you for the PR! Rebased and merged as d9bbb0e.

Just for my understanding, would this be the same case for [Spring Integration](https://github.com/spring-projects/spring-integration/blob/master/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContextCustomizerFactory.java#L41-L42)?

BR
Mahmoud

sbrannen:
> Thank you for the PR! Rebased and merged as [d9bbb0e](https://github.com/spring-projects/spring-batch/commit/d9bbb0ecda8c918ec4071a45865f5c4daf842225).

You're welcome.

> Just for my understanding, would this be the same case for [Spring Integration](https://github.com/spring-projects/spring-integration/blob/master/spring-integration-test/src/main/java/org/springframework/integration/test/context/MockIntegrationContextCustomizerFactory.java#L41-L42)?

No, that would not apply to Spring Integration since `@SpringIntegrationTest` is not a marker annotation. On the contrary, it has a `noAutoStartup` attribute that is read in the following location:

https://github.com/spring-projects/spring-integration/blob/14c67421ef9ceaa0057845628a5bad41c4c61d7b/spring-integration-test/src/main/java/org/springframework/integration/test/context/IntegrationEndpointsInitializer.java#L57-L59

benas:
ok great! Thank you for the clarification.

PR #690, opened 2019-03-15
Initial support for batch metrics with Micrometer

benas:
Implements [BATCH-2774](https://jira.spring.io/browse/BATCH-2774)


spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
```diff
@@ -60,6 +63,8 @@
 public abstract class AbstractJob implements Job, StepLocator, BeanNameAware,
 InitializingBean {
 
+	private static final String METRICS_PREFIX = "spring.batch.";
```
benas: @mminella What is the best place to define this constant? It is currently duplicated in 4 places.

spring-batch-core/src/main/java/org/springframework/batch/core/step/item/SimpleChunkProcessor.java
```diff
@@ -292,16 +304,32 @@ protected void write(StepContribution contribution, Chunk<I> inputs, Chunk<O> ou
...
+				.register(Metrics.globalRegistry)
+		);
+	}
+
 	protected Chunk<O> transform(StepContribution contribution, Chunk<I> inputs) throws Exception {
 		Chunk<O> outputs = new Chunk<>();
 		for (Chunk<I>.ChunkIterator iterator = inputs.iterator(); iterator.hasNext();) {
 			final I item = iterator.next();
 			O output;
+			Timer.Sample sample = Timer.start(Metrics.globalRegistry);
```
benas: @jkschneider Should this be moved outside the loop? Do we need a sample for each processed item?

jkschneider: I suppose that's up to you and what you want to monitor. No right/wrong answer here.

build.gradle
```diff
@@ -98,6 +98,7 @@ allprojects {
 		beanshellVersion = '2.0b5'
 		jaxbApiVersion = '2.3.1'
 		jaxbImplVersion = '2.3.0.1'
+		micrometerVersion = '1.1.1'
```
jkschneider: 1.1.3 is out.

spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
```diff
@@ -353,6 +362,13 @@ public final void execute(JobExecution execution) {
 					execution.setExitStatus(exitStatus.and(newExitStatus));
 				}
 
+				timerSample.stop(Timer.builder(METRICS_PREFIX + "job")
+						.description("Job duration in seconds")
```
jkschneider: Don't refer to "seconds" here. The actual base unit of time will depend on the monitoring system, e.g. seconds for prometheus, nanos for Atlas, millis for Datadog. Just call it "Job duration".

spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java
```diff
@@ -256,6 +261,13 @@ public final void execute(StepExecution stepExecution) throws JobInterruptedExce
 						+ "This job is now in an unknown state and should not be restarted.", name, stepExecution.getJobExecution().getJobInstance().getJobName()), e);
 			}
 
+			sample.stop(Timer.builder(METRICS_PREFIX + "step")
+							.description("Step duration in seconds")
```
jkschneider: See above comment about "seconds".

spring-batch-samples/src/test/java/org/springframework/batch/sample/metrics/BatchMetricsTests.java
```diff
@@ -0,0 +1,168 @@
...
+		
+		// when
+		JobExecution jobExecution = jobLauncher.run(job, new JobParameters());
+		
+		// then
+		assertEquals(ExitStatus.COMPLETED, jobExecution.getExitStatus());
+		List<Meter> meters = Metrics.globalRegistry.getMeters();
+		assertThat(meters, Matchers.hasSize(7));
+
+		Meter.Id itemReadMeter = meters.get(0).getId();
```
jkschneider: Micrometer has a built-in testing facility for this. Use:
```
Metrics.globalRegistry.get("spring.batch.item.read")
   .tag("job.name", "job") // single tag at a time
   .tag("step.name", "step2", "status", "SUCCESS") // or you can do them all at once
   .timer();
```

If you use `get(..)` and there isn't a meter with this type/name/tags, Micrometer will throw a runtime exception and generate a detailed description of why it couldn't find it and which meters got closest to matching your requirements.

This has the added benefit of not requiring the meters to be in any certain order.

spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
mminella: I'm struggling with an answer here.  The reason is that I can't think of anywhere else in the framework (short of logging) that we have a library like this that we have sprinkled through out the framework in this decentralized way.  Because of that, there isn't an obvious place to hang that constant (or the `SUCCESS`/`FAILURE` ones either for that matter).  I considered suggesting putting the logic we need into a `MetricsHandler` or something of the like that would encapsulate just our usage of micrometer (static util methods).  That would be a logical place to drop these constants as well.  However, you then get closer to the facade of a facade that we were trying to avoid with that approach (it's not the same thing, but it is similar)...

build.gradle
benas: Updated to 1.1.3. I see 1.1.4 is planned for today. I will update to 1.1.4 before our release planned for next Friday.

spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
benas: Make sense. Removed the unit from all descriptions.

spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java
benas: Make sense. Removed the unit from all descriptions.

spring-batch-samples/src/test/java/org/springframework/batch/sample/metrics/BatchMetricsTests.java
benas: Got it. I updated the test to get each meter with its name/tags. That's indeed a better approach to avoid getting metrics in any certain order.

spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
```diff
@@ -297,7 +302,11 @@ public final void execute(JobExecution execution) {
 		}
 
 		JobSynchronizationManager.register(execution);
-
+		LongTaskTimer longTaskTimer = LongTaskTimer.builder(METRICS_PREFIX + "jobs.active")
+				.description("Active jobs")
```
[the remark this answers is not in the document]
benas: I agree. I changed `jobs.active` to `job.active` for consistency. However, the description should remain in plural because this is what it really is: Currently "in flight" jobs to use micrometer terminology.

spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java
benas: Good idea. I moved all interactions with micrometer to a `BatchMetrics` class. This makes the code cleaner and is a good place to define metrics constants.

spring-batch-core/src/main/java/org/springframework/batch/core/StepContribution.java
```diff
@@ -191,6 +195,14 @@ public int getProcessSkipCount() {
 		return processSkipCount;
 	}
 
+	/**
+	 * Public getter for the parent step execution of this contribution.
+	 * @return parent step execution of this contribution
+	 */
+	public StepExecution getStepExecution() {
+		return stepExecution;
+	}
```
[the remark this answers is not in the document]
benas: we might need other things in addition to job/step name so I would stick to `StepExecution`.

build.gradle
benas: I saw micrometer v1.1.4 has been released today. I updated the PR to use 1.1.4.

benas:
@Buzzardo Could you please review the new docs related to this PR? See here: https://github.com/spring-projects/spring-batch/pull/690/commits/5eb2d6f8eb880b2d854e9092eff10e28e0ab0f1a

spring-batch-docs/asciidoc/monitoring-and-metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+|spring.batch.item.read|TIMER|Duration of item reading
+|spring.batch.item.process|TIMER|Duration of item processing
+|spring.batch.chunk.write|TIMER|Duration of chunk writing
+|===============
+
+[[custom-metrics]]
+
+=== Custom metrics
+
+If you want to use your own metrics in your custom components, we recommend to use
```
Buzzardo: Should be "we recommend using" rather than "we recommend to use".

spring-batch-docs/asciidoc/monitoring-and-metrics.adoc
```diff
@@ -0,0 +1,69 @@
...
+=== Built-in metrics
+
+Metrics collection does not require any specific configuration. All metrics provided
+by the framework are registered in
+link:$$https://micrometer.io/docs/concepts#_global_registry$$[Micrometer's global registry]
+under the `spring.batch.` prefix. The following table explains all metrics in details:
+
+|===============
+|__Metric Name__|__Type__|__Description__
+|spring.batch.job|TIMER|Duration of job execution
```
Buzzardo: I would use the code font for the Metric Name and Type values (not the headers but the contents of each row.

spring-batch-core/src/main/java/org/springframework/batch/core/StepContribution.java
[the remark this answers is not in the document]
mminella: The more I think about this, the less I like it.  The `StepContribution` is basically supposed to represent the `StepExecution` so wrapping the `StepExecution` in the `StepContribution` just doesn't feel right to me.  I'd rather just copy the fields over into this class.  I get that it's more code, but I believe it's more clear on the intent.  If you agree with this, I can make the update on merge.

spring-batch-core/src/main/java/org/springframework/batch/core/metrics/BatchMetrics.java
```diff
@@ -0,0 +1,116 @@
...
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * Main entry point to interact with Micrometer's {@link Metrics#globalRegistry}.
+ * Provides common metrics such as {@link Timer}, {@link Counter} and {@link Gauge}.
+ *
+ * Only intended for internal use.
+ *
+ * @author Mahmoud Ben Hassine
+ */
```
mminella: `@since` tag.

spring-batch-core/src/main/java/org/springframework/batch/core/metrics/BatchMetrics.java
```diff
@@ -0,0 +1,116 @@
...
+
+	private BatchMetrics() {}
+
+	/**
+	 * Create a {@link Timer}.
+	 *
+	 * @param name        of the timer. Will be prefixed with {@link BatchMetrics#METRICS_PREFIX}.
+	 * @param description of the timer
+	 * @param tags        of the timer
+	 * @return a new timer instance
```
mminella: I'm not saying we need to make a big deal out of it right now, but your column based layout for the `@param` javadoc is different than the rest of the codebase.  We should come to an agreement on one way or the other.

spring-batch-core/src/main/java/org/springframework/batch/core/metrics/BatchMetrics.java
benas: Added `@since` tag.

spring-batch-core/src/main/java/org/springframework/batch/core/metrics/BatchMetrics.java
benas: Fixed to be consistent with the rest of the code base.

spring-batch-core/src/main/java/org/springframework/batch/core/StepContribution.java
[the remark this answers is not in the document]
benas: > so wrapping the `StepExecution` in the `StepContribution` just doesn't feel right to me

We are not wrapping the `StepExecution` in the `StepContribution`, that would be wrong! We are just keeping a reference to the enclosing `StepExecution`. FTR, when we create a step execution from a job execution using [`JobExecution#createStepExecution`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/JobExecution.java#L228), the returned `StepExecution` has a [reference to the enclosing job execution](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/StepExecution.java#L102) and a [getter for it](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/StepExecution.java#L331). Same thing  for the step contribution here, when we create a step contribution from a step executionusing [StepExecution#createStepContribution](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/StepExecution.java#L340-L342), we keep a reference to the enclosing step execution. This is probably better explained with a diagram:

step-contribution 001

The dotted links were missing, I added them here. What I like about this approach is:

* Design consistency with Job Execution -> Step Execution
* Simplicity: I don't like passing job name, step name, etc everywhere in method signatures
* Proactive: we might need other infos in addition to job name and step name

If your concern is getting access to the step execution from the step contribution in `Tasklet#execute(StepContribution contribution, ChunkContext chunkContext)` and messing with it, then this would not be an issue since the step execution is accessible from the chunk context through `chunkContext.getStepContext().getStepExecution()`.

benas:
Rebased and merged as 57c8d70ff3f400b54389337809332eebb14d87a2.

benas:
Rebased and merged as 57c8d70ff3f400b54389337809332eebb14d87a2.

hbourada:
Hello
I think creating a direct micro meter dependency directly in spring batch core is a bad design for jobs that do not want use micro meter or are monitoring directly their job by other means (listeners by example)
Also I think that metrics should be moved in spring boot part in actuator which includes such functionality 
Regards
Hichem B

benas:
Hi @hbourada 

Thank you for your feedback!

> I think creating a direct micro meter dependency in spring batch core is a bad design for jobs that do not want use micro meter or are monitoring directly their job by other means (listeners by example)

It would have been a bad design if we added a dependency to a specific monitoring library or a specific monitoring system client, but the idea here is to use a metrics facade (which is micrometer) and let users choose the monitoring backend of their choice. In the logging world, this would be the same as adding a dependency to SLF4J (rather than a specific logging framework), and this is exactly what we did but for metrics.

That said, nothing prevents you form using another metrics library to monitor your job with listeners.

> Also I think that metrics should be moved in spring boot part in actuator which includes such functionality

What about non boot users? As Spring Batch framework developers, we need to think about Spring Batch users that are **not** Spring Boot users. Non boot users will have to configure their metrics client manually, while boot users will get this for free, but **both of them** should be able to consume batch metrics from Micrometer's registry.

Regards,
Mahmoud

PascalSchumacher:
Some methods of `BatchMetrics` like `createCounter`, `createGauge` are currently unused.

I'm wondering if these should be deleted, or is it planned to add other metrics like counters in the future?

benas:
@PascalSchumacher We are gathering feedback on this initial micrometer support until the next GA (planned for August) and we might introduce new metrics of type counter and gauge in the upcoming milestones. If it's not the case, those methods will be removed.

PascalSchumacher:
Thank for the information. Imho more metrics would be great to have.

benas:
> If it's not the case, those methods will be removed.

We are going to release 4.2.0.RC1 soon. We did not receive feedback to add new metrics, so I [removed](63d3e4e043bd5ae8c2d7ab71bac6c89dcff1fd8f) these unused methods. We can always re-add them if needed in the future.

PR #713, opened 2019-04-10
BATCH-2805 Fix . hibernate.default_batch_fetch_size option does not work in JpaPagingItemReader. 

jojoldu:
Unlike JpaRepository, the ```hibernate.default_batch_fetch_size``` options does not work with JpaPagingItemReader.
(Below is the result of my test code)

image

In other words, using JpaPagingItemReader causes an N + 1 problem.

As you can see, the JpaPagingItemReader has a transaction scope inside doReadPage().

So I deleted the contents of the transaction to raise this range in chunk units.

image

I do not see any more problems.

Send PR

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaPagingItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaPagingItemReader.java
@@ -81,6 +81,7 @@ import org.springframework.util.ClassUtils;
  * @author Thomas Risberg
  * @author Dave Syer
  * @author Will Schipp
+ * @author jojoldu
  * @since 2.0
  */
 public class JpaPagingItemReader<T> extends AbstractPagingItemReader<T> {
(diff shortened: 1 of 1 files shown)


benas:
Thank you for opening this PR. If you want to disable the transaction around each page read, you can set `transacted` to false (instead of removing the transaction handling code like in this PR).

From the Javadoc of [JpaPagingItemReader](https://docs.spring.io/spring-batch/docs/4.2.x/api/org/springframework/batch/item/database/JpaPagingItemReader.html) :

```
All entity access is performed within a new transaction, independent of any existing Spring managed transactions.
```

There is a good reason for that which is to make the JPA reader consistent with other readers (Like the [JdbcCursorItemReader](https://github.com/spring-projects/spring-batch/blob/a7092a21e428cf904f210aff2682b518dc3649c5/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JdbcCursorItemReader.java#L40-L41) for example).

This PR basically removes that, which we cannot accept as it is now. There is no failing test in this PR that passes with the suggested changes. Without such failing test, I cannot confirm that there is an issue and that this PR fixes it.

Thank you anyway!

jojoldu:
Thanks for the answers @benas !

However, I think there is a problem with making transacted false as you said.
If transacted is set to false, entity read from Reader due to entityManager.detach causes Lazy initialize a collection in Processor.

image

It would be nice if entityManager.detach could be selected as a separate option, but I think this will be a little different from the intention of JpaPagingItemReader.

So, we decided to solve this problem by creating a Custom ItemReader.

Thanks for the kind answer.

PR #717, opened 2019-05-21
Master slave replacement

Buzzardo:
Per a request from Michael Minella, I replaced "master" with "manager" and "slave" with "worker" in the Spring Batch Reference Guide. I heartily concur with replacing terms that have such dark connotations.


mminella:
Added a polish commit, squashed, rebased, and merged onto master as 4b9554fcce80e38f7adeffc82a26436199c28191

PR #718, opened 2019-06-25
add type safe getters to execution context

IlyaNerd:


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
```diff
@@ -273,6 +273,43 @@ public Object get(String key) {
...
+	}
+
+	/**
+	 * Typesafe getter for the value represented by the provided key, with cast to given class.
+	 *
+	 * @param key The key to get a value for
+	 * @param clazz The class of return type
+	 * @param defaultValue Default value in case element is not present
+	 * @param <V> Type of returned value
+	 * @return The value of given type represented by the given key or {@code null} if the key
```
fmbenhassine, line 301: I believe this should be "or default value" instead of "or {@code null}". I will fix it on merge.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
```diff
@@ -273,6 +273,43 @@ public Object get(String key) {
...
+	 * is not present
+	 */
+	@Nullable
+	public <V> V get(String key, Class<V> clazz, @Nullable V defaultValue) {
+		Object value = this.map.get(key);
+		if (value == null) {
+			return defaultValue;
+		}
+		return clazz.cast(value);
```
fmbenhassine, line 310: We need to check if the value if of the given type before the case here, similar to what is done in `readAndValidate`. I will add this check.

fmbenhassine:
Rebased and merged as 306d7379bb87236a903a6b2bb68ec53eeb2487ea. Refined as explained in 54239fe9e8b9977b40ae665e261d34b8409a8d8f.

PR #719, opened 2019-06-26
Add new methods to get the last job instance/execution in JobInstance/JobExecution DAOs

benas:
Before this PR, getting the last job instance/execution required to load all job instances/executions from the database and filter them on the client side in memory.

This PR introduces new methods that use database queries to get the last job instance/execution without the need to load all job instances/executions (filtering is done on the database side). This change improves memory consumption as well as the performance of starting the next instance of a job.

Resolves [BATCH-1784](https://jira.spring.io/browse/BATCH-1784)

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,6 +50,16 @@ public interface JobExplorer {
 	List<JobInstance> getJobInstances(String jobName, int start, int count);
 
 	/**
+	 * Find the last job instance for the given job.
+	 * @param jobName name of the job
+	 * @return the last job instance if any or null otherwise
+	 */
+	@Nullable
+	default JobInstance getLastJobInstance(String jobName) {
+		return null;
+	}
+
+	/**
 	 * Retrieve a {@link JobExecution} by its id. The complete object graph for
 	 * this execution should be returned (unless otherwise indicated) including
 	 * the parent {@link JobInstance} and associated {@link ExecutionContext}
@@ -94,6 +104,18 @@ public interface JobExplorer {
 	 * @return the set of all executions for the specified {@link JobInstance}
 	 */
 	List<JobExecution> getJobExecutions(JobInstance jobInstance);
+
+	/**
+	 * Find the last {@link JobExecution} to have been created for a given
+	 * {@link JobInstance}.
+	 * @param jobInstance the {@link JobInstance}
+	 * @return the last {@link JobExecution} to execute for this instance or
+	 * {@code null} if no job execution is found for the given job instance.
+	 */
+	@Nullable
+	default JobExecution getLastJobExecution(JobInstance jobInstance) {
+		return null;
+	}
 
 	/**
 	 * Retrieve running job executions. The corresponding step executions may
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
@@ -50,13 +50,15 @@ public interface JobExplorer {
 	List<JobInstance> getJobInstances(String jobName, int start, int count);
 
 	/**
-	 * Find the last job instance for the given job.
+	 * Find the last job instance by Id for the given job.
 	 * @param jobName name of the job
-	 * @return the last job instance if any or null otherwise
+	 * @return the last job instance by Id if any or null otherwise
+	 *
+	 * @since 4.2
 	 */
 	@Nullable
 	default JobInstance getLastJobInstance(String jobName) {
-		return null;
+		throw new UnsupportedOperationException();
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
```diff
@@ -49,6 +49,16 @@
 	 */
 	List<JobInstance> getJobInstances(String jobName, int start, int count);
 
+	/**
+	 * Find the last job instance for the given job.
+	 * @param jobName name of the job
+	 * @return the last job instance if any or null otherwise
+	 */
```
mminella: `@since`
Also, what criteria do we use for "last"?  Last started?  Last completed?  Last by id?

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
```diff
@@ -95,6 +105,18 @@
 	 */
 	List<JobExecution> getJobExecutions(JobInstance jobInstance);
 
+	/**
+	 * Find the last {@link JobExecution} to have been created for a given
+	 * {@link JobInstance}.
+	 * @param jobInstance the {@link JobInstance}
+	 * @return the last {@link JobExecution} to execute for this instance or
+	 * {@code null} if no job execution is found for the given job instance.
+	 */
```
mminella: `@since`.  Also what does "last" mean (see previous comments)?

spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JobInstanceDao.java
```diff
@@ -96,6 +96,16 @@
 	 */
 	List<JobInstance> getJobInstances(String jobName, int start, int count);
 
+	/**
+	 * Fetch the last job instance for the given job.
+	 * @param jobName name of the job
+	 * @return the last job instance if any or null otherwise
+	 */
```
mminella: `@since` and define what "last" means

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
```diff
@@ -49,6 +49,16 @@
...
 	List<JobInstance> getJobInstances(String jobName, int start, int count);
 
+	/**
+	 * Find the last job instance for the given job.
+	 * @param jobName name of the job
+	 * @return the last job instance if any or null otherwise
+	 */
+	@Nullable
+	default JobInstance getLastJobInstance(String jobName) {
+		return null;
```
mminella: Is `null` the right thing to do here as the default or should we do something like `throw new UnsuportedOperationException()`?

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
```diff
@@ -95,6 +105,18 @@
...
+	/**
+	 * Find the last {@link JobExecution} to have been created for a given
+	 * {@link JobInstance}.
+	 * @param jobInstance the {@link JobInstance}
+	 * @return the last {@link JobExecution} to execute for this instance or
+	 * {@code null} if no job execution is found for the given job instance.
+	 */
+	@Nullable
+	default JobExecution getLastJobExecution(JobInstance jobInstance) {
+		return null;
```
mminella: Is `null` the right thing to do here as the default or should we do something like `throw new UnsuportedOperationException()`?

spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JobInstanceDao.java
```diff
@@ -96,6 +96,16 @@
...
+	@Nullable
+	default JobInstance getLastJobInstance(String jobName) {
+		return null;
```
mminella: Is `null` the right thing to do here as the default or should we do something like `throw new UnsuportedOperationException()`?

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
benas: It is the last by Id. I updated the Javadoc accordingly.

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
benas: Added the `since` tag. The javadoc mentions that it's the last job execution that has been created.

spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JobInstanceDao.java
benas: It is the last job instance by id. I updated the Javadoc.

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
benas: Fixed

spring-batch-core/src/main/java/org/springframework/batch/core/explore/JobExplorer.java
benas: Fixed

spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JobInstanceDao.java
benas: Fixed

mminella:
LGTM.  Squashed and merged into master as 7c1fbede1b53f1f6e00f6dd1dc614d1a99faf09c

PR #720, opened 2019-07-01
Improve the lookup for running job executions in CommandLineJobRunner

benas:
This PR resolves [BATCH-2422](https://jira.spring.io/browse/BATCH-2422).

Before this PR, getting the running job executions loaded all job executions from the database and filtered them in-memory.

This PR uses `JobExplorer#findRunningJobExecutions` which issues a query that does the filtering on the database side. This change considerably improves the performance of stopping a job.


mminella:
LGTM.  Rebased and merged as 872845cae8f5922c1b8747c13c3a3b9d4956e8bc

PR #721, opened 2019-07-17
Implement AvroItemReader and AvroItemWriter

dturanski:
Resolves https://jira.spring.io/browse/BATCH-2833


benas:
Hi @dturanski ,

Thank you for this PR! I have a couple of comments:

* We try to not leak third party APIs in our batch APIs. In this case, we have public constructors with Avro APIs  in them (`DatumReader`, `DataFileWriter`, `Schema`, etc)
* All item readers and writers work with `org.springframework.core.io.Resource` objects. For consistency, I would remove constructors accepting `File`, `InputStream`, etc
* According to [Avro's docs](https://avro.apache.org/docs/current/index.html#schemas), when Avro data is stored in a file, its schema is stored with it. It is also mentioned that when Avro data is read, the schema used when writing it is always present.  So I'm not sure we need the `embeddedHeader` flag in the `AvroItemReader` or support files without schema in them (like `user-data-no-schema.avro`)

I tried to illustrate these points in the [PoC here](https://github.com/benas/spring-batch/commit/a03a0ddc3983673450a135f3a2eec76c7b3449f3).

Let me know if you have any question.

Kr,
Mahmoud

dturanski:
The schema is typically stored in the file, but not always.  AFAICT excluding the schema header is more common in streaming scenarios to avoid overhead. I'm not sure how common this is in file processing scenarios. In fact, the use case that motivated this did not include the schema so I want to support either way.  Additionally,  we  were ingesting an s3 input stream directly. I can look into using an S3 Resource.  I'm not sure I agree with the constraints to not expose 3rd party APIs here:  1) Anyone using Avro will be familiar with them  2) They are complex and  may be configured in various ways, so abstracting them to allow for every possibility is not practical.  Similar batch components, e.g. GemFire, Kafka delegate to  GemfireTemplate and KafkaTemplate respectively.  These things necessarily do expose native APIs. We don't currently have anything like an AvroTemplate AFAIK.

benas:
Hi David ,

Thank you for updating the PR!

For the item reader, it looks like it is not possible to read user defined types (like `PlainOldUser`) unless I'm missing something. The following tests are missing in `AvroItemReaderTests` and are failing:

```java
@Test
public void readPlainOldUserUsingResources() throws Exception {

	AvroItemReader<PlainOldUser> itemReader = new AvroItemReader<>(dataResource, PlainOldUser.class); // should it also work with `dataResourceWithSchema`?
	itemReader.setName(itemReader.getClass().getSimpleName());
	itemReader.setEmbeddedHeader(false);

	verify(itemReader, plainOldUsers());

        // fails with `org.apache.avro.AvroRuntimeException: Malformed data. Length is negative: -49`
}

@Test
public void readPlainOldUsersWithEmbeddedSchema() throws Exception {

	AvroItemReader<PlainOldUser> itemReader = new AvroItemReader<>(dataResourceWithSchema, PlainOldUser.class);
	itemReader.setEmbeddedHeader(true);
	itemReader.setName(itemReader.getClass().getSimpleName());

	verify(itemReader, plainOldUsers());

        // fails with org.apache.avro.AvroTypeException: Found org.springframework.batch.item.avro.example.User, expecting org.springframework.batch.item.avro.support.AvroTestFixtures.PlainOldUser, missing required field favoriteColor
}
```

Can you please take a look?

Kr,
Mahmoud

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/AvroItemReader.java
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param data the {@link Resource} containing the data to be read.
+	 * @param schema the {@link Resource} containing the
+	 */
+	public AvroItemReader(Resource data, Resource schema) {
+		Assert.notNull(data, "'data' is required.");
+		Assert.state(data.exists(), "'data' " + data.getFilename() +" does not exist.");
+		Assert.notNull(schema, "'schema' is required");
+		Assert.state(schema.exists(), "'schema' " + schema.getFilename() +" does not exist.");
+		try {
```
benas: Same as previously mentioned. This code can be moved to the `doOpen` method.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/AvroItemWriter.java
```diff
@@ -0,0 +1,204 @@
...
+
+	private OutputStreamWriter<T> outputStreamWriter;
+
+	private WritableResource resource;
+
+	private Schema schema;
+
+	private Class<T> clazz;
+
+	private boolean embedHeader = true;
```
benas: `embedSchema` is more explicit IMO.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/avro/AvroItemWriterTests.java
```diff
@@ -0,0 +1,113 @@
...
+		avroItemWriter.close();
+
+		verifyRecordsWithEmbeddedHeader(outputStream.toByteArray(), this.genericPlainOldUsers(), GenericRecord.class);
+
+	}
+
+	@Test
+	public void itemWriterForPojos() throws Exception {
+
+		AvroItemWriter avroItemWriter = new AvroItemWriter(output,plainOldUserSchemaResource, PlainOldUser.class);
```
benas: Should specify the generic type to avoid warnings: `AvroItemWriter< PlainOldUser >`. We have a couple of them in this test class.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/avro/AvroItemWriterTests.java
```diff
@@ -0,0 +1,113 @@
...
+		avroItemWriter.open(new ExecutionContext());
+		avroItemWriter.write(this.plainOldUsers());
+		avroItemWriter.close();
+
+		verifyRecordsWithEmbeddedHeader(outputStream.toByteArray(), this.plainOldUsers(), PlainOldUser.class);
+
+	}
+
+	@Test
+	public void itemWriterWithNoEmbeddedHeaders() throws Exception {
```
benas: I'm a bit confused by this test and `shouldFailWitNoSchema` test, especially the following part:

```java
AvroItemWriter<PlainOldUser> avroItemWriter = new AvroItemWriter<>(output, plainOldUserSchemaResource,PlainOldUser.class);
avroItemWriter.setEmbedHeader(false);
```

 Why should I provide a Schema if I can set `embedHeader` to false?

For the reader, we can read `PlainOldUser`s from an avro file not containing a schema without providing a schema. In the same way, we should be able to write `PlainOldUser`s to an avro file without embedding any schema. In that case, why is the schema mandatory? To be specific, having a mandatory schema and an option to not embed it is confusing to me.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/AvroItemReader.java
```diff
@@ -0,0 +1,185 @@
...
+
+import org.springframework.batch.item.ItemReader;
+import org.springframework.batch.item.ItemStreamException;
+import org.springframework.batch.item.support.AbstractItemCountingItemStreamItemReader;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.core.io.Resource;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link ItemReader} that deserializes data from a File or Input Stream containing serialized Avro objects.
```
benas: `File or Input Stream` should be updated to `Resource`.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/AvroItemWriter.java
```diff
@@ -0,0 +1,204 @@
...
+import org.springframework.batch.item.ItemStreamException;
+import org.springframework.batch.item.ItemWriter;
+import org.springframework.batch.item.support.AbstractItemStreamItemWriter;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.WritableResource;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link ItemWriter} that serializes data to an OutputStream or File using Avro.
```
benas: `OutputStream or File` should be replaced with `WritableResource`

dturanski:
Re, the additional tests. They won't work since PlainOldUser and (Avro generated) User are different types. POU has its own schema which is `ReflectData.get().getSchema(PlainOldUser.class);` So it has camel case field names, etc. It is not intended to be related to User, I made it "the same" out of laziness.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/avro/AvroItemWriterTests.java
dturanski: Good catch. The schema is not required by the writer if embedSchema is false.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/AvroItemReader.java
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param resource the {@link Resource} containing objects serialized with Avro.
+	 * @param clazz the class of the objects to be deserialized. This should be an Avro generated class.
+	 */
+	public AvroItemReader(Resource resource, Class<T> clazz) {
+		Assert.notNull(resource, "'resource' is required.");
+		Assert.notNull(clazz, "'class' is required.");
+
+		try {
+			this.inputStream = resource.getInputStream();
```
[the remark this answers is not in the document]
dturanski: This would require keeping the resource and schema as member variables just to have them in doOpen.  I don't see how that's cleaner.

benas:
I know that `User` and `PlainOldUser` are different types. My point is that since we provide a constructor accepting a data resource and a user defined type (no schema), the reader should be able to read avro data and map it to my domain object using reflection, without having to generate or provide a schema with  `ReflectData.get().getSchema(PlainOldUser.class)`. Does this make sense? I'm probably misunderstanding something but to me the following setup should be enough to be able to read my domain object `PlainOldUser `:

```
AvroItemReader<PlainOldUser> itemReader = new AvroItemReader<>(dataResource, PlainOldUser.class);
itemReader.setEmbeddedSchema(false); // given that `dataResource` does not embed any schema
```

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/avro/AvroItemWriterTests.java
dturanski: The latest commit provides a "schemaless" constructor and sets `embedHeader=false` if that is used.  `setEmbedHeader()` and corresponding build method have been removed.

dturanski:
For some reason the serialized data for Avro generated User without embedded schema,  and PlainOldUser are not compatible. I created a new test resource by serializing PlainOldUsers without embedded schema. You can open both files with a text editor and see that they are different, so I assume this is an Avro limitation.  I  added an  ItemReaderTest that uses the new resource and that works fine.

benas:
LGTM. Rebased and merged as bd53168ddb613aa5a5f8bc550f20a828ce3cf17d.

Thank you very much @dturanski !

PR #722, opened 2019-07-18
Fixed a refresh bug in the document toggle

Buzzardo:
While working on another feature, I noticed that the document toggle wasn't updating the ToC. I fixed that. (It matters only in the case where entire headings are toggled, but we do have one such case.)


mminella:
LGTM.  Rebased, merged, and back ported to all 4.x lines.

PR #724, opened 2019-07-26
Allow access from subclasses to JdbcBatchItemWriter instance variables.

garydgregory:
Allow access from subclasses to JdbcBatchItemWriter instance variables.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JdbcBatchItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JdbcBatchItemWriter.java
@@ -63,19 +63,19 @@ public class JdbcBatchItemWriter<T> implements ItemWriter<T>, InitializingBean {
 
 	protected static final Log logger = LogFactory.getLog(JdbcBatchItemWriter.class);
 
-	private NamedParameterJdbcOperations namedParameterJdbcTemplate;
+	protected NamedParameterJdbcOperations namedParameterJdbcTemplate;
 
-	private ItemPreparedStatementSetter<T> itemPreparedStatementSetter;
+	protected ItemPreparedStatementSetter<T> itemPreparedStatementSetter;
 
-	private ItemSqlParameterSourceProvider<T> itemSqlParameterSourceProvider;
+	protected ItemSqlParameterSourceProvider<T> itemSqlParameterSourceProvider;
 
-	private String sql;
+	protected String sql;
 
-	private boolean assertUpdates = true;
+	protected boolean assertUpdates = true;
 
-	private int parameterCount;
+	protected int parameterCount;
 
-	private boolean usingNamedParameters;
+	protected boolean usingNamedParameters;
 
 	/**
 	 * Public setter for the flag that determines whether an assertion is made


pivotal-issuemaster:
@garydgregory Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=724)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=724) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@garydgregory Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=724)!

benas:
Hi @garydgregory ,

Thank you for the PR. These changes make sense to me. However, I'm curious to know more about the use case that requires extending `JdbcBatchItemWriter`. What was the motivation behind that?

KR,
Mahmoud

garydgregory:
Hi @benas ,
Here is the hacking going on, it all makes sense in the end:
- I copied SpingBatch's `JdbcBatchItemWriter` to our code base as `SpingBatchJdbcBatchItemWriter` and made the changes listed in this PR and I also changed:
```
            if (usingNamedParameters) {
                if (items.get(0) instanceof Map) {
                    updateCounts = namedParameterJdbcTemplate.batchUpdate(sql, items.toArray(new Map[0]));
```
to:
```
			if (usingNamedParameters) {
				if(items.get(0) instanceof Map && this.itemSqlParameterSourceProvider == null) {
					updateCounts = namedParameterJdbcTemplate.batchUpdate(sql, items.toArray(new Map[items.size()]));
```
I did not include the above in this PR because at the time I was not sure if it made sense for Spring Batch, but we need it.
- Then, I have a subclass OurProductJdbcBatchItemWriter<T> extends SpingBatchJdbcBatchItemWriter<T> which tracks the following in instance variables:
```
    private ParsedSql parsedSql;
    private List<String> parameterNames;
```
Which are initialized through an ugly hack:
```
    @Override
    public void afterPropertiesSet() throws Exception {
        parsedSql = invokeMethod(namedParameterJdbcTemplate, "getParsedSql", sql);
        parameterNames = invokeMethod(parsedSql, "getParameterNames");
        super.afterPropertiesSet();
    }

    private <R> R invokeMethod(final Object object, final String methodName, final Object... args)
            throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
        return (R) MethodUtils.invokeMethod(object, true, methodName, args);
    }
```
Where `MethodUtils` is from Apache Commons Lang 3.9.

Then we actually implement a product feature using Apache Commons Text's string substitution which I do not expect SB to support so it still belongs in our subclass no matter what:
```
    @Override
    public void setSql(String sql) {
        super.setSql(StringSubstitutor.createInterpolator().replace(sql));
    }
```

But the meat of the class is adding nulls for missing entries for expected parameters:
```
    @Override
    public void write(final List<? extends T> items) throws Exception {
        addMissingEntries(items);
        super.write(items);
    }

    /**
     * Adds missing entries for expected parameters.
     *
     * @param items
     */
    private void addMissingEntries(final List<? extends T> items) {
        if (items != null) {
            for (final T item : items) {
                if (item instanceof Map) {
                    // TODO Handle 0 vs. empty string
                    parameterNames.forEach(name -> ((Map<Object, Object>) item).putIfAbsent(name, null));
                }
            }
        }
    }
```

So the patch could be more extensive but I wanted to keep it to a minimum. 

We would not need to clone the SB class `JdbcBatchItemWriter` in order to subclass it if it included the test `items.get(0) instanceof Map && this.itemSqlParameterSourceProvider == null` instead of just `items.get(0) instanceof Map`

I can update the PR with this test if you think it is OK.

benas:
Thank you for your feedback! LGTM. rebased and merged as c4010fbffa6b71cbcfe79d523023251ce73666a4.

PR #716, opened 2019-05-15
Performance improvement for partitions with a large number of steps

NaanProphet:
New method on JobRepository that enables better performance of SimpleStepExecutionSplitter. By retrieving all step executions at the time of partitioning outside the for loop, the number of calls to the database is reduced.

Also updated JSR version to be performant. Deprecated the older step execution methods and cleaned up the javadocs. Stub implementations provided for common support classes

https://jira.spring.io/browse/BATCH-2716

PR #725, opened 2019-07-30
Improve the performance of partitioned step

benas:
This PR is based on #716. It adds the minor changes requested in the code review.

The performance improvement can be measured using [this sample app](https://github.com/benas/spring-batch-sandbox/tree/master/batch2716) (See `result-*.md` files).

Resolves [BATCH-2716](https://jira.spring.io/browse/BATCH-2716).


benas:
Hi @NaanProphet ,

After further review of this PR, we had a couple of concerns:

* In this PR, we are loading all step executions from all job executions for a given job instance and doing the filtering in memory to get the last step execution
* The signature change introduced in `SimpleStepExecutionSplitter#isStartable` method is a breaking change (which requires deprecation and removal in 2 different releases)

After digging deeper, we thought it should be possible to improve the performance of getting the last step execution for a given job instance using a sql query instead of grabbing all step executions from the database and doing the filtering in memory. This is what we've done in #727 and the results are comparable to those of this PR. Hence, we merged the other PR as it does not introduce the couple of issues mentioned previously.

I'm closing this PR. Thank you anyway for your contribution!

Best regards,
Mahmoud

PR #726, opened 2019-09-01
Added @Nullable annotation to overriding methods.

mikeldpl:
Resolves BATCH-2839

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/step/PartitionStep.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/step/PartitionStep.java
@@ -32,6 +32,7 @@ import org.springframework.batch.core.partition.support.StepExecutionAggregator;
 import org.springframework.batch.core.step.NoSuchStepException;
 import org.springframework.batch.core.step.StepLocator;
 import org.springframework.batch.item.ExecutionContext;
+import org.springframework.lang.Nullable;
 
 /**
  * An extension of the {@link PartitionStep} that provides additional semantics
@@ -102,6 +103,7 @@ public class PartitionStep extends org.springframework.batch.core.partition.supp
 	/* (non-Javadoc)
 	 * @see org.springframework.batch.core.step.StepLocator#getStep(java.lang.String)
 	 */
+	@Nullable
 	@Override
 	public Step getStep(String stepName) throws NoSuchStepException {
 		JsrPartitionHandler partitionHandler =  (JsrPartitionHandler) getPartitionHandler();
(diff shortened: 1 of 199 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/jsr/step/PartitionStep.java
```diff
@@ -102,6 +103,7 @@ protected void doExecute(StepExecution stepExecution) throws Exception {
 	/* (non-Javadoc)
 	 * @see org.springframework.batch.core.step.StepLocator#getStep(java.lang.String)
 	 */
+	@Nullable
```
benas: This `@Nullable` here does not make sense as the implementation throws `NoSuchStepException` when no step is found with the given name. The issue is that the behaviour is not consistent across the framework: some implementations return `null` (like  [SimpleJob](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/job/SimpleJob.java#L117)) and some others throw `NoSuchStepException` like in here. I believe it is the contract `StepLocator#getStep` that should be fixed to throw an exception instead of returning null (in order to make it consistent with [`JobLocator`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/JobLocator.java#L42)), but my concern is that it would be a breaking change.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemReader.java
```diff
@@ -189,7 +191,8 @@ protected T doRead() throws Exception {
 	/**
 	 * @return next line (skip comments).getCurrentResource
 	 */
-	private String readLine() {
+	@Nullable
+	protected String readLine() {
```
benas: The visibility of `readLine` has been changed from `private` to `protected`. Why is that?

benas:
@mikeldpl Could you please update the year in the license headers of the changed files? Thank you upfront.

spring-batch-core/src/main/java/org/springframework/batch/core/jsr/step/PartitionStep.java
benas: It looks like the contract `StepLocator` has been changed incorrectly [here](https://github.com/spring-projects/spring-batch/commit/276acd1960b0e2578596a0d3507ad624ace9037c). I reverted that change and adapted the PR accordingly.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemReader.java
benas: Fixed in 230614182667378924ec557913df75b72af0ddc5 as I believe this should not change.

benas:
Rebased and merged as a7092a21e428cf904f210aff2682b518dc3649c5.

@mikeldpl Thank you for your PR!

spring-batch-core/src/main/java/org/springframework/batch/core/jsr/step/PartitionStep.java
mikeldpl: Yes, you are right. Thanks for the changes.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemReader.java
mikeldpl: It was done for my internal tests. Thanks for the change.

PR #727, opened 2019-09-03
Improve the performance of step partitioning

benas:
This PR improves the performance of splitting a step execution by 20x (See benchmark [here](https://github.com/benas/spring-batch-sandbox/tree/master/batch2716)). It moves the logic of finding the last step execution of a job instance to the database (instead of doing it in memory).

This PR is functionally the same as #725. However, it does **not**:

* Load all step executions and all job executions of a given job instance in memory to get the last step execution
* change the signature of `SimpleStepExecutionSplitter#isStartable` (which is a breaking change)

This PR resolves [BATCH-2716](https://jira.spring.io/browse/BATCH-2716).


mminella:
Rebased, polished, and merged.

PR #729, opened 2019-10-03
BATCH-2844: Support empty delimiter in DelimitedBuilder of FlatFileItemWriterBuilder

drumonii:
Previously, if supplying an empty delimiter `""`, to the `delimited().delimiter()` of `DelimitedBuilder` from the parent `FlatFileItemWriterBuilder`, it would be ignored and instead use the default delimiter `","`.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 the original author or authors.
+ * Copyright 2016-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
```diff
@@ -459,9 +458,7 @@ protected DelimitedBuilder(FlatFileItemWriterBuilder<T> parent) {
 					"A list of field names or a field extractor is required");
 
 			DelimitedLineAggregator<T> delimitedLineAggregator = new DelimitedLineAggregator<>();
-			if (StringUtils.hasLength(this.delimiter)) {
-				delimitedLineAggregator.setDelimiter(this.delimiter);
-			}
+			delimitedLineAggregator.setDelimiter(this.delimiter);
```
benas: We should at least check that `this.delimiter` is not `null` (which was part of `StringUtils.hasLength`). Otherwise the new test `testDelimitedOutputWithEmptyDelimiter ` fails with a NPE in `StringJoiner.java:122` when I pass `null`.

benas:
Rebased and merged. I added the null check. Thank you for the PR!

PR #731, opened 2019-10-03
Allow customisation of Deserializer for KafkaConsumer

Rouche:
Hi,

I have a case where i need a full integration test with EmbeddedKafka

To make the integration test work, i need to use MockSchemaRegistryClient from io.confluent, because our serializers are KafkaAvroSerializer/Deserializer.

Those 2 needs a schemaRegistry, and the only way to mock it is to explicitly create them with the mock.
```
    @Bean
    KafkaAvroDeserializer kafkaAvroDeserializer() {
        KafkaAvroDeserializer deserializer = new KafkaAvroDeserializer(schemaRegistryClient(), props.buildConsumerProperties());
        return deserializer;
    }
```
Exposing Deserializer in KafkaItemReaderBuilder make customization possible, since KafkaConsumer constructor allow it. (KafkaItemReader line 167)
This is not only for test case but any deserialiser could be configured to use Spring injected beans.

Producer do not have this problem since it goes through a KafkaTemplate

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java
@@ -49,6 +49,7 @@ import org.springframework.util.Assert;
  *
  * @author Mathieu Ouellet
  * @author Mahmoud Ben Hassine
+ * @author Jean-Francois Larouche
  * @since 4.2
  */
 public class KafkaItemReader<K, V> extends AbstractItemStreamItemReader<V> {
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java
@@ -30,6 +30,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.common.TopicPartition;
 
+import org.apache.kafka.common.serialization.Deserializer;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.batch.item.support.AbstractItemStreamItemReader;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@Rouche Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=731)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=731) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Rouche Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=731)!

benas:
Hi @Rouche ,

Thank you for this PR. Why do we need these setters? We can already specify avro key/value deserializer as well as the schema registry URL through consumer properties:

```java
Properties consumerProperties = new Properties();
consumerProperties.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, KafkaAvroDeserializer.class);
consumerProperties.put(AbstractKafkaSchemaSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "http://localhost:8081");

// create KakfaItemReader with these properties
```

If you want to mock the schema registry, you can use the `mock://` prefix as documented by confluent: https://github.com/confluentinc/schema-registry/blob/027aacfb20c9daf801068a5273ae6f2ac9f455d7/schema-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaSchemaSerDeConfig.java#L52-L55.

Rouche:
Like i explained. Its because of Avro.

benas:
> Like i explained. Its because of Avro.

What do you mean by "Its because of Avro"? What I'm saying is that you can use **Avro** key/value serializer/deserializer and configure them through consumer properties along with the url to schema registry (similar to the example from [official confluent docs](https://docs.confluent.io/current/schema-registry/schema_registry_tutorial.html#java-consumers) which does **not** instantiate the deserializer and pass the registry to it at construction time). The registry can be mocked with `mock://` if needed. Have you tried this approach?

Rouche:
Unfortunatelly, i am no longer on the project and Kafka have sadly been abandoned.

i did not tryed your solution maybe it was not available at that time.  The only way was to do it that way.

I think the customization should be available. Not only for unit tests. 

But the decision is yours.

benas:
I have  no objection against adding these setters if they are needed, but they aren't. The kafka way of configuring serializers/deserializers is through consumer properties, so we prefer following the same approach for consistency. Having two ways of doing the same thing would be confusing (which one takes precedence, what is the preferred way, etc).

Thank you for your PR anyway.

PR #733, opened 2019-10-04
BATCH-2837: Support empty or no comments in FlatFileItemReaderBuilder

drumonii:
Previously, if supplying empty comments or no comments to `comments()` of the `FlatFileItemReaderBuilder`, the comments would be ignored and instead use the default comments `"#"` provided by the `FlatFileItemReader`.

Resolves BATCH-2837

### DIFF
--- a/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/file/builder/FlatFileItemReaderBuilderTests.java
+++ b/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/file/builder/FlatFileItemReaderBuilderTests.java
@@ -274,6 +274,51 @@ public class FlatFileItemReaderBuilderTests {
 	}
 
 	@Test
+	public void testEmptyComments() throws Exception {
+		FlatFileItemReader<Foo> reader = new FlatFileItemReaderBuilder<Foo>()
+				.name("fooReader")
+				.resource(getResource("1,2,3\n4,5,6"))
+				.comments(new String[]{})
+				.delimited()
+				.names("first", "second", "third")
+				.targetType(Foo.class)
+				.build();
+
+		reader.open(new ExecutionContext());
+		Foo item = reader.read();
+		assertEquals(1, item.getFirst());
+		assertEquals(2, item.getSecond());
+		assertEquals("3", item.getThird());
+		item = reader.read();
+		assertEquals(4, item.getFirst());
+		assertEquals(5, item.getSecond());
+		assertEquals("6", item.getThird());
+		assertNull(reader.read());
+	}
+
+	@Test
+	public void testDefaultComments() throws Exception {
+		FlatFileItemReader<Foo> reader = new FlatFileItemReaderBuilder<Foo>()
+				.name("fooReader")
+				.resource(getResource("1,2,3\n4,5,6"))
+				.delimited()
+				.names("first", "second", "third")
+				.targetType(Foo.class)
+				.build();
+
+		reader.open(new ExecutionContext());
+		Foo item = reader.read();
+		assertEquals(1, item.getFirst());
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/test/java/org/springframework/batch/item/file/builder/FlatFileItemReaderBuilderTests.java
```diff
@@ -273,6 +273,51 @@ public void testComments() throws Exception {
 		assertNull(reader.read());
 	}
 
+	@Test
+	public void testEmptyComments() throws Exception {
```
benas: I agree on the fix, however this test does not test the case reported in BATCH-2837 because the input resource does not contain a line with the default comment prefix `#`.  Here is what I mean (I changed the target type to `String` since it is easier to test for this case):

```java
@Test
public void testEmptyComments() throws Exception {
	FlatFileItemReader<String> reader = new FlatFileItemReaderBuilder<String>()
			.name("itemReader")
			.resource(getResource("line1\n#line2"))
			.comments(new String[]{})
			.lineMapper(new PassThroughLineMapper())
			.build();

	reader.open(new ExecutionContext());
	String item = reader.read();
	assertEquals("line1", item);
	item = reader.read();
	assertEquals("#line2", item); // "#line2" should not be considered as comment
	assertNull(reader.read());
	reader.close();
}
```

Do you agree? BTW, this test passes with this PR.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/file/builder/FlatFileItemReaderBuilderTests.java
```diff
@@ -273,6 +273,51 @@ public void testComments() throws Exception {
...
+		assertEquals("3", item.getThird());
+		item = reader.read();
+		assertEquals(4, item.getFirst());
+		assertEquals(5, item.getSecond());
+		assertEquals("6", item.getThird());
+		assertNull(reader.read());
+	}
+
+	@Test
+	public void testDefaultComments() throws Exception {
```
benas: Same here, to test default comment prefixes, I would use a resource having a line starting with `#` and assert that it is ignored by default:

```java
@Test
public void testDefaultComments() throws Exception {
	FlatFileItemReader<String> reader = new FlatFileItemReaderBuilder<String>()
			.name("itemReader")
			.resource(getResource("line1\n#line2"))
			.lineMapper(new PassThroughLineMapper())
			.build();

	reader.open(new ExecutionContext());
	String item = reader.read();
	assertEquals("line1", item);
	assertNull(reader.read()); // line "#line2" should be ignored by default
	reader.close();
}
```

However, this test fails not due to this PR, but due to another bug that I discovered while testing this PR, see [BATCH-2861](https://jira.spring.io/browse/BATCH-2861) and the suggested fix in #747. So please update the test and I will merge the PRs in the right order.

benas: The method `FlatFileItemReaderBuilder#comments` currently adds comments to the default empty list (it should have been called `addComments` IMO). I believe it should be rather:

```java
public FlatFileItemReaderBuilder<T> comments(String[] comments) {
	this.comments = Arrays.asList(comments);
	return this;
}
```

to be able to override the default comments (as described in BATCH-2837).

@drumonii Can you please add this change along with those in #747 here? Otherwise let me know if you agree and I will do it.

benas:
@drumonii You might have seen some comments from my part here which I deleted (after further digging, I found they were out of scope of this issue). Sorry for the noise.

This PR LGTM. Thank you for your contribution!

PR #735, opened 2019-10-10
Add connection parameter to AbstractCursorItemReader#cleanupOnClose

benas:
This change makes it possible to (re)use the same connection used in the
openCursor method to clean up resources when the reader is closed.

Resolves [BATCH-1899](https://jira.spring.io/browse/BATCH-1899).


benas:
Rebased and merged as e7ea16c9bc8bb3cbe174c8b190fff206b957e1bc.

PR #736, opened 2019-10-11
Fix typos, code style and trim pictures in docs

michalfotyga:
Just fixed a typo and cut top part of two pictures that had too much white space there.


spring-batch-docs/asciidoc/common-patterns.adoc
```diff
@@ -531,7 +531,7 @@ public Trade read() throws Exception {
             return t; // Record must end with footer
         }
     }
-    Assert.isNull(t, "No 'END' was found.");
+    Assert.notNull(t, "No 'END' was found.");
```
benas: To be consistent, we need to update the code accordingly. But when I apply this change to `MultiLineTradeItemReader`, an `IllegalArgumentException` is thrown in `org.springframework.batch.sample.iosample.MultiLineFunctionalTests#testJob` (the test still passes though..). This is out of scope of this PR, so please revert this change for now. The rest LGTM.

michalfotyga:
Nice catch @benas 👏 I have reverted that part.

benas:
LGTM now. Thank you!

PR #738, opened 2019-10-18
BATCH-2849 Account For Different Whitespaces In Keyword Removal

BenjaminHetzjelli:
Modify `removeKeyWord(...)` such that the keyword is removed regardless
of what kind of whitespace follows. This is especially useful for those
who read in SQL from a file which has been formatted such that keywords
live on their own lines.

Add unit tests for trimming whitespace.

Resolves [BATCH-2849](https://jira.spring.io/browse/BATCH-2849)

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2019 the original author or authors.
+ * Copyright 2006-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
@@ -242,9 +242,9 @@ public abstract class AbstractSqlPagingQueryProvider implements PagingQueryProvi
 
 	private String removeKeyWord(String keyWord, String clause) {
 		String temp = clause.trim();
-		String keyWordString = keyWord + " ";
-		if (temp.toLowerCase().startsWith(keyWordString) && temp.length() > keyWordString.length()) {
-			return temp.substring(keyWordString.length());
+		int len = keyWord.length();
+		if (temp.toLowerCase().startsWith(keyWord) && Character.isWhitespace(temp.charAt(len)) && temp.length() > len + 1) {
+			return temp.substring(len + 1);
 		}
 		else {
 			return temp;
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


BenjaminHetzjelli:
Hello @benas. I notice you are the one marking most of the recent pull requests as ready to review etc. I'm just wondering if I missed a step in the contribution guidelines. If it's just a case of this being a low priority change for a relatively infrequent use case that's fine. I just wanted to make sure that I didn't forget something.

benas:
Hi @BenjaminHetzjelli 

Thank you for the PR! I will let you know in which release BATCH-2849 will be included.

> I'm just wondering if I missed a step in the contribution guidelines.

The PR LGTM at first glance. Don't hesitate to update the year in the license header and add your name to the authors list, see: https://github.com/spring-projects/spring-batch/blob/master/CONTRIBUTING.md#update-license-header-to-modified-files-as-necessary

BenjaminHetzjelli:
Thank you!

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProviderTests.java
```diff
@@ -104,6 +104,51 @@ public void testGenerateJumpToItemQueryForFirstPageWithMultipleSortKeys() {
...
+		
+		
+		assertEquals("id, 'yes', false", pagingQueryProvider.getSelectClause());
+		assertEquals("test.verification_table", pagingQueryProvider.getFromClause());
+		assertEquals("TRUE", pagingQueryProvider.getWhereClause());
+	}
+
+	@Test
+	public void testRemoveKeyWordsFollowedByTabChar() {
+		String selectWithSpace = "SELECT\tid, 'yes', false";
```
benas: This variable should be named `selectWithTab`, do you agree?

```suggestion
		String selectWithTab = "SELECT\tid, 'yes', false";
```

Same for `fromWithSpace` and `whereWithSpace`. But I think it is better to use something like `selectClause`, `fromClause` and `whereClause` since the test name is already explicit enough about the case we are testing. Wdyt?

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProviderTests.java
```diff
@@ -104,6 +104,51 @@ public void testGenerateJumpToItemQueryForFirstPageWithMultipleSortKeys() {
...
+		assertEquals("id, 'yes', false", pagingQueryProvider.getSelectClause());
+		assertEquals("test.verification_table", pagingQueryProvider.getFromClause());
+		assertEquals("TRUE", pagingQueryProvider.getWhereClause());
+	}
+
+	@Test
+	public void testRemoveKeyWordsFollowedByNewLineChar() {
+		String selectWithSpace = "SELECT\nid, 'yes', false";
+		String fromWithSpace = "FROM\ntest.verification_table";
+		String whereWithSpace = "WHERE\nTRUE";
```
benas: Same here, IMO `selectClause`, `fromClause` and `whereClause` are ok since the test method name is explicit enough.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProvider.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2019 the original author or authors.
```
benas: Please update the year to 2020.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/database/support/AbstractSqlPagingQueryProviderTests.java
BenjaminHetzjelli: You are certainly correct. I will get this fixed.

benas:
LGTM now. Rebased, squashed and merged as 7f01bf78cfce89e446f57d7f0410ccb14be4646b. Thank you!

PR #739, opened 2019-10-31
BATCH-2480 Regression due to the fix of BATCH-2442

wklaus:
This PR fixes the problem described in BATCH-2480 which was introduced with BATCH-2442. An aditional regression test is in teh PR to close the gap in the test suite.


benas:
Rebased and merged as cc6f60871e7515f539b9876f57962d9baae5c743. Thank you for the PR!

PR #740, opened 2019-11-06
Add https schemas

quaff:
Like https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/resources/META-INF/spring.schemas


benas:
LGTM. Thanks!

PR #742, opened 2019-11-20
Fix typo

quaff:


spring-batch-docs/asciidoc/spring-batch-intro.adoc
```diff
@@ -436,7 +436,7 @@ _4. Addition of a Processing Indicator_
 This involves the addition of a new column to the input table, which acts as an
 indicator. As a preprocessing step, all indicators are marked as being non-processed.
 During the record fetch stage of the batch application, records are read on the condition
-that that record is marked as being non-processed, and once they are read (with lock),
```
Buzzardo: "that that" is correct here. The first "that" is a conjunction. The second "that" is a pronoun. It's rare, but it's legal and correct in this case. If someone really hates a double word, it could be "that the" or "that each".

benas:
Based on Jay's comment, I'm closing this PR. Thanks anyway!

PR #743, opened 2019-11-21
Modifying col SERIALIZED_CONTEXT to LONGTEXT from TEXT

sapatel0912:
My processing data contains json string and some times json string length got exceed existing TEXT type limit which is 65535 characters. This causes the failure of whole job. So, I think it would be good if we increase the column type to LONGTEXT which will practically rare to  exceed in any case.


pivotal-issuemaster:
@sapatel0912 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=743)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=743) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sapatel0912 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=743)!

mminella:
What are you putting in the `ExecutionContext` that requires so much space?  We normally don't recommend using the context in ways that require that much so I'd rather not change the framework without understanding the use case first.

sapatel0912:
@mminella I have one `CDC` application on the `Mysql` server and one spring batch application both are constantly running. I capture `CDC` data to convert it into `JSON` and store it to one `cdctable`. It will have a separate row for each table event. The `cdctable` has one column called `UpdatedTableName` which stores the name of the table which got an update. So, whenever the batch's job executed it will fetch all the updated `JSON` data group with `UpdatedTableName` process the data and will send it to the writer to store the data. So to send the data into the writer Step I prepare one ArrayList<ProcessedDataVO> put it in `ExecutionContext`. In normal cases, everything works fine. However, if for any reason Spring batch stopped working but the `CDC` application is still going on then it will continue to store data in the `cdctable`. Now after sometime when the Spring batch application will up again, the reader will pick the first 2000 unprocessed `JSON` data and start processing. As we know `CDC` has whole changed data as well as old data so the `JSON` becomes pretty much big, especially field like `description` may have longer strings. Eventually, it tends to exceed the TEXT size of the `SERIALIZED_CONTEXT`  column and resulting in the failure of job execution. I am not facing this issue constantly but in mention scenario, I am facing the issue. If it's not recommended to change in the framework itself then it will be really helpful to change it via a configuration file. So one can change the column's type based on the need. I am using the `Tasklets` approach to implement the job. So, I think I need to use `ExecutionContext` to send data between different steps. If I can achieve this in any other way then I am open to it.

mminella:
We do not recommend storing actual data in the `ExecutionContext` like this.  This sounds like a use case for the processed indicator pattern.  Using this pattern, you use an additional column on the table that indicates the state.  So in your case, when the step starts, you'd use a listener to tag all of the records that need to be processed within this batch as in process.  Then, you'd have another listener that tags each record as they are done being processed as complete.  This allows the state to be external to the `ExecutionContext` and is a more scalable solution.

guptaBrothers:
It is recommended to have context light weight for chunck step/ multi-step Job, since there will more number of context switiching. Otherwise job execution performance will be impacted heavily.

In you use case , either you go as Micheal suggested using indecator or make use of history table during housekeeping step before actual processing.

sapatel0912:
I think I got my answer. Thank you for your valuable feedback @mminella and @guptaBrothers.

PR #746, opened 2019-11-27
Fix wrong indent

quaff:


benas:
I'm accepting this PR for now but please beware that we prefer to handle this internally. Thank you.

PR #748, opened 2019-11-29
Update schema-appendix.adoc to fix wrong names

quaff:
BATCH_EXECUTION_CONTEXT should be BATCH_JOB_EXECUTION_PARAMS
EXECUTION_ID should be JOB_EXECUTION_ID

Please check `Frequency` if it's up to date.

### DIFF
--- a/spring-batch-docs/asciidoc/schema-appendix.adoc
+++ b/spring-batch-docs/asciidoc/schema-appendix.adoc
@@ -396,7 +396,6 @@ about indexing:
 |Default Table Name|Where Clause|Frequency
 |BATCH_JOB_INSTANCE|JOB_NAME = ? and JOB_KEY = ?|Every time a job is launched
 |BATCH_JOB_EXECUTION|JOB_INSTANCE_ID = ?|Every time a job is restarted
-|BATCH_EXECUTION_CONTEXT|EXECUTION_ID = ? and KEY_NAME = ?|On commit interval, a.k.a. chunk
 |BATCH_STEP_EXECUTION|VERSION = ?|On commit interval, a.k.a. chunk (and at start and end of
             step)
 |BATCH_STEP_EXECUTION|STEP_NAME = ? and JOB_EXECUTION_ID = ?|Before each step execution


benas:
Is this related to https://jira.spring.io/browse/BATCH-2853?

quaff:
@benas Yes.

benas:
ok thanks, I added a link to this PR in BATCH-2853.

benas:
> BATCH_EXECUTION_CONTEXT should be BATCH_JOB_EXECUTION_PARAMS

This is actually not correct. `BATCH_EXECUTION_CONTEXT` has been replaced with `BATCH_STEP_EXECUTION_CONTEXT` and `BATCH_JOB_EXECUTION_CONTEXT` in https://github.com/spring-projects/spring-batch/commit/d272105479896ea543ce971fbb0298809b450df2#diff-9df3d792ac9c1d09e598d04018ee1f6eL71-L80. And based on the frequency "On commit interval, a.k.a. chunk" mentioned in the appendix, it is definitely not `BATCH_JOB_EXECUTION_PARAMS`: this table is not queried/updated at each commit interval. It is `BATCH_STEP_EXECUTION_CONTEXT` that is updated at this frequency, here is an excerpt from the [ExecutionContext](https://docs.spring.io/spring-batch/docs/4.2.x/reference/html/domain.html#executioncontext) section:

```
They are two different ExecutionContexts. The one scoped to the Step is saved at every commit point in the Step, whereas the one scoped to the Job is saved in between every Step execution.
```

@quaff Could you please update the PR accordingly?

quaff:
@benas `BATCH_STEP_EXECUTION_CONTEXT` and `BATCH_STEP_EXECUTION_CONTEXT` are queried by primary key, so they should be excluded from this table.

benas:
Indeed, `BATCH_STEP_EXECUTION_CONTEXT` and `BATCH_JOB_EXECUTION_CONTEXT` are queried by primary keys (which are indexed by default), so this entry can be removed from the table. Thanks for the PR!

PR #749, opened 2019-12-04
Update BeanWrapperFieldExtractor.java to support delimiter and quote character.

ADANSAR:
Supporting the quote character like the delimited tokenizer does.
If the value of the item contains the delimiter, the quote character is added to the value of the constructed array.
the quote character can be used to extend a field across line endings or to enclose a String which contains the delimiter.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/BeanWrapperFieldExtractor.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/BeanWrapperFieldExtractor.java
@@ -27,16 +27,78 @@ import org.springframework.util.Assert;
 
 /**
  * This is a field extractor for a java bean. Given an array of property names,
- * it will reflectively call getters on the item and return an array of all the
- * values.
+ * a delimiter and a quote character, it will reflectively call getters on the
+ * item and return an array of all the values with eventually the quote character.
  * 
  * @author Dan Garrette
+ * @author ADANSAR Mohamed
  * @since 2.0
  */
 public class BeanWrapperFieldExtractor<T> implements FieldExtractor<T>, InitializingBean {
 
+	/**
+	 * Convenient constant for the common case of a comma delimiter.
+	 */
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ADANSAR Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=749)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-batch&pullRequestId=749) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ADANSAR Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-batch&pullRequestId=749)!

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/BeanWrapperFieldExtractor.java
```diff
@@ -54,7 +116,12 @@ public Object[] extract(T item) {
 
 		BeanWrapper bw = new BeanWrapperImpl(item);
 		for (String propertyName : this.names) {
-			values.add(bw.getPropertyValue(propertyName));
+			Object myObject = bw.getPropertyValue(propertyName);
+			if (myObject instanceof String && ((String) myObject).contains(delimiter)) {
```
fmbenhassine, line 120: Analysing the content of objects and adding quotes is not the responsibility of the `BeanWrapperFieldExtractor`. If we do it here, we should also do it in other `FieldExtractor`s like `RecordFieldExtractor`.

This could be added as a feature by the user in a subclass if needed, like `SmartBeanWrapperFieldExtractor` or `DelimiterAwareBeanWrapperFieldExtractor`.

fmbenhassine:
Thank you for this PR! While I understand the use case, I believe it is not the right place to add this feature (as explained in the inline comment). This feature can be added in a subclass if needed.

For this reason, I am closing this PR. Thank you for your contribution anyway!

PR #3637, opened 2019-12-27
FEATURE: open SimpleJobLauncher for extension

hosuaby:
CONTEXT: In my project I was needed a custom `JobLauncher` implementation that delays execution of the job until some condition become true. In order to achieve it, I copied all code from `SimpleJobLauncher`. This work could be done more easily if `SimpleJobLauncher` was open for extension.

In this PR, I added two protected methods `createJobExecution` and `launchJobExecution` to class  `SimpleJobLauncher`. Creation of `JobExecution` is now decoupled from its execution.


fmbenhassine:
Good idea! It took time but it is finally here, see 76e723e41939b1ab6910f9ce8d61053abb1d0575. A lot has changed since this PR was opened and I was not able to rebase it correctly.

Moreover, there are some changes that I preferred omitting:

- The job execution creation should not be open for extension as it contains a lot of internal logic that is not designed to be customized (closed for modification)
- I don't see a real added value from a user PoV of the `JobExecutionTask` added in this PR since the user will have to extend the class and override `launchJobExecution` anyway.

Other than that, I think this is nice addition, and v6 seems like a good time to include it. I gave you credits in 76e723e41939b1ab6910f9ce8d61053abb1d0575, thank you!

Issue #3638, opened 2019-12-28
Order of Transition Elements Does Not Adhere to Documentation


kmb385:
In the documentation it provides this example:

```
@Bean
public Job job() {
return this.jobBuilderFactory.get("job")
.start(stepA())
.on("*").to(stepB())
.from(stepA()).on("FAILED").to(stepC())
.end()
.build();
}
```
And then states:

> The framework automatically orders transitions from most specific to least specific.
> This means that, even if the ordering were swapped for "stepA" in the example above, an ExitStatus of "FAILED" would still go to "stepC".

I have noticed the ordering of the transition elements seems to matter.  An exception thrown in `stepA` still sends execution of the job to `StepB`.  If I reorder the transition elements to:

```
@Bean
public Job job() {
return this.jobBuilderFactory.get("job")
.start(stepA())
.on("FAILED").to(stepC())
.from(stepA()).on("*").to(stepB())
.end()
.build();
}
```
The job routes correctly causing a failed `stepA` to direct the flow to `stepB`.  Either I'm interpreting the documentation incorrectly, there is an issue with my simple example or the framework is not behaving according to the documentation.

After roughly an hour of testing, it seems to me the order of the transition elements must be specified from most specific to least specific.

kmb385:
If determined to be valid, I submitted this pull request to update the documentation: https://github.com/spring-projects/spring-batch/pull/3639

kmb385:
If this is determined to be a valid discrepancy another option would be to modify the framework to behave as described in the documentation.

mminella:
This is not a documentation issue.  This is a bug in the java configuration.  This works as expected when using XML configuration.  I'm working on a fix now.

kmb385:
@mminella Thanks for taking a look.  I'll keep following the issue, I'm interested in seeing the code behind the fix.

PR #3639, opened 2019-12-28
Update step.adoc

kmb385:
These are updates for Github issue 3638 found here: https://github.com/spring-projects/spring-batch/issues/3638

This aligns the documentation with the behavior of the framework, which seems to require the transition elements to be specified in most to least specific order, contrary to the modified sections of the documentation.


mminella:
I'm closing this PR because the issue is actually a bug in Spring Batch and not a documentation issue.  Thanks for identifying it though.  I've kept the issue open so that we can resolve it via a proper fix.

PR #3645, opened 2020-01-14
Added comparator for state transitions when using java config

### LINKED ISSUE #3638: Order of Transition Elements Does Not Adhere to Documentation

mminella:
Spring Batch orders the transitions as it goes from state to state based
on specificity.  The XML configuration has always had this
functionality.  However, when creating the JSR-352 implementation, the
mechanism for which this occured was refactored.  That occured at about
the same time as the java builders were introduced.  Because of this
crossing of paths, the java configuration option for defining jobs has
never correctly sorted the transitions.  This PR applys the sorting
algorithm to the java configuration, making XML and java configuration
behave the same.

Resolves #3638

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2013 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


benas:
LGTM. Rebased and merged as 8e8f9c8101b52b64050d8e92363de2d28319e3b0. Backported to 4.1.4 and 4.2.2 as well. Thank you Michael!

PR #3648, opened 2020-01-17
Add documentation for how an ItemProcessor.process can never receive …

mouyang:
…a null item.

I realized this while working on a task.  I think it could prompt someone to save an unnecessary null check in an ItemProcessor.  Let me know if my logic is incorrect or could be better worded.


benas:
LGTM, rebased and merged as f37fae670b454699092abd0494a29706af6a271f. Thank you for your PR!

PR #3652, opened 2020-01-29
Updated email addresses from Pivotal to VMware

mminella:


benas:
Fixed in b83c81e680e38f6504c1dd8600724e2d221a6acb.

PR #3653, opened 2020-01-29
Use local storage rather than a cookie

Buzzardo:
I learnt that local storage is the preferred way to save a user's preferences. A cookie is sent to the server on every request, which makes for needless traffic. I "borrowed" the idea from the Spring Framework docs.


benas:
LGTM. rebased and merged. Thanks!

PR #3655, opened 2020-02-02
Add usePersist chain method to JpaItemWriterBuilder

mustaev-ruslan:
Hi!

I found that commit "BATCH-2462 Allow JpaItemWriter to support persist rather than merge for improved performance." have not appropriate change in builder.
I add new chain method to builder "usePersist" by analogy from original commit.


benas:
LGTM. rebased and merged as bd016a94440025bc13070ac7e1f929570f7d03df. Thank you!

PR #3662, opened 2020-02-12
[BATCH-3654] Add topic to KafkaItemWriter

ta7uw:
This PR makes it possible to set a topic other than the default topic of `KafkaTemplate` to `KafkaItemWriter`.
We can use `KafkaItemWriter` as a thread-safe kafka producer by setting a topic.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2019-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import org.springframework.batch.item.ItemWriter;
 import org.springframework.batch.item.KeyValueItemWriter;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * <p>
@@ -28,6 +29,7 @@ import org.springframework.util.Assert;
  * </p>
  *
  * @author Mathieu Ouellet
+ * @author Takaaki Shimbo
  * @since 4.2
  *
  */
@@ -35,20 +37,28 @@ public class KafkaItemWriter<K, T> extends KeyValueItemWriter<K, T> {
 
 	private KafkaTemplate<K, T> kafkaTemplate;
 
+	/**
+	 * @since 4.3
+	 */
+	private String topic;
+
 	@Override
 	protected void writeKeyValue(K key, T value) {
 		if (this.delete) {
-			this.kafkaTemplate.sendDefault(key, null);
+			this.kafkaTemplate.send(topic, key, null);
 		}
 		else {
-			this.kafkaTemplate.sendDefault(key, value);
+			this.kafkaTemplate.send(topic, key, value);
 		}
 	}
 
 	@Override
 	protected void init() {
 		Assert.notNull(this.kafkaTemplate, "KafkaTemplate must not be null.");
-		Assert.notNull(this.kafkaTemplate.getDefaultTopic(), "KafkaTemplate must have the default topic set.");
+		if (StringUtils.isEmpty(topic)) {
+			Assert.notNull(this.kafkaTemplate.getDefaultTopic(), "KafkaTemplate must have a topic set.");
+			topic = this.kafkaTemplate.getDefaultTopic();
+		}
 	}
 
 	/**
@@ -57,5 +67,15 @@ public class KafkaItemWriter<K, T> extends KeyValueItemWriter<K, T> {
 	 */
 	public void setKafkaTemplate(KafkaTemplate<K, T> kafkaTemplate) {
 		this.kafkaTemplate = kafkaTemplate;
+	}
+
+	/**
+	 * Set the topic that the record will be appended to.
+	 *
+	 * @param topic name
+	 * @since 4.3
+	 */
+	public void setTopic(String topic) {
+		this.topic = topic;
 	}
 }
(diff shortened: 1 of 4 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemWriter.java
```diff
@@ -28,27 +29,36 @@
...
-			this.kafkaTemplate.sendDefault(key, value);
+			this.kafkaTemplate.send(topic, key, value);
 		}
 	}
 
 	@Override
 	protected void init() {
 		Assert.notNull(this.kafkaTemplate, "KafkaTemplate must not be null.");
-		Assert.notNull(this.kafkaTemplate.getDefaultTopic(), "KafkaTemplate must have the default topic set.");
+		if (StringUtils.isEmpty(topic)) {
```
mminella: To make this be a non-breaking change (and to simplify the user's experience if they don't need this functionality), we'd want this to be an optional requirement.  Can you please update the PR to only use the topic name if it has been set (and use the default if it has not).

ta7uw: If the topic name has not been set, the default topic of `KafkaTemplate` will be regard as a topic that the records will be appended to.

https://github.com/spring-projects/spring-batch/blob/03ef7e0c75c55a5ce3e573aab94739ffd3a8c112/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemWriter.java#L60

> Can you please update the PR to only use the topic name if it has been set (and use the default if it has not).

I think that this PR meets this requirements.
If it does not meet , please tell me how to fix it.

ta7uw: @mminella 
Would you please check the above?

benas:
> We can use KafkaItemWriter as a thread-safe kafka producer by setting a topic.

How is that related to thread safety? The kafka template is already thread-safe and hence the writer is thread-safe as well.

What is the added value of writing this:

```java
KafkaTemplate kafkaTemplate = ..;
KafkaItemWriter kafkaItemWriter = new KafkaItemWriter();
kafkaItemWriter.setKafkaTemplate(kafkaTemplate);
kafkaItemWriter.setTopic("my-topic"); // requested feature
```

over this:

```java
KafkaTemplate kafkaTemplate = ..;
kafkaTemplate.setDefaultTopic("my-topic");
KafkaItemWriter kafkaItemWriter = new KafkaItemWriter();
kafkaItemWriter.setKafkaTemplate(kafkaTemplate);
```

?

My concern with this feature request is that it will make the `KafkaItemWriter` inconsistent with [`JmsItemWriter`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/jms/JmsItemWriter.java) and [`AmqpItemWriter`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/amqp/AmqpItemWriter.java) where we use the default destination/exchange to send messages (there is no way to specify a destination/exchange other than the default ones set on the template).

The reason for relying on the default destination/exchange in `JmsItemWriter`/`AmqpItemWriter` is to avoid exposing setters for all other properties of the template. For example in this PR, if we add a setter for the topic, why don't we add a setter for `closeTimeout`, `micrometerEnabled`, etc.. ?

So unless I'm missing something, I believe this feature request will introduce inconsistency and confusion for no real benefit:

* The code needed with the current version is almost the same as with the feature request (as shown previously)
* The kafka template is already thread safe so the writer is thread safe as well

@ta7uw Do you agree?

ta7uw:
@benas 
I don't want to set a default topic to `KafkaTemplate` because `setDefaultTopic` method is a public method. 
The default topic might be changed. 

https://github.com/spring-projects/spring-kafka/blob/d76781d2168b10f4fd9d6aa568b9a03455eda293/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L214-L221

I always use the sending method with a specific topic in `KafkaTemplate`.

https://github.com/spring-projects/spring-kafka/blob/d76781d2168b10f4fd9d6aa568b9a03455eda293/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L366-L369

Like this, I also don't want to set a default topic to `KafkaItemWriter`.
The default topic that is set in initializing might be changed during a batch processing by changing `KafkaTemplate` 's default topic.

I think `KafkaItemWriter` should have an immutable topic.

benas:
> I don't want to set a default topic to KafkaTemplate because setDefaultTopic method is a public method.
The default topic might be changed.

Well you don't want to set a default topic on the template but even with this PR you would still need to [set it](https://github.com/spring-projects/spring-batch/pull/3662/files#diff-18d7c693b5d863d3bb45f5d94f4c8bf3R59) in case a topic is not provided via the setter.

Moreover, you say the reason is that `setDefaultTopic` method is a public, but the setter suggested here is public as well and one might change the topic on the writer afterwards.

> I think KafkaItemWriter should have an immutable topic.

If we really want to make the topic immutable, we need to pass it at construction time without providing a setter for it and by creating the kafka template internally (do not provide a public setter for the template as [it is now](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemWriter.java#L58), which would be a breaking change).

That said, you can always extend the writer and override `writeKeyValue` to send messages to a given topic:

```java
public class MyKafkaItemWriter<K, V> extends KafkaItemWriter<K, V> {

	private String topic;
	private KafkaTemplate<K, V> kafkaTemplate;
	// setters for topic + template

	@Override
	protected void writeKeyValue(K key, V value) {
		this.kafkaTemplate.send(this.topic, key, value);
	}
}
```

Again, my concern with this addition is that it will make the `KafkaItemWriter` inconsistent with other message writers as explained in my previous comment. So since it's possible to achieve the desired feature through an extension point (the protected `writeKeyValue` method), I prefer keeping things consistent on the framework side.

Thank you for your PR anyway!

Issue #3647, opened 2020-01-15
Meta-annotate @SpringBatchTest with @ExtendWith(SpringExtension.class)


sbrannen:
Similar to how many Spring Boot test annotations are meta-annotated with `@ExtendWith(SpringExtension.class)` -- for example, `@SpringBootTest`, `@WebMvcTest`, etc.) -- I recommend that `@SpringBatchTest` be meta-annotated with `@ExtendWith(SpringExtension.class)` in order to simplify configuration when writing tests with JUnit Jupiter.

dimitrisli:
@benas could I pick this one up? The @ExtendWith() will require junit-jupiter-api dependency, is that ok?

benas:
@dimitrisli I have already a branch for that: https://github.com/benas/spring-batch/tree/GH-3647 (See https://github.com/benas/spring-batch/commit/2fc8f6e66e9fb119a93e406c2f779b0c3dc594d0). I just did not open a PR yet. But if you want to do it, go ahead you are welcome!

benas:
Since github added the ability to create draft PRs recently, I created a draft PR for this here: https://github.com/spring-projects/spring-batch/pull/3697.

> But if you want to do it, go ahead you are welcome!

@dimitrisli I created the draft PR to avoid duplicate efforts, but feel free to review or test the change. Any contribution is welcome! Thank you upfront.

dimitrisli:
@benas no problem at all, please do continue on this

PR #3697, opened 2020-04-18
Meta-annotate @SpringBatchTest with @ExtendWith(SpringExtension.class)

benas:
This PR meta-annotates `@SpringBatchTest` with `@ExtendWith(SpringExtension.class)` so it can be used in a JUnit Jupiter test without having to extend the test class with `SpringExtension`.

Issue #3647

### DIFF
--- a/spring-batch-test/src/main/java/org/springframework/batch/test/context/SpringBatchTest.java
+++ b/spring-batch-test/src/main/java/org/springframework/batch/test/context/SpringBatchTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 the original author or authors.
+ * Copyright 2018-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,11 +22,14 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.junit.jupiter.api.extension.ExtendWith;
+
 import org.springframework.batch.test.JobLauncherTestUtils;
 import org.springframework.batch.test.JobRepositoryTestUtils;
 import org.springframework.batch.test.JobScopeTestExecutionListener;
 import org.springframework.batch.test.StepScopeTestExecutionListener;
 import org.springframework.test.context.TestExecutionListeners;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
 
 /**
  * Annotation that can be specified on a test class that runs Spring Batch based tests.
@@ -45,7 +48,7 @@ import org.springframework.test.context.TestExecutionListeners;
  * </li>
  * </ul>
  * <p>
- * A typical usage of this annotation is like:
+ * A typical usage of this annotation with JUnit 4 is like:
  *
  * <pre class="code">
  * &#064;RunWith(SpringRunner.class)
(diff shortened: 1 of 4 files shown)


benas:
@sbrannen Hi Sam,

Could you please take a look at this PR? Let me know if it needs any updates.

Many thanks upfront!

Best regards,
Mahmoud

spring-batch-test/src/main/java/org/springframework/batch/test/context/SpringBatchTest.java
```diff
@@ -78,7 +81,40 @@
  *
  * }
  * </pre>
+ * 
+ * For JUnit 5, this annotation can be used without having to extend the test class
+ * with {@link SpringExtension} as it is meta-annotated with {@code @ExtendWith(SpringExtension.class)}:
```
sbrannen, line 86: For the sake of clarity, I would suggest rewording this as follows.

```suggestion
 * For JUnit 5, this annotation can be used without having to manually register the
 * {@link SpringExtension} since {@code @SpringBatchTest} is meta-annotated with
 * {@code @ExtendWith(SpringExtension.class)}:
```

spring-batch-test/src/test/java/org/springframework/batch/test/SpringBatchTestJUnit4Tests.java
```diff
@@ -46,14 +46,14 @@
...
- * Test cases for usage of {@link SpringBatchTest} annotation.
+ * Test cases for usage of {@link SpringBatchTest} annotation with JUnit 4.
  *
  * @author Mahmoud Ben Hassine
  */
 @RunWith(SpringRunner.class)
 @SpringBatchTest
-@ContextConfiguration(classes = SpringBatchTestTests.JobConfiguration.class)
-public class SpringBatchTestTests {
+@ContextConfiguration(classes = SpringBatchTestJUnit4Tests.JobConfiguration.class)
```
sbrannen, line 55: Just as a side note, there is no reason to explicitly configure `JobConfiguration` here since `spring-test` will automatically find any static nested `@Configuration` class by default.

```suggestion
@ContextConfiguration
```

spring-batch-test/src/test/java/org/springframework/batch/test/SpringBatchTestJUnit5Tests.java
```diff
@@ -0,0 +1,146 @@
...
+import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseType;
+import org.springframework.test.context.ContextConfiguration;
+
+/**
+ * Test cases for usage of {@link SpringBatchTest} annotation with JUnit 5.
+ *
+ * @author Mahmoud Ben Hassine
+ */
+@SpringBatchTest
+@ContextConfiguration(classes = SpringBatchTestJUnit5Tests.JobConfiguration.class)
```
sbrannen, line 54: ditto

benas:
Thank you for your feedback! I pushed an update that addresses your suggestions. Let me know if anything else should be updated.

> As a sanity check, you might want to manually verify that JUnit Jupiter artifacts don't end up as mandatory dependencies in the generated POM.

I did a quick check, and the dependency is optional as expected, here is an extract from the generated pom:

```xml
    <dependency>
      <groupId>org.junit.jupiter</groupId>
      <artifactId>junit-jupiter-api</artifactId>
      <version>5.6.2</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
```

sbrannen:
> Thank you for your feedback! 

You're welcome.

> Let me know if anything else should be updated.

LGTM 👍

PR #3701, opened 2020-04-26
BATCH-2783: Change getUniqueJobParameters() to have a lower chance of repeating

saikatbhadra:
> ##  Background
> We noticed that integration tests for a batch job (with many tests) were failing intermittently. After debugging, I found that JobLauncherTestUtils.launchJob was occasionally relaunching the same job instance. After looking at the code, I saw that a random number generator was being used which does not guarantee that the number generate would be unique. I believe a UUID would be more appropriate as strongly guarantees that an ID does not repeat (except in highly exceptional cases). This change strongly assures that the JobParameter is unique every time getUniqueJobParameters() is called.
>

### DIFF
--- a/spring-batch-test/src/test/java/org/springframework/batch/test/JobLauncherTestUtilsTests.java
+++ b/spring-batch-test/src/test/java/org/springframework/batch/test/JobLauncherTestUtilsTests.java
@@ -17,7 +17,12 @@ package org.springframework.batch.test;
 
 import org.junit.Test;
 
-import org.springframework.batch.core.*;
+import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.Job;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.Step;
+import org.springframework.batch.core.StepContribution;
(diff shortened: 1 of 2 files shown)


saikatbhadra:
cc @mminella @benas

spring-batch-test/src/test/java/org/springframework/batch/test/JobLauncherTestUtilsTests.java
```diff
@@ -17,11 +17,7 @@
 
 import org.junit.Test;
 
-import org.springframework.batch.core.ExitStatus;
-import org.springframework.batch.core.Job;
-import org.springframework.batch.core.JobExecution;
-import org.springframework.batch.core.Step;
-import org.springframework.batch.core.StepContribution;
+import org.springframework.batch.core.*;
```
mminella, line 20: Can we restore the explicit imports?  Spring's coding standards prefers to avoid `.*` imports.

saikatbhadra: This is fixed

benas:
LGTM. Rebased, squashed and merged as 670b6408daf80e15d1311671252d728c70bdaa3f. Thank you for your contribution!

Issue #1777, opened 2011-11-18
Introduce BufferedWriterFactory abstraction [BATCH-1816]


spring-issuemaster:
**[Jimmy Praet](https://jira.spring.io/secure/ViewProfile.jspa?name=jpraet)** opened **[BATCH-1816](https://jira.spring.io/browse/BATCH-1816?redirect=false)** and commented

A BufferedWriterFactory abstraction could be introduced in the FlatFileItemWriter, which for example would make it possible to write directly to a gzipped output stream. This is consistent with the BufferedReaderFactory that already exists for FlatFileItemReader.

---
No further details from [BATCH-1816](https://jira.spring.io/browse/BATCH-1816?redirect=false)

parikshitdutta:
Hi @benas, please assign this to me.

Thank you.

parikshitdutta:
Hi @benas, please take a look at my PR #3724 as you find time. I will look forward to your feedback.

Thank you.

benas:
Closing this issue as explained in #3724 .

PR #3724, opened 2020-05-31
Added BufferedWriterFactory abstraction in AbstractFlatFileItemWriter

parikshitdutta:
- added factory interface with BufferedWriter and TransactionAwareBufferedWriter abstraction
- added default implementation for the introduced abstract factory
- added respective tests to test BufferedWriter and TransactionAwareBufferedWriter creation
- added setter in AbstractFileItemWriter to allow custom factory implementation
- updated AbstractFileItemWriter to create BufferedWriter and TransactionAwareBufferedWriter using newly introduced abstract factory

Closes BATCH-1816 [#1777]

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/BufferedWriterFactory.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/BufferedWriterFactory.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.batch.item.file;
+
+import java.io.BufferedWriter;
+import java.nio.channels.FileChannel;
+
+import org.springframework.batch.support.transaction.TransactionAwareBufferedWriter;
+
+/**
+ * An abstract factory strategy for custom extensions of both {@link BufferedWriter} and
+ * {@link TransactionAwareBufferedWriter} allowing customisation of the standard behaviour of the
+ * <code>java.io</code> variety.
+ *
+ * @author Parikshit Dutta
+ *
+ * @since 4.3
+ */
+public interface BufferedWriterFactory {
+	BufferedWriter createBufferedWriter(FileChannel fileChannel, String encoding, boolean forceSync);
+	TransactionAwareBufferedWriter createTransactionAwareBufferedWriter(FileChannel fileChannel, String encoding,
+			boolean forceSync, Runnable closeCallback);
+}
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/DefaultBufferedWriterFactory.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/DefaultBufferedWriterFactory.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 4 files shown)


benas:
Hi @parikshitdutta ,

Thank you for this PR. While the changes look good to me, there are a couple of issues with the suggested new interface:

```java
public interface BufferedWriterFactory {
 	BufferedWriter createBufferedWriter(FileChannel fileChannel, String encoding, boolean forceSync);
 	TransactionAwareBufferedWriter createTransactionAwareBufferedWriter(FileChannel fileChannel, String encoding,
 			boolean forceSync, Runnable closeCallback);
 }
```

* `TransactionAwareBufferedWriter` is a support class located in `org.springframework.batch.support.transaction` package. Support classes should not be "leaked" in public APIs like `BufferedWriterFactory` which is in `org.springframework.batch.item.file`, otherwise we would create an incorrect dependency between these two packages.
* Transaction related stuff (including the `forceSync` flag) is an implementation detail, it should not be part of the contract. As a user, if I want to provide a custom buffered writer through the factory without worrying about transactions, then I should not be "forced" to implement `createTransactionAwareBufferedWriter`.
* `FileChannel` is a low level construct for a high level contract like `BufferedWriterFactory` (one might ask why the interface accepts a `FileChannel` and not a `FileOutputStream` for example). We use a `FileChannel` internally to be able to truncate the file to the last committed offset in a restart scenario after a failure, but this is an implementation detail. Ideally, similar to the [BufferedReaderFactory](https://github.com/spring-projects/spring-batch/blob/d8fc58338d3b059b67b5f777adc132d2564d7402/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/BufferedReaderFactory.java#L46), we should accept a `Resource` (and even better, a `WritableResource` on the writer side)  + an encoding, something like:

```java
public interface BufferedWriterFactory {
	BufferedWriter create(WritableResource resource, String encoding) throws IOException;
}

public class DefaultBufferedWriterFactory implements BufferedWriterFactory {
	@Override
	public BufferedWriter create(WritableResource resource, String encoding) throws IOException {
		return new BufferedWriter(new OutputStreamWriter(resource.getOutputStream(), encoding));
	}
}
```

Do you agree?

Now the problem is that there are a few impediments with this approach as well:

1. `AbstractFileItemWriter` works with a [`Writer`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/AbstractFileItemWriter.java#L390) and not a `BufferedWriter`, so we can't use the factory in [getBufferedWriter()](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/AbstractFileItemWriter.java#L589). `AbstractFileItemWriter` used to work with a `BufferedWriter` but this has been changed in cfc9529babcf16e6cd56047b085bee67be736cca.
2. Trying to fix 1) requires changing `TransactionAwareBufferedWriter` to extend `BufferedWriter` and not `Writer` (IMO, this should have been the case from the beginning as this a decorator of a buffered writer). This in turn requires making `TransactionAwareBufferedWriter` accept the decorated writer at construction time (it used to be the case but it has been changed in 491359aacce4da643f8e41e3b8c83aa089e9c867). Moreover, [this cast](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/AbstractFileItemWriter.java#L428) prevents from working against a `BufferedWriter`.
3. Using a `WritableResource` instead of `Resource` in `AbstractFileItemWriter` (and in turn in `ResourceAwareItemWriterItemStream`) is a breaking change.

I'm not suggesting to implement these points, I'm just trying to measure the impact of introducing such feature in a clean way. So we need to find a middle ground. Wdyt?

Looking forward to your feedback.

parikshitdutta:
Hi @benas,
I was unaware about support package, that should not be exposed in public api, it's good that I know now.

However, I thought of keeping forceSync out of this, and I was sure FileChannel would be too low level for this high level abstraction.

But, as you pointed out in your proposed solution, TransactionAwareBufferedWriter having BufferedWriter as base would have helped to design the proposed interface using factory method than abstract factory. There I would have reason to design a common contract to serve both the writers.   

In addition, TransactionAwareBufferedWriter also needs FileChannel at construct time.

I was trying to touch as little as possible, but I am with you, I will look forward to any suggestion you come up with, and since I know, it's okay to dig little deeper for better solution. if I find something reasonable, I will share as well.

Thank you.

benas:
@parikshitdutta Thank you for your feedback. We discussed this issue with @mminella and we think the effort/impact needed to introduce this feature in a clean way (with all the required breaking changes) is too high compared to its added value, so we decided to close it.

Thank you for your PR anyway, much appreciated!

parikshitdutta:
Hi @benas,
Good to see you back! Totally make sense, we can rather spend time on something more productive. Thank you & @mminella for keeping me informed.

Thank you.

PR #3742, opened 2020-07-10
Add getChunkSize in SimpleCompletionPolicy

### LINKED ISSUE #3738: `SimpleCompletionPolicy` should have `getChunkSize()` method.
Currently, `SimpleCompletionPolicy` doesn't have `getChunkSize()`.
But `toString()` method returns a string that include own chunk size by some format.

You can see them on [this](https://github.com/spring-projects/spring-batch/blob/d8fc58338d3b059b67b5f777adc132d2564d7402/spring-batch-infrastructure/src/main/java/org/springframework/batch/repeat/policy/SimpleCompletionPolicy.java#L117).
```java
SimpleCompletionPolicy examplePolicy = new SimpleCompletionPolicy(50);
System.out.println(examplePolicy.toString());
// SimpleCompletionPolicy: chunkSize=50
```

I think it is a **code smell**.
If I want to get its chunk size, I have to parse `toString()` result. It is a very ugly way.

I suggest two options to solve this.
1. Make `toString()` doesn't show own chunk size.
2. **Make `getChunkSize()` public method.**

In my opinion, It should be second option.
If you approve this idea, comment, please. And then, I will make a pull request for this.

Thank you.

MinJunKweon:
- Add `getChunkSize()` in `SimpleCompletionPolicy`

resolved #3738

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/repeat/policy/SimpleCompletionPolicy.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/repeat/policy/SimpleCompletionPolicy.java
@@ -51,6 +51,10 @@ public class SimpleCompletionPolicy extends DefaultResultCompletionPolicy {
 		this.chunkSize = chunkSize;
 	}
 
+	public int getChunkSize() {
+		return chunkSize;
+	}
+
 	/**
 	 * Reset the counter.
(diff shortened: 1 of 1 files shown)


benas:
LGTM, rebased and merged. Thank you for your contribution!

Issue #3736, opened 2020-06-26
Resource should be optional in StaxEventItemReaderBuilder


benas:
Same as #3731 but for XML files.

parikshitdutta:
Hi @benas, would you like to assign this to me? I can give it a try.

Thank you.

parikshitdutta:
Hi @benas, Please have a look at my submitted PR #3752, I will look forward to your feedback.

Thank you.

PR #3752, opened 2020-07-25
Updated StaxEventItemReaderBuilder To Make Resource Optional

parikshitdutta:
- updated StaxEventItemReaderBuilder to build without Resource (optional now)
- added respective test to validate Resource can be null at build
- updated existing testValidation test case to pass the scenario

Fixes BATCH-3736 [#3736]


benas:
LGTM. Rebased and merged as c2625b1f870325ba3515eba469fd0a62d042427a. Thank you for your contribution!

Issue #758, opened 2019-11-06
Add setter for "standalone" attribute in StaxEventItemWriter [BATCH-2856]


spring-issuemaster:
**[Mahmoud Ben Hassine](https://jira.spring.io/secure/ViewProfile.jspa?name=mbenhassine)** opened **[BATCH-2856](https://jira.spring.io/browse/BATCH-2856?redirect=false)** and commented

As of v4.2.0, the only way to add the `standalone` attribute in the XML declaration of the output file of `StaxEventItemWriter` is to extend the writer and override `startDocument` (as stated in its [javadoc](https://docs.spring.io/spring-batch/docs/4.2.x/api/org/springframework/batch/item/xml/StaxEventItemWriter.html#startDocument-javax.xml.stream.XMLEventWriter-)). Here is an example:

```java
@Bean
public StaxEventItemWriter<Person> itemWriter() {
    StaxEventItemWriter<Person> itemWriter = new StaxEventItemWriter<Person>() {
        @Override
        protected void startDocument(XMLEventWriter writer) throws XMLStreamException {
            writer.add(createXmlEventFactory().createStartDocument(DEFAULT_ENCODING, DEFAULT_XML_VERSION, true));
            writer.add(xmlEventFactory.createStartElement(getRootTagNamespacePrefix(), getRootTagNamespace(), getRootTagName()));
        }
    };
    // TODO set other properties on the writer
    return itemWriter;
}
```

The `StaxEventItemWriter` provides setters for the `version` and `encoding` attributes, but not for the `standalone` attribute. I believe it can also provide a setter for `standalone` as this will make it easier to customize the XML declaration rather than having to extend the writer and override `startDocument` (which is error prone BTW if we forget to add the root element as shown in the example above).

NB: The example shown above works with Java 9+, see https://bugs.openjdk.java.net/browse/JDK-8139584 

---

**Affects:** 4.2.0

**Reference URL:** https://stackoverflow.com/questions/58714744/staxeventitemwriter-xml-header

parikshitdutta:
Hi @mminella, just wanted to check with you, if you are active on this one, or I can give it a try.

Thank you.

mminella:
Hi @parikshitdutta I assigned this to you as your requested. Let me know if you don't have time to work on it.

parikshitdutta:
Hi @mminella,
I was thinking PR#3752 to pass through so that there is no effort in conflict resolution, Please suggest.

Thank you.

[...]

PR #3764, opened 2020-08-20
Added support for 'standalone' attribute in StaxEventItemWriter

parikshitdutta:
- updated woodstox version from '6.2.0' to '6.2.1' as mentioned: https://github.com/spring-projects/spring-batch/issues/758#issuecomment-677382726
- added setter for standalone attribute in StaxEventItemWriter
- added relevant tests in StaxEventItemWriterTests
- added builder method for standalone attribute in StaxEventItemWriterBuilder
- added relevant tests in StaxEventItemWriterBuilderTests

Closes BATCH-2856 [#758 ]

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/xml/StaxEventItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/xml/StaxEventItemWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2017 the original author or authors.
+ * Copyright 2006-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -72,7 +72,7 @@ import org.springframework.util.StringUtils;
  * @author Peter Zozom
  * @author Robert Kasanicky
  * @author Michael Minella
- * 
+ * @author Parikshit Dutta
  */
 public class StaxEventItemWriter<T> extends AbstractItemStreamItemWriter<T> implements
 ResourceAwareItemWriterItemStream<T>, InitializingBean {
@@ -84,6 +84,9 @@ ResourceAwareItemWriterItemStream<T>, InitializingBean {
 
 	// default encoding
 	public static final String DEFAULT_XML_VERSION = "1.0";
+
+	// default standalone document declaration, value not set
+	public static final Boolean DEFAULT_STANDALONE_DOCUMENT = null;
 
 	// default root tag name
 	public static final String DEFAULT_ROOT_TAG_NAME = "root";
@@ -108,6 +111,9 @@ ResourceAwareItemWriterItemStream<T>, InitializingBean {
 
 	// XML version
 	private String version = DEFAULT_XML_VERSION;
+
+	// standalone header attribute
+	private Boolean standalone = DEFAULT_STANDALONE_DOCUMENT;
 
 	// name of the root tag
 	private String rootTagName = DEFAULT_ROOT_TAG_NAME;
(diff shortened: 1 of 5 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/xml/StaxEventItemWriter.java
```diff
@@ -606,7 +630,12 @@ protected void startDocument(XMLEventWriter writer) throws XMLStreamException {
 		XMLEventFactory factory = createXmlEventFactory();
 
 		// write start document
-		writer.add(factory.createStartDocument(getEncoding(), getVersion()));
+		if (getStandalone()) {
+			writer.add(factory.createStartDocument(getEncoding(), getVersion(), getStandalone()));
+		}
+		else {
+			writer.add(factory.createStartDocument(getEncoding(), getVersion()));
+		}
```
benas, line 638: This means we will have either `standalone='yes'` or no standalone attribute at all. Is there any reason for that choice? There is no way to output `standalone='no'` with this approach. I know that the absence of the standalone attribute is equivalent to `standalone='no'` which is the default, but there is a section from the [w3c spec](https://www.w3.org/TR/xml/#sec-rmd) that concerns me:

```
The standalone document declaration must have the value "no" if any external markup declarations contain declarations of:

...
``` 

So unless I'm misunderstanding this spec, there are some circumstances where the standalone attribute must have the value `no` (Now I'm not sure if this means the attribute must be present and set to `no` or completely absent which is also interpreted as `no`..). Since we are not sure how XML parsers would behave regrading to this situation, we prefer to give the user the ability to write the standalone attribute with either `yes` or `no`, or omit it completely. This can be done using an enum (for example with YES, NO, NONE (the default) values) instead of a boolean. 

@parikshitdutta Wdyt? If you agree, please update the PR accordingly. Otherwise let me know and I can take care of the changes.

parikshitdutta:
Hi @benas,
Thanks for reviewing the PR.

I knew ```no``` is the default value, and it is better to omit any attribute value that is optional and default (I had this knowledge), So:

1. Standalone attribute is optional, and default is ```no```

2. As I looked into ```stax2-api-4.2.1``` code, I found they too only writing the attribute if it's ```yes``` (true), please check here: https://github.com/FasterXML/stax2-api/blob/8f10bb50124cb70048a3ec8dd2dca4421a04733c/src/main/java/org/codehaus/stax2/ri/Stax2EventWriterImpl.java#L86-L87

That is why, I went with that notion, and didn't see any reason to offer different options.

But then I also understand, giving ```yes/no/none``` would make it less ambiguous, more flexible, and that then does not depend on any particular parser or underlying implementation. But I am not very sure if that would be redundant or not.

I am okay to make the necessary changes, please share your final thought.

Thank you.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/xml/StaxEventItemWriter.java
parikshitdutta: I knew ```no``` is the default value, and it is better to omit any ```attribute value``` that is ```optional``` and ```default``` (I had this knowledge), So:

1. Standalone attribute is optional, and default is no

2. As I looked into stax2-api-4.2.1 code, I found they too only writing the attribute if it's yes (true), please check here: https://github.com/FasterXML/stax2-api/blob/8f10bb50124cb70048a3ec8dd2dca4421a04733c/src/main/java/org/codehaus/stax2/ri/Stax2EventWriterImpl.java#L86-L87

That's why, I went with that notion, and didn't see any reason to offer different options.

But then I also understand, giving ```yes/no/none``` would make it less ambiguous, more flexible, and that then does not depend on any particular parser or underlying implementation. But I am not very sure if that would be redundant or not.

I am okay to make the necessary changes, please share your final thought.

benas: Thank you for your feedback!

> But then I also understand, giving yes/no/none would make it less ambiguous, more flexible, and that then does not depend on any particular parser or underlying implementation.

I agree, I think we need to give the user the ability to explicitly output `standalone='no'` even if omitting it is equivalent (this might be a requirement for some users and it is currently impossible with this PR). However, in hindsight, we do not need an enum, we can use the wrapper `Boolean` type (and not the primitive boolean) and use `null` as default:

* `null` : no standalone attribute at all (the default)
* Boolean.FALSE: standalone='no'
* Boolean.TRUE: standalone='yes'

If you agree, please update the PR accordingly and it should be good to merge. Thank you upfront!

parikshitdutta:
Hi @benas, 
Please review the PR now, I have made changes based on our discussion.

Thank you.

benas:
@parikshitdutta LGTM now, rebased and merged as d65203e2d3532aaae8ef1b2ac3a80a5149613f3e.

Thank you for your contribution 👍

Issue #901, opened 2018-03-27
JpaCursorItemReader [BATCH-2707]


spring-issuemaster:
**[Michael Minella](https://jira.spring.io/secure/ViewProfile.jspa?name=mminella)** opened **[BATCH-2707](https://jira.spring.io/browse/BATCH-2707?redirect=false)** and commented

JPA 2.2 added the ability to stream results as a cursor instead of just paging. This story is to explore the practicality of using that API to provide something similar to the functionality provided via the `HibernateCursorItemReader` but using the JPA APIs.

---

**Affects:** 4.0.1

benas:
Resolved with #3775 .

PR #3775, opened 2020-09-15
Add JpaCursorItemReader implementation

benas:
This PR is for Issue #901. It adds a cursor-based implementation of an `ItemStreamReader` based on `Query#getResultStream()` which was added in JPA 2.2.

The new reader passes the same test suite as other cursor-based implementations.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaCursorItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaCursorItemReader.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JpaCursorItemReaderBuilder.java
```diff
@@ -0,0 +1,182 @@
...
+	 */
+	public JpaCursorItemReaderBuilder<T> queryString(String queryString) {
+		this.queryString = queryString;
+
+		return this;
+	}
+
+	/**
+	 * The {@link EntityManagerFactory} to be used for executing the configured
+	 * {@link #queryString}.
```
mminella, line 145: We use this with the `QueryProvider` approach as well.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaCursorItemReader.java
```diff
@@ -0,0 +1,153 @@
...
+	private String queryString;
+	private JpaQueryProvider queryProvider;
+	private Map<String, Object> parameterValues;
+	private Iterator<T> iterator;
+
+	/**
+	 * Create a new {@link JpaCursorItemReader}.
+	 */
+	public JpaCursorItemReader() {
+		setName(ClassUtils.getShortName(JpaCursorItemReader.class));
```
mminella, line 62: Why do this if you are going to require a name to be set in the builder?

benas:
Merged in 41854008e8c1f22cd2f7d4f95326b0a3f159b58b.

Issue #3765, opened 2020-08-24
Jackson2ExecutionContextStringSerializer: customization


fprochazka:
I would like to be able to extend the `TRUSTED_CLASS_NAMES` without having to copy&paste the entire `Jackson2ExecutionContextStringSerializer`.

I can see that you had to fix a security vuln in #3732, but it broke (de)serialization in a lot of apps. I have no problem adding the annotations to my classes, but I have no idea how to allow `java.util.UUID`. 

I suggest that you introduce a mechanism, that would allow me to extend the list of trusted classes in case there is a JDK/library class I cannot modify.

----

Also it's really hard to override the serializer and I had to extend a bunch of configuration and bean factory classes to accomplish it.

mminella:
Mixins are the expected way to handle classes you do not have the ability to modify with Jackson. Is there a reason you cannot use one for UUID? That being said, if we can confirm UUID is not a "gadget class", it also feels like one that would be reasonable to just add to our list.

fprochazka:
@mminella Can you please show me how to configure the `Jackson2ExecutionContextStringSerializer` to allow UUID? How do I tell the `ObjectMapper` instance inside the class about the mixin?

fprochazka:
@mminella Can you please show me how to configure the `Jackson2ExecutionContextStringSerializer` to allow UUID? How do I tell the `ObjectMapper` instance inside the class about the mixin? 

I do not believe it's currently possible. I could configure my own ObjectMapper and add the mixin there, but then I lose the default, because the `createTrustedDefaultTyping()` is private and I see no way to hook into that without reflection magic.

snussbaumer:
I had the same issue than @fprochazka : maybe we're both missing something ... but as already stated if we create our own ObjectMapper we loose the TrustedTypeIdResolver entirely ...

Also I'm using spring-boot and used to having everything working out of the box with sensible defaults but still being configurable ... now to accept UUID (and java.sql.Timestamp in my case) I have to define a BatchConfigurer, configure a JobLauncher, JobRepository, JobExplorer (hoping I have everything setup correctly) ... 

A property with a list of additional trusted types would be really great !

benas:
[...]

PR #3787, opened 2020-10-06
Add ability to amend trusted classes in Jackson2ExecutionContextStringSerializer

benas:
This PR is for #3765. It adds the ability to specify additional trusted classes (at construction time of the `Jackson2ExecutionContextStringSerializer`) without having to provide a custom object mapper.

The usage of varargs in the constructor of `Jackson2ExecutionContextStringSerializer` makes the change backward compatible with 4.2.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/Jackson2ExecutionContextStringSerializer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/Jackson2ExecutionContextStringSerializer.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


fprochazka:
This looks like it solves it for me. It's still not very easy to replace the service itself (`Jackson2ExecutionContextStringSerializer`), but now at least I don't have to copy&paste the entire class. Thanks!

nstdio:
@benas Seems this change is not binary compatible with client compile against `4.2.x`.

benas:
@nstdio This enhancement/feature has not been backported to 4.2.x, even though my initial comment implies it. We typically backport bug fixes only to patch releases of previous supported lines (4.2.x is approaching its EOL anyway). Anything preventing you from upgrading to 4.3?

nstdio:
@benas Thanks for your response. Great work! We already upgraded to `4.3` with minimal effort.

The issue was that internal library was compiled against `4.2.x`. After upgrading to Spring Batch version to `4.3` in the consumer project, we've got `java.lang.NoSuchMethodError`. The runtime cannot find the `Jackson2ExecutionContextStringSerializer.<init>()` method, because in `4.3` constructor signature was changed.

benas:
@dthng You need to provide a custom `BatchConfigurer` and override `getJobRepository()`, see [Configuring a JobRepository](https://docs.spring.io/spring-batch/docs/4.3.x/reference/html/job.html#configuringJobRepository). You can set your custom serializer on the [factory bean](https://docs.spring.io/spring-batch/docs/4.3.x/api/org/springframework/batch/core/repository/support/JobRepositoryFactoryBean.html#setSerializer-org.springframework.batch.core.repository.ExecutionContextSerializer-) used to create the job repository (same approach for the job explorer if needed).

PR #3791, opened 2020-10-12
Poll the count of running step executions

### LINKED ISSUE #3790: High memory consumption during long running jobs
**Bug description**
We found that memory consumption is fairly high on one of the service nodes that uses the Spring Batch. Even though both data nodes did a similar amount of work, the memory consumption across nodes was not even - 15GB vs 1.5GB (see memory use screenshot).

We have some jobs that could run for seconds while others might run for hours, so we set the polling interval (MessageChannelPartitionHandler#setPollInterval) to 1 second rather than 10 seconds that is the default. In a large running job scenario, we ended up creating 837 step executions.

What I found was that MessageChannelPartitionHandler#pollReplies gets a full StepExecution representation for each step, which contains a JobExecution which would also contain StepExecutions for each. However, they are retrieved at different times and stages. This means that we end up with square number of StepExecution objects, e.g. 837*837=700569 StepExecutions (see screenshot below)

**Environment**
Initially reproduced on Spring Batch 4.1.4. 

**Expected behavior**
My proposal would be to:
1. Issue a SQL query to get the count of running StepExecutions instead of retrieving DTOs. This way there is less objects loaded into the heap.
2. Once all steps are finished, then query for all StepExecutions for that job. We can then assign the same JobExecution to each step.

Memory usage graph comparison between two service nodes, doing roughly equal amount of work:

memoryUse - redacted

My apologies for a messy screenshot, but it does explain the number of StepExecution objects:

stepExecutions - redacted

PauliusPeciura:
In a combination of short poll interval and large number of step executions that take a long time to run, memory consumption can go high - each step execution has its own reference to a job execution, which refers step executions for the job. This means that the job executions are different instances, resulting in creating a lot of objects.
Instead, we query the database to get the number of step executions that are still running. Once all of them are finished, we then would fetch all steps and assign the same job execution that can be a shared instance.

Closes #3790

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/BatchStatus.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/BatchStatus.java
@@ -16,6 +16,9 @@
 
 package org.springframework.batch.core;
 
+import java.util.Arrays;
+import java.util.List;
+
 /**
  * Enumeration representing the status of an Execution.
  * 
@@ -39,6 +42,8 @@ public enum BatchStatus {
 	 */
 	COMPLETED, STARTING, STARTED, STOPPING, STOPPED, FAILED, ABANDONED, UNKNOWN;
 
+	public static final List<BatchStatus> RUNNING_STATUSES = Arrays.asList(STARTING, STARTED);
+
 	public static BatchStatus max(BatchStatus status1, BatchStatus status2) {
 		return status1.isGreaterThan(status2) ? status1 : status2;
 	}
@@ -49,7 +54,7 @@ public enum BatchStatus {
 	 * @return true if the status is STARTING, STARTED
 	 */
 	public boolean isRunning() {
-		return this == STARTING || this == STARTED;
+		return RUNNING_STATUSES.contains(this);
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


thesoundofsilent:
Is there any plan to fix this issue?

vitenkov (approved):

fmbenhassine:
Thank you for this PR! I believe this change set goes in the right direction as mentioned in https://github.com/spring-projects/spring-batch/issues/3790#issuecomment-1439615319, which is similar to your proposal in the initial issue description. 

However, since 4.3 is the last minor version of v4, we cannot introduce such a change set in a patch version of 4.3.x. The earliest version to introduce such changes is 5.1.0. Moreover, there are other code changes (not related to this PR) that need to be introduced in order to fix the issue in a clean way. 

So I will to postpone this PR to 5.1.0 and let you know when it is a good time to rebase it on the main branch. Is that ok for you? Otherwise please let me know and I will take care of all the changes.

In the meantime, the performance issue will be addressed with #4208.

fmbenhassine:
> So I will to postpone this PR to 5.1.0 and let you know when it is a good time to rebase it on the main branch.

Could you please rebase this PR on the latest `main`? Thank you upfront.

fmbenhassine:
No rebase on the latest main as requested, so I am closing this PR. Thank you anyway.

PR #3798, opened 2020-11-03
Do not throw NPE for job instances without executions

hpoettker:
When working with Spring Cloud Dataflow, it does occasionally happen that the database contains batch job instances without any batch job executions. When launching an execution for such an instance, this currently leads to a `NullPointerException`.

See for example:
- https://github.com/spring-cloud/spring-cloud-dataflow/issues/4233
- https://github.com/spring-cloud/spring-cloud-dataflow/issues/4142

Spring Batch should handle such a situation more gracefully and throw an exception with a hint to the problem.


benas:
> Feature: Do not throw NPE for job instances without executions

If Spring Batch fails with a NPE, I consider this as a bug and not a feature. I edited the title accordingly.

benas:
LGTM, rebased and merged as b6c3fb6d8ce6167e80ab10109a698b7767c97cd4. Thank you for your contribution!

PR #3800, opened 2020-11-08
Don't call wrapper constructors directly

marschall:
The constructors of the wrapper classes for primitives (Double,
Integer, Long) have been deprecated since JDK 9 and should no longer be
called. Instead the static factory #valueOf should be used.


marschall:
The same tests fail as currently fail on the master.

benas:
Rebased and merged as c9e06b2a7f62d03b7ce33836cad9aebcd5660178. Thank you for your contribution.

PR #3804, opened 2020-11-15
Replace Assert.assertThat with MatcherAssert.assertThat

marschall:
Assert.assertThat is deprecated with MatcherAssert.assertThat as the
replacement.


benas:
Rebased and merged as 1f3dd5f40190d89fb0244d3c8f285c383de62129 (which partially resolves #3838). Thank you for your contribution!

PR #3805, opened 2020-11-15
Replace #initMocks with MockitoRule

marschall:
Replace all references of MockitoAnnotations#initMocks with the
MockitoRule JUnit rule.

MockitoAnnotations#initMocks is deprecated with MockitoAnnotations#openMocks was a replacement. However #openMocks requires closing in an @After method. Using MockitoRule is easier as it automatically closes the mocks after the test.


benas:
Rebased and merged as 2aa94d09d12bfa6dd1a9df606e7ef5f986fc10b2 which partially resolves #3838 . Thank you for your contribution!

PR #3812, opened 2020-12-02
Reduce the toCharArray overhead

mdeinum:
Both the PatternMatcher and JdbcParameterUtils use String.toCharArray for
processing. This will create a new char[] which needs to be collected. The
same code has been rewritten to use String.charAt to reduce the overhead.

Additionally unneeded object creation in the PatternMatcher has been removed
as well, to reduce further objects needed to be gc'ed.


benas:
LGTM. Rebased and merged as 11bac7c4e1903553a1e3b7411efc56dadd59234a. Thank you for your contribution.

PR #3813, opened 2020-12-03
Better sizes for StringBuilder

mdeinum:
StringBuilder by default has a size of 16, however there
are some places that already create larger strings already.
For efficiency and garbage reduction it would be better to have
larger sizes to begin with.


benas:
LGTM. Rebased and merged as e9bdcc6d66fa7b3d91d0817501e3885e91e4e55e. Thank you for your contribution!

PR #3814, opened 2020-12-03
Added isXXXEnabled for logging statements

mdeinum:
Guarded the logging statements which do concatenation of strings
or calling toString on objects. When a log level isn't enabled
this still would produce garbage that would need to be collected.

Guarded all logging up to info, warn and error can be assumed to be
enabled on a production system.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/configuration/xml/JsrBeanDefinitionDocumentReader.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/configuration/xml/JsrBeanDefinitionDocumentReader.java
@@ -182,7 +182,7 @@ public class JsrBeanDefinitionDocumentReader extends DefaultBeanDefinitionDocume
 	}
 
 	private String resolveValue(String value) {
-		StringBuffer valueBuffer = new StringBuffer();
+		StringBuilder valueBuffer = new StringBuilder();
 		Matcher jobParameterMatcher = OPERATOR_PATTERN.matcher(value);
 
 		while (jobParameterMatcher.find()) {
@@ -200,7 +200,7 @@ public class JsrBeanDefinitionDocumentReader extends DefaultBeanDefinitionDocume
 
 				String resolvedProperty = properties.getProperty(extractedProperty, NULL);
 
-				if (NULL.equals(resolvedProperty)) {
+				if (NULL.equals(resolvedProperty) && LOG.isInfoEnabled()) {
 					LOG.info(propertyType + " with key of: " + extractedProperty + " could not be resolved. Possible configuration error?");
 				}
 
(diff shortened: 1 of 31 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/AbstractNeo4jItemReader.java
```diff
@@ -184,9 +184,7 @@ protected String generateLimitCypherQuery() {
 
 		String resultingQuery = query.toString();
 
-		if (logger.isDebugEnabled()) {
-			logger.debug(resultingQuery);
-		}
+		logger.debug(resultingQuery);
```
benas, line 187: This seems an unintentional change (as it goes against the idea of the PR). I will revert it on merge.

spring-batch-samples/src/main/java/org/springframework/batch/sample/common/InfiniteLoopWriter.java
```diff
@@ -61,6 +61,8 @@ public void write(List<? extends Object> items) throws Exception {
 		}
 
 		stepExecution.setWriteCount(++count);
-		LOG.info("Executing infinite loop, at count=" + count);
+		if (logger.isInfoEnabled()) {
```
benas, line 64: This line should be `if (LOG.isInfoEnabled()) {`. I will fix it on merge.

spring-batch-samples/src/main/java/org/springframework/batch/sample/common/LogAdvice.java
```diff
@@ -28,7 +28,9 @@
 public class LogAdvice {
     private static Log log = LogFactory.getLog(LogAdvice.class);
 
-    public void doStronglyTypedLogging(Object item){
-    	log.info("Processed: " + item);
+    public void doStronglyTypedLogging(Object item) {
+        if (logger.isInfoEnabled()) {
```
benas, line 32: This line should be `if (log.isInfoEnabled()) {`. I will fix it on merge.

spring-batch-core/src/main/java/org/springframework/batch/core/jsr/configuration/xml/JsrBeanDefinitionDocumentReader.java
```diff
@@ -182,7 +182,7 @@ private String resolvePropertyValue(String propertyValue) {
 	}
 
 	private String resolveValue(String value) {
-		StringBuffer valueBuffer = new StringBuffer();
+		StringBuilder valueBuffer = new StringBuilder();
```
benas, line 185: This change creates a compilation error in `jobParameterMatcher.appendReplacement` and `jobParameterMatcher.appendTail`. Those methods accept a `StringBuilder` starting from Java 9 onward. I will revert that change on merge. Please make sure to use Java 8 when building your PR.

benas:
Rebased and merged as b7d144c10d43e18b06f7921d751f97e21ab4a9a7. Thank you for your contribution!

PR #3819, opened 2020-12-14
Provide constructors for injection

mdeinum:
Currently when reading/writing JSON and using the reader/marshaller provided
and you want a custom Gson or ObjectMapper instance it still creates the not
needed instance. Move the construction to a constructor and provide a constructor
to directly pass in the pre-configured Gson or ObjectMapper instance.

The same approach is used in Spring itself where Gson or ObjectMapper instances
can be passed in.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/json/GsonJsonObjectMarshaller.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/json/GsonJsonObjectMarshaller.java
@@ -30,9 +30,18 @@ public class GsonJsonObjectMarshaller<T> implements JsonObjectMarshaller<T> {
 
 	private Gson gson = new Gson();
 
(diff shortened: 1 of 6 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/json/GsonJsonObjectMarshaller.java
```diff
@@ -30,9 +30,18 @@
 
 	private Gson gson = new Gson();
```
benas, line 31: According to the improvement suggested in this PR, there shouldn't be an inline instantiation of `Gson` here (similar to what has been done in other classes in this change set). I will update that on merge.

benas:
Rebased and merged as af14a2e4718f7c51411df19a562738a4d136f200. Thank you for your contribution!

PR #3824, opened 2021-01-01
Switch mongoItemReader to use a cursor instead of paging

jbotuck:
Paging in mongo runs in On^2 time since each skip needs to count all the records it is skipping


benas:
Thank you for this PR! We can't change the `MongoItemReader` in such a way: you have not only changed the behaviour of the reader, but also the signature of several public APIs (some of them have even been removed in this PR). This would be a big breaking change and we can't introduce it as such.

I suggest to create a new cursor-based mongo item reader named `MongoCursorItemReader` next to the current one, similar to what we already have for JDBC (`JdbcCursorItemReader` / `JdbcPagingItemReader`), JPA (`JpaCursorItemReader`, `JpaPagingItemReader`), etc.

The current `MongoItemReader` could be deprecated and extended with a class named `MongoPagingItemReader` to have a consistent naming.

Please update the PR accordingly and we can continue from there.

fmbenhassine:
We have a PR for the cursor-based implementation here: #4323 , which seems similar to what we have here. Please do not hesitate to review it if you want.

As mentioned previously, the current `MongoItemReader` should be renamed to `MongoPagingItemReader` to have a consistent naming with other readers. I will take care of that in #4341, and I will close this PR in favor of #4323.

Thank you for your contribution anyway.

jbotuck:
Makes sense to me. Reworking this pr has been on the bottom of my to do list for a while.

Issue #3773, opened 2020-09-11
KafkaItemWriter should wait for results of kafkaTemplate.sendDefault call


jbotuck:
I noticed the kafkaItemWriter calls sendDefault (an asynchronous call) without checking the future returned by the call to confirm the data was actually sent. This seems bad to me. I would think it should call flush on the template after the batch is written and then check all the futures to confirm nothing went wrong.

PR #3827, opened 2021-01-05
KafkaItemWriter.write should not return until items are confirmed to have been written

jbotuck:
This PR is for issue #3773.


benas:
LGTM. Rebased and merged as c886a60b4d39e11ca2d558dab0902cd4fbceaba6. Refined in 572a30272560e1c4464631b6f726510a80bc568d. I also added a timeout to prevent the job from running indefinitely if `future.get` never completes (see a15496805b4a2d054620d642aa3314cb4bc5c73c), similar to what we have in [MessageChannelPartitionHandler](https://github.com/spring-projects/spring-batch/blob/4b9554fcce80e38f7adeffc82a26436199c28191/spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/MessageChannelPartitionHandler.java#L275-L280). Thank you for your contribution!

PR #3840, opened 2021-01-28
Override SimpleStepBuilder.faultTolerant() in FaultTolerantStepBuilder

marbon87:
```
		@Bean
		public Step faultTolerantStep(StepBuilderFactory stepBuilderFactory) throws Exception {
			return stepBuilderFactory.get("faultTolerantStep")
					.chunk(2)
					.reader(reader())
					.processor(processor())
					.writer(writer())
					.faultTolerant()
					.listener(chunkListener)
					.skipLimit(1)
					.faultTolerant()
                                        .build()
```

I had a failure in a step configuration when calling .faultTolerant() twice. That led to the problem that the chunkListener was not registered. 
The reason is that calling .faultTolerant() on an existing FaultTolerantStepBuilder creates a new FaultTolerantStepBuilder but not all configurations, like ChunkListeners, are kept.
Although it was a mistake in my job config, i think it would be better to return the current FaultTolerantStepBuilder to prevent such mistakes.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/FaultTolerantStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/FaultTolerantStepBuilder.java
@@ -432,6 +432,14 @@ public class FaultTolerantStepBuilder<I, O> extends SimpleStepBuilder<I, O> {
 		}
 		return this;
 	}
+	
+	/**
+	 * Override parent method to prevent creation of a new FaultTolerantStepBuilder
+	 */
+	@Override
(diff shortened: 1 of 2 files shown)


benas:
Good catch! Thank you for your PR. This will make the `FaultTolerantStepBuilder` more tolerant to configuration faults 😉 

However, could you please add a failing test that passes with this PR? We need to get this covered by a test. Thank you upfront.

marbon87:
Hi @benas, thanks for the feedback. I added a test.

marbon87:
Hey @benas , can you merge this pr please?

benas:
@marbon87 Thank you for updating the PR by adding a test. I planned to merge the fix in the upcoming 4.3.3.

Issue #3650, opened 2020-01-27
StepExecution counts integer overflow


ArtyomGabeev:
Hi,

In StepExecution class counts like readCount/writeCount, etc has int type.
In case when step operates on a big amount of items, there may be an int overflow and may affect e.g. progress monitoring if someone use these counters.

Should we change it to long instead?

ArtyomGabeev:
Also noted, apply method which increments counts is synchronized. Should we switch from volatile to atomic long?

benas:
Similar/Related to #1055.

ArtyomGabeev:
@benas any update on this? we are getting integer overflow :)

adityausathe:
@benas, If this is up for grabs, I would like to take a shot at it. It would be helpful if you could provide some pointers.

benas:
@ArtyomGabeev Thank you for reporting this issue. The problem can also happen in a locally or remotely partitioned/chunked step: each worker could contribute a read/write count and the aggregation might silently fail with an integer overflow on the manager side, leading to incorrect counters in the aggregate `StepExecution`. An example can be found [here](https://github.com/benas/spring-batch-lab/blob/main/issues/gh3650/src/main/java/io/spring/PartitionJobSample.java).

Changing the type of metrics from `int` to `long` seems reasonable to me for a major version. Moreover, columns in the `BATCH_STEP_EXECUTION` table are already designed to accept `long` values.

PR #3845, opened 2021-02-04
Change metrics type from int to long in StepExecution

benas:
This draft PR is for #3650 . This work, which is still in progress, has been done to evaluate the impact of the change requested in #3650: change type of metrics from `int` to `long` in `StepExecution` . As shown here, many related public APIs are impacted like `StepContribution`, `SkipPolicy`, etc and we need to make sure all changes are still conform to JSR-352 (fortunately, the return type of `javax.batch.runtime.Metric#getValue` is `long`).


benas:
Updated and merged as 327160a3ab02a4d94e1d951f64e73bcb5b5cfc24.

Issue #3794, opened 2020-10-19
"IllegalArgumentException: Only 1 out of 2 arguments could be assigned" using MethodInvokingTaskletAdapter + NamedParameterJdbcTemplate


alessandrococco:
I'm using Spring Boot 2.3.4 in a kotlin 1.4.10 project.

**Bug description**

I have the following `MethodInvokingTaskletAdapter`:

```
    @Bean
    @StepScope
    fun myTasklet(): Tasklet {
        val tasklet = MethodInvokingTaskletAdapter()

        // jdbcTemplate.update("UPDATE ...",  MapSqlParameterSource("value", value))

        tasklet.setTargetObject(jdbcTemplate) // NamedParameterJdbcTemplate
        tasklet.setTargetMethod("update")
        tasklet.setArguments(
            arrayOf(
                "UPDATE ...",
                MapSqlParameterSource("value", value)
            )
        )
        return tasklet
    }
```

When the tasklet starts I get this exception:

```

java.lang.IllegalArgumentException: Only 1 out of 2 arguments could be assigned.
	at org.springframework.batch.item.adapter.HippyMethodInvoker.findMatchingMethod(HippyMethodInvoker.java:84) ~[spring-batch-infrastructure-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.util.MethodInvoker.prepare(MethodInvoker.java:188) ~[spring-core-5.2.9.RELEASE.jar:5.2.9.RELEASE]
	at org.springframework.batch.item.adapter.AbstractMethodInvokingDelegator.doInvoke(AbstractMethodInvokingDelegator.java:112) ~[spring-batch-infrastructure-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.batch.item.adapter.AbstractMethodInvokingDelegator.invokeDelegateMethod(AbstractMethodInvokingDelegator.java:62) ~[spring-batch-infrastructure-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at org.springframework.batch.core.step.tasklet.MethodInvokingTaskletAdapter.execute(MethodInvokingTaskletAdapter.java:53) ~[spring-batch-core-4.2.4.RELEASE.jar:4.2.4.RELEASE]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_211]
[...]

PR #3853, opened 2021-02-12
Fix HippyMethodInvoker by not considering unsuitable candidates

hpoettker:
Fixes https://github.com/spring-projects/spring-batch/issues/3794

The `HippyMethodInvoker` currently calculates the `typeDifferenceWeight` based not on all required arguments but only those that are actually assignable. If a candidate method has one argument that's an exact match and another that does not match at all, this logic yields a perfect `typeDifferenceWeight` of 0 which prevents actually suitable candidates from being considered.

The idea is to not consider candidate methods with unassignable arguments since these lead to an `IllegalArgumentException` if selected anyways.


benas:
LGTM. Rebased and merged as 9f4127fb1d1499f1d7d7a78c7dbaf04204e8d405. Thank you for your contribution.

Issue #3852, opened 2021-02-11
Intermittent failure in AsynchronousTests on windows


benas:
`org.springframework.batch.repeat.jms.AsynchronousTests` [fails intermittently on windows](https://build.spring.io/download/BATCH-BATCHWIN-JOB1/build_logs/BATCH-BATCHWIN-JOB1-299.log) with the following error:

```
error	11-Feb-2021 08:30:35	[main] INFO org.apache.activemq.broker.BrokerService - Apache ActiveMQ 5.15.13 (broker, ID:WIN-OUDKQ7UFSBD-53071-1613032189853-0:1) is shutdown
build	11-Feb-2021 08:30:35	[ERROR] Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 2.468 s <<< FAILURE! - in org.springframework.batch.repeat.jms.AsynchronousTests
build	11-Feb-2021 08:30:35	[ERROR] testSunnyDay(org.springframework.batch.repeat.jms.AsynchronousTests)  Time elapsed: 1.223 s  <<< ERROR!
build	11-Feb-2021 08:30:35	org.springframework.dao.DataAccessResourceFailureException: StatementCallback; SQL [delete from T_BARS]; connection exception: closed; nested exception is java.sql.SQLNonTransientConnectionException: connection exception: closed
build	11-Feb-2021 08:30:35		at org.springframework.batch.repeat.jms.AsynchronousTests.onSetUp(AsynchronousTests.java:78)
build	11-Feb-2021 08:30:35	Caused by: java.sql.SQLNonTransientConnectionException: connection exception: closed
build	11-Feb-2021 08:30:35		at org.springframework.batch.repeat.jms.AsynchronousTests.onSetUp(AsynchronousTests.java:78)
build	11-Feb-2021 08:30:35	Caused by: org.hsqldb.HsqlException: connection exception: closed
build	11-Feb-2021 08:30:35		at org.springframework.batch.repeat.jms.AsynchronousTests.onSetUp(AsynchronousTests.java:78)
build	11-Feb-2021 08:30:35	
build	11-Feb-2021 08:30:35	[ERROR] testRollback(org.springframework.batch.repeat.jms.AsynchronousTests)  Time elapsed: 1.23 s  <<< ERROR!
[...]

Issue #3851, opened 2021-02-11
Intermittent failure in ConcurrentTransactionTests on windows


benas:
`org.springframework.batch.core.test.concurrent.ConcurrentTransactionTests.testConcurrentLongRunningJobExecutions` fails intermittently on windows from the command line with the following error:

```
error	11-Feb-2021 08:38:26	[main] INFO org.springframework.batch.core.job.SimpleStepHandler - Executing step: [gameLoad]
build	11-Feb-2021 08:38:26	[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.078 s <<< FAILURE! - in org.springframework.batch.core.test.concurrent.ConcurrentTransactionTests
build	11-Feb-2021 08:38:26	[ERROR] testConcurrentLongRunningJobExecutions(org.springframework.batch.core.test.concurrent.ConcurrentTransactionTests)  Time elapsed: 0.063 s  <<< FAILURE!
build	11-Feb-2021 08:38:26	java.lang.AssertionError: expected:<FAILED> but was:<COMPLETED>
build	11-Feb-2021 08:38:26		at org.springframework.batch.core.test.concurrent.ConcurrentTransactionTests.testConcurrentLongRunningJobExecutions(ConcurrentTransactionTests.java:83)
build	11-Feb-2021 08:38:26	
build	11-Feb-2021 08:38:26	[INFO] Running org.springframework.batch.core.test.football.FootballJobIntegrationTests
```

nfjcore:
This test also fails for me when running all tests in spring-batch-core.  The test passes in isolation. I'm on MacOS. Let me know if I can provide additional information.

benas:
@nfjcore I'm also on MacOS (Catalina v10.15.7) and this test didn't fail up until now when I build locally with Java 8:

```
$ java -version
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

It did not fail neither on our [CI server](https://build.spring.io/browse/BATCH-MCI-17). But if you say it fails on your machine, I trust you. 

> Let me know if I can provide additional information.

Yes, please. Which java version do you use? Do you use the maven wrapper or your local maven installation?

nfjcore:
I am using the following java version:

```
java -version
openjdk version "1.8.0_282"
[...]

PR #3856, opened 2021-02-20
Reduce naming collisions in consecutive tests

hpoettker:
fixes https://github.com/spring-projects/spring-batch/issues/3851
fixes https://github.com/spring-projects/spring-batch/issues/3852

The root cause is that there are too many embedded databases with the same name. The issue is described here: https://github.com/spring-projects/spring-framework/issues/13491

I'm using Windows 10 with Oracle JDK 14.0.2 and it works on my machine: 😉 
- the build passes, when started form the command line,
- the tests pass when started per module in Eclipse, which triggers the test classes in a fixed but non-alphabetical order

Simply removing the ignore annotations on the master branch led to consistently reproducible errors.


benas:
Rebased and merged as 0ad68037b662ab01701a5b041eedac4b3130ee27. Thank you for your contribution!

Issue #3859, opened 2021-02-27
AvroItemWriter uses GenericDatumWriter instead of SpecificDatumWriter for type that extends SpecificRecordBase


adrianbielewicz:
**Bug description**
I am generating Java code from an Avro schema. It creates classes that extend SpecificRecordBase. In this case AvroItemWriter automatically chooses GenericDatumWriter as the underlying writer. This causes multiple exceptions when logical types are used in the schema (for example decimal or timestamp-millis).

It would work perfectly if SpecificDatumWriter is used. I believe that the problem lies in the order in which datum writer is chosen. The code checks first if my item class is an instance of GenericRecord and only after that returns false, it check if it is an instance of SpecificRecordBase. As SpecificRecordBase implements GenericRecord the first check will always return true and this leads to GenericDatumWriter to be used.

**Environment**
Spring Batch 4.3.1

**Steps to reproduce**
1. Create a simple schema that uses logical types like decimal and timestamp-millis.
2. Generate java code using avro-tools.
3. Try to write using the generated classes.

**Expected behavior**
SpecificDatumWriter should be used instead of GenericDatumWriter.

PR #3860, opened 2021-02-27
BATCH-3859 Changing the order in which datum writer is selected (due …

adrianbielewicz:
It fixes the problem described in #3859


benas:
Thank you for your PR. The fix makes sense to me. Indeed the Datum selection should start from the most specific one to the generic one and finally fallback to the reflection based one.

Rebased and merged as f4e14ddf66f8a2d43dd24ac0419c641a31bddbe7. Thank you for raising the issue and contributing a fix!

PR #3861, opened 2021-03-01
Remove whitespace to support multiple execution statement

caodangtinh:
For now if we try multiple execution with drop statement, we got error like
`SQL Error [1064] [42000]: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DROP TABLE IF EXISTS BATCH_JOB_EXECUTION_CONTEXT ;
DROP TABLE IF EXISTS BATCH_ST' at line 2`
This merge request remove all redundant white-space at the end of each drop command to allow multiple execution at one time.


benas:
Rebased and merged as 324a769aa09aa3e3a0ca82bec15f25f791dbc089. Thank you for your contribution.

Issue #834, opened 2018-11-13
org.springframework.batch.core.JobParameter.toString may return null [BATCH-2772]


spring-projects-issues:
**[Dmitry Smeliansky](https://jira.spring.io/secure/ViewProfile.jspa?name=dmitrysm2000)** opened **[BATCH-2772](https://jira.spring.io/browse/BATCH-2772?redirect=false)** and commented

```java
public String toString() { public String toString() { return parameter == null ? null : (parameterType == ParameterType.DATE ? "" + ((Date) parameter).getTime() : parameter.toString()); }
```

This behavior causes NPE in the JopParameters.toProperties method

```java
java.lang.NullPointerException at java.util.Hashtable.put(Hashtable.java:459) at org.springframework.batch.core.JobParameters.toProperties(JobParameters.java:219) 
```

---

**Affects:** 3.0.7

benas:
Fixed in #3864 .

PR #3864, opened 2021-03-11
Fix bug in JobParameters.toProperties on null value

acktsap:
fixes https://github.com/spring-projects/spring-batch/issues/834

Put "" to Properties when null parameter value. Also add some test for it.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,8 @@ import java.io.Serializable;
 import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
```diff
@@ -269,8 +271,8 @@ public Properties toProperties() {
 		Properties props = new Properties();
 
 		for (Map.Entry<String, JobParameter> param : parameters.entrySet()) {
-			if(param.getValue() != null) {
-				props.put(param.getKey(), param.getValue().toString());
+			if (param.getValue() != null) {
+				props.put(param.getKey(), Optional.of(param.getValue()).map(Object::toString).orElse(""));
```
PascalSchumacher, line 275: I think this could be simplified to `Objects.toString(param.getValue(), "")`.

acktsap: Thanks for your comment. I've changed to `Objects.toString(param.getValue().toString(), ""))`.

benas:
Hi @acktsap ,

Thank you for this PR. However, this actually fixes the symptom of the problem and not its root cause. The real issue is that the value of the parameter must not be `null` in the first place (as well as the result of its `toString` implementation), as explained in #3913 . Since the correct fix is a breaking change and has been scheduled for v5, I will merge this as a temporary fix for the v4.3 version since it is still maintained (and will be maintained for some time).

Note that I would have fixed the issue by making `JobParameter#toString` return an empty String instead of `null` [here](https://github.com/spring-projects/spring-batch/blob/d8fc58338d3b059b67b5f777adc132d2564d7402/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameter.java#L175) (a `toString` method is not expected to return `null`, otherwise it not a good idea), but this breaks other tests (namely [this one](https://github.com/spring-projects/spring-batch/blob/d8fc58338d3b059b67b5f777adc132d2564d7402/spring-batch-core/src/test/java/org/springframework/batch/core/JobParametersBuilderTests.java#L137)) where it is expected that `toString` returns `null`.

Rebased and merged as 4af4b05e338d295d7cb060491ae3a832d0b05eb1. Thank you for your contribution.

PR #3867, opened 2021-03-13
Use batch inserts for job parameters

marschall:
Use batch inserts for job parameters to reduce the number of database
round trips.


fmbenhassine:
Rebased and merged as c707284f80cb93aacf28c9612378f9eec97fd0e1. Thank you for your contribution!

PR #3868, opened 2021-03-13
Remove double brace initialization

marschall:
Remove double brace collection initialization as it is considered bad
practice [1] [2].

> However, it is quite obscure, and it costs an extra class at each usage. It also holds hidden references to the enclosing instance and to any captured objects. This may cause memory leaks or problems with serialization. For these reasons, this technique is best avoided.

> I think the example of "double-brace initialization" should be even more clearly labeled as pure evil. :-) You could also mention all the horrible consequences if anyone ever serializes such a collection.

 [1] https://mail.openjdk.java.net/pipermail/core-libs-dev/2015-September/035441.html
 [2] https://openjdk.java.net/jeps/269


benas:
Rebased and merged as 3d0cb905903ca81af88f1fd89c8a9ae6fbb0d90b. Thank you for your contribution.

Issue #1577, opened 2013-04-25
In JOB_PARAMS table DATE_VAL column is updated incorrectly Like "1/1/1970 1:00:00.000000 AM" instead to the current date. [BATCH-2015]


spring-projects-issues:
**[Sonali Singhal](https://jira.spring.io/secure/ViewProfile.jspa?name=sonalisinghal)** opened **[BATCH-2015](https://jira.spring.io/browse/BATCH-2015?redirect=false)** and commented

---

**Affects:** 2.1.8

0 votes, 5 watchers

benas:
I agree with Michael, `null` is a better default value than the suggested current date or the start or epoch date (as implemented currently). I just checked all DDLs and they all declare the `DATE_VAL` column to be nullable by default. So changing `new Timestamp(0L)` with `null` in [`JdbcJobExecutionDao#insertParameter`](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobExecutionDao.java#L346-L354) should be safe.

parikshitdutta:
Hi @benas, Please have a look at my submitted PR #3869, I will look forward to your feedback.

Thank you.

PR #3869, opened 2021-03-21
Change date param default to null for non-date type job params

parikshitdutta:
- updated JdbcJobExecutionDao to change date default to null from epoch
- added respective test to validate date param is null at rest for non-date type job parameters

Fixes BATCH-2015 [#1577]


benas:
Rebased and merged as 637226064ce84d4a37cf499ac11c8e42ff08ce2b. Thank you for your contribution!

PR #3889, opened 2021-04-18
Incorrect code examples in documentation

LichKing-lee:
This example code fails the build. `end()` method is not exists


benas:
Rebased and merged as b486e245d7b23b90eee3aadfbbf05ec8031ceb80. There are other occurrences of the same error which I fixed and amended in your commit. Thank you for your contribution!

PR #3896, opened 2021-04-23
updated verifyCursorPosition default to true

### LINKED ISSUE #3893: JdbcCursorItemReaderBuilder defaults verifyCursorPosition to false, but its Javadoc says the opposite
**Bug description**
`JdbcCursorItemReaderBuilder` defaults `verifyCursorPosition` to false, despite its Javadoc claiming the opposite.  This is also opposite the behavior you get if you build a `JdbcCursorItemReader` directly.  The javadoc currently reads

	/**
	 * Indicates if the reader should verify the current position of the
	 * {@link java.sql.ResultSet} after being passed to the {@link RowMapper}.  Defaults
	 * to true.
	 *
	 * @param verifyCursorPosition indicator
	 * @return this instance for method chaining
	 * @see JdbcCursorItemReader#setVerifyCursorPosition(boolean)
	 */
	 

**Environment**
Spring Batch 4.2.4.  Appears to be the case in newer versions as well.

**Expected behavior**
I would expect `JdbcCursorItemReaderBuilder` to build a reader with `verifyCursorPosition = true` if not explicitly set, since this matches the default behavior of `JdbcCursorItemReader`.

Or, at minimum, the javadoc on `JdbcCursorItemReaderBuilder#verifyCursorPosition` should be updated.

```@Test
public void testVerifyCursorPosition() {

    JdbcCursorItemReader<Object> reader = new JdbcCursorItemReaderBuilder<>()
            .name("name")
            .sql("select 1")
            .dataSource(new SimpleDriverDataSource())
            .rowMapper((rs, i) -> null)
            //.verifyCursorPosition(true) <-- the default according to the javadoc
            .build();
    assertTrue((boolean) ReflectionTestUtils.getField(reader, "verifyCursorPosition")); // fails
}

parikshitdutta:
Fixes #3893

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcCursorItemReaderBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/builder/JdbcCursorItemReaderBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


benas:
LGTM. Rebased and merged as d76c083b046e8979c38e47d4f6401d5977b64f76. Thank you @parikshitdutta !

PR #3902, opened 2021-05-03
Fix misused 'end()' in examples

acktsap:
`use()` can be used in `FlowJobBuilder` usage. Similar with https://github.com/spring-projects/spring-batch/pull/3889.

### DIFF
--- a/spring-batch-docs/src/main/asciidoc/domain.adoc
+++ b/spring-batch-docs/src/main/asciidoc/domain.adoc
@@ -77,7 +77,6 @@ public Job footballJob() {
                      .start(playerLoad())
                      .next(gameLoad())
                      .next(playerSummarization())
-                     .end()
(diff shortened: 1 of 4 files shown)


spring-batch-docs/src/main/asciidoc/domain.adoc
```diff
@@ -77,7 +77,6 @@ public Job footballJob() {
                      .start(playerLoad())
                      .next(gameLoad())
                      .next(playerSummarization())
-                     .end()
```
acktsap, line 80: This can be used if `playerLoad()` is flow. But it seems that `playerLoad()` is step [according to context](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-docs/src/main/asciidoc/step.adoc#step-restart-configuration-example).

Issue #2218, opened 2009-08-10
Deprecate throttle limit in favour of using similar features in TaskExecutor implementations


spring-projects-issues:
**[Dave Syer](https://jira.spring.io/secure/ViewProfile.jspa?name=david_syer)** opened **[BATCH-1364](https://jira.spring.io/browse/BATCH-1364?redirect=false)** and commented

---
No further details from [BATCH-1364](https://jira.spring.io/browse/BATCH-1364?redirect=false)

hpoettker:
I'm not sure that it's a good idea to deprecate the throttle limit. As explained in https://github.com/spring-projects/spring-batch/issues/3857#issuecomment-827901009 , the thread pool must be big enough to cover the waiting of `SplitState` nodes if there are nested splits in the job.

As far as I know, the throttle limit is the only effective way to set a maximal number of concurrent steps if a job has nested splits.

Issue #1008, opened 2017-05-08
Spring Batch and Reactive Streams [BATCH-2596]


spring-projects-issues:
**[Christian Trutz](https://jira.spring.io/secure/ViewProfile.jspa?name=christian.trutz@gmail.com)** opened **[BATCH-2596](https://jira.spring.io/browse/BATCH-2596?redirect=false)** and commented

Hello Spring Batch team,

are there any plans to incorporate Reactive Streams (https://projectreactor.io/) in Spring Batch? I think about concepts like backpressure (between ItemReaders and ItemWriters within a Spring Batch step). I talked today with Mark Paluch `@JAX` and he suggested to open an discussion issue.

Christian

---

1 votes, 4 watchers

sayjavajava:
@spring-issuemaster any update on this thread ?

philipbel:
Any updates? It's been 3+ years since the original request?

philipbel:
> @philipbel We're really interested in working with someone with a concrete use case for this. If you have one you can share, let us know!

@mminella,
I am working on a server where upon user signup, a job is triggered to fetch some resources from 3rd-party ReST APIs and populate a MongoDB collection for the user.
A second use case is to periodically (once every few hours) run a batch job, again fetching some external resources, and doing a computation on them and store the result in MongoDB and in a 3rd-party ReST API.
A third use case I have is running machine learning jobs (tied to AWS) once a day, which, too, store their results in MongoDB.

For all jobs, I would like to provide an overview to the user of what the job's status is, preferably with a progress.
In terms of data, not much, < 10 MB of JSON per job per user. Each job has a few steps.

The application uses WebFlux + WebClient and MongoDB Reactive.

mminella:
@philipbel So what are you looking to get out of a Reactive Streams integration with Spring Batch?

omar-napoleon:
Hello. Is there a way to integrate rsocket (reactor) into an ItemProcessor? I have the following problem. I have to call a server with tcp but when sending the data I need the response to process the data and be able to send it to the itemwriter, but waiting for the response delays all my processing. I need the communication to be asynchronous with the server but not to queue the next iterations of the itemprocessor. I have already converted the itemProcessor to asynchronous but in the same way when I run out of threads due to the delay of the external server my processing is delayed. I must meet a minimum amount of TPS to send to the server.

benas:
> Spring Batch and Reactive Streams

[...]

PR #3903, opened 2021-05-03
Explicit handling of nested splits in SplitState

### LINKED ISSUE #3857: SplitBuilder.add(Flow) causes hung execution in some cases
**Bug description**
If `FlowBuild.SplitBuilder.add(Flow)` is called multiple times to dynamically build a list of Flows, the result is a Flow that will hang.
Calling `FlowBuild.SplitBuilder.add(Flow[])` with an array of Flows works as expected.

For instance, this does not error, but does not work as expected:
```
FlowBuilder<SimpleFlow> myFlow = new FlowBuilder<SimpleFlow>("DoesntWork");
FlowBuilder.SplitBuilder<SimpleFlow> split = myFlow.split(exe);

for (int i = 0; i < 10; i++) {
	split.add(createANewFlow(i)); // <<< Key change - call add() multiple times
}
```

However, this does work:
```
FlowBuilder<SimpleFlow> myFlow = new FlowBuilder<SimpleFlow>("Works");

List<Flow> flows = new ArrayList<>();
for (int i = 0; i < 10; i++) {
	flows.add(createANewFlow(i));
}

myFlow.split(exe).add(flows.toArray(new Flow[flows.size()]));   //Add flows all in one 'add' operation
```

**Why would you call `FlowBuild.SplitBuilder.add(Flow)` multiple times??**
I need to dynamically build my list of tasks - I don't know the size of the list or what the tasks are ahead of time.  I do know, however, that they can all be run in parallel and there are a lot of them (hundreds).  Calling `FlowBuild.SplitBuilder.add(Flow)` for each as I build up the list makes the most intuitive sense.

**Environment**
Note:  I'm pretty sure this is related to how SpringBatch works, not any sort of JDK issue.
SpringBatchCore 4.3.1
OpenJDK 12
Mac OS 10.14

**Steps to reproduce**
See the code example in GitHub, below.

The first flows runs as expected, the second flow hangs after two steps are executed.

**Expected behavior**
Calling `FlowBuild.SplitBuilder.add(Flow)` should just add to the list of flows to run in parallel.  It shouldn't matter if they are added all at once or add is called multiple times.

I created a Maven based demo project that shows this bug:
https://github.com/ee-usgs/springbatch_parallel_bug

This application should be runnable via Maven or any IDE.

hpoettker:
Fixes #3857 and allows to control the maximal number of concurrently running steps with the thread pool size of the `TaskExecutor`.

Nested splits are explicitly handled in `SplitState` and do not require dedicated threads. The added test stalls if executed against main.

@benas Do you see a chance to merge this PR (with requested changes, of course) such that it is included in Spring Batch 5.0?

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -300,9 +300,9 @@ public class FlowBuilder<Q> {
 		return result;
 	}
 
-	private SplitState createState(Collection<Flow> flows, TaskExecutor executor) {
+	private SplitState createState(Collection<Flow> flows, TaskExecutor executor, SplitState parentSplit) {
 		if (!states.containsKey(flows)) {
-			states.put(flows, new SplitState(flows, prefix + "split" + (splitCounter++)));
+			states.put(flows, new SplitState(flows, prefix + "split" + (splitCounter++), parentSplit));
 		}
 		SplitState result = (SplitState) states.get(flows);
 		if (executor != null) {
@@ -606,7 +606,7 @@ public class FlowBuilder<Q> {
 
 		private final FlowBuilder<Q> parent;
 
-		private TaskExecutor executor;
+		private final TaskExecutor executor;
 
 		/**
 		 * @param parent the parent builder
@@ -626,23 +626,23 @@ public class FlowBuilder<Q> {
 		public FlowBuilder<Q> add(Flow... flows) {
 			Collection<Flow> list = new ArrayList<>(Arrays.asList(flows));
 			String name = "split" + (parent.splitCounter++);
-			int counter = 0;
 			State one = parent.currentState;
-			Flow flow = null;
+
+			if (one instanceof SplitState) {
+				parent.currentState = parent.createState(list, executor, (SplitState) one);
+				return parent;
+			}
+
 			if (!(one == null || one instanceof FlowState)) {
-				FlowBuilder<Flow> stateBuilder = new FlowBuilder<>(name + "_" + (counter++));
+				FlowBuilder<Flow> stateBuilder = new FlowBuilder<>(name + "_0");
 				stateBuilder.currentState = one;
-				flow = stateBuilder.build();
+				list.add(stateBuilder.build());
 			}
 			else if (one instanceof FlowState && parent.states.size() == 1) {
 				list.add(((FlowState) one).getFlows().iterator().next());
 			}
 
-			if (flow != null) {
-				list.add(flow);
-			}
-			State next = parent.createState(list, executor);
-			parent.currentState = next;
+			parent.currentState = parent.createState(list, executor, null);
 			return parent;
 		}
 
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/state/SplitState.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/state/SplitState.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2013 the original author or authors.
+ * Copyright 2006-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


benas:
Hi @hpoettker, thank you for your PR!

> @benas Do you see a chance to merge this PR (with requested changes, of course) such that it is included in Spring Batch 5.0?

I did not check this PR and its corresponding bug report in details yet, but it seems that it could introduce a small breaking change. So yes, this could be addressed in v5. However, I'm not sure it will be in v5.0 as we are planning to focus on major changes first. This is a good candidate for a minor release of the v5 line (v5.1, v5.2, etc). Please follow the updates on this issue to see in which release it will be addressed. Thank you.

hpoettker:
Thanks for the update! I'd also consider the PR a breaking change in behavior although I assume that no one knowingly relies on the current behavior as it is rather unexpected.

Just for completeness: The fix would also solve a problem in the Data Flow's Composed Task Runner which actively uses nested splits and the thread pool size of a `ThreadPoolTaskExecutor` to control concurreny: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4279

hpoettker:
I've resolved some merge conflicts and bumped the year in the copyright header.

It would be great to have open discussion about the PR and the issue that it addresses. I think it's closely linked to the deprecation of the throttle limit, i.e. #2218. Concurrency of batch steps should be fully controllable by using only a suitable `TaskExecutor`. This is currently not possible if nested splits are involved as explained in the conversation of #3857.

Input from the Spring Cloud Dataflow team would be particularly interesting as the Composed Task Runner would benefit greatly from this change, in my opinion.

hpoettker:
This PR is now more than 2 years old but I'm convinced more than ever that it addresses an important issue. I think that the concept of the chunk-oriented step in Spring Batch is extremely valuable. But it can only reach its full potential if it allows to scale batch jobs in a straightforward manner by "just" adding a thread pool. The removal of the throttle limit is the most important part of this story. But concurrency control for split jobs follows right after that in my opinion.

> This is a good candidate for a minor release of the v5 line (v5.1, v5.2, etc).

As we're already closer in time to the release of 5.1 than that of 5.0, I'd say that this is now!

Please share your ideas on this topic. I'd be happy to implement any change or test that is required.

fmbenhassine:
Yes, this PR can now be merged. I will plan it for 5.1.

> Please share your ideas on this topic

Intra-step concurrency or step-level concurrency (ie processing items concurrently within the same step) is different than inter-step / job-level concurrency (ie processing entire steps concurrently or in parallel within a flow). Those two types of concurrency are at different levels and should not be related in any way.

This PR addresses inter-step concurrency, and I don't think the deprecation of throttle-limit is related to the handling of nested splits. The only case I see a possible entanglement is probably when the _same_ task executor is used to run both step flows and concurrent threads of the same step. The throttle-limit is really an implementation detail of the `TaskExecutorRepeatTemplate`. If I provide a custom `RepeatTemplate` that does **not** use a throttle-limit, I am not sure I would solve #3857 anyway.

I correctly read your comment that `The removal of the throttle limit is the most important part of this story. But concurrency control for split jobs follows right after that in my opinion` and I agree. I just want to make clear that those are two different, unrelated issues. 

This PR addresses #3857 and will be merged in 5.1. However, the deprecation/removal of throttle-limit will be reviewed as part of #3950.

fmbenhassine:
Rebased and merged as 58a4134c328c56e68c6e0c25936c20a0665233ef. Thanks again for yet another great contribution!

hpoettker:
@fmbenhassine Thanks for sharing your insight and clearly delineating the distinction between job-level and step-level concurrency. It's really appreciated!

The throttle limit is indeed not directly related to this job-level concurreny issue. They are only connected in the broader sense that both are about controlling the maximum degree of concurrency in a job/step.

From a user perspective, I think it is nice if I can reliably control concurrency by passing task executors with the desired maximal thread pool size. For the job level, this didn't work before this change as jobs experienced liveness failure when thread pools were too small or executed a far smaller number of steps in parallel than the maximal thread pool size. The ideal behavior is of course that there are no liveness failures in any case and that as many steps are executed in parallel as there are threads assigned to this purpose (and where the flow of the job allows it).

The same should be true for the step-level analogously. Here the problem is currently more that it's not enough to provide a large thread pool if I want a high degree of concurrency because I still have to set the throttle limit or a custom `RepeatTemplate`. But this is of course addressed by the deprecation of the throttle limit.

The plans for the step level in #3950 are much more ambitious but also very welcome. Is there anything the community could do to assist?

hpoettker:
I'm wondering whether my reasoning about concurrency control above through thread pools still makes sense in the age of virtual threads. It may very well not! The threads for the job level don't do much work at all and the threads for for the step level are usually much more IO bound than CPU bound.

Thinking along this line could lead to the conclusion that the throttle limit should **not** be removed. It should rather be introduced for the job level too. Then I could use `newVirtualThreadPerTaskExecutor` for both levels but still have caps to not overload any systems.

But despite the lively discussion in #1008, I'm not sure that the overhead of platform threads is real life problem for batch work loads. 😄

PR #3906, opened 2021-05-13
Added support for quotes in for the DelimitedLineAggregator.

### LINKED ISSUE #1139: DelimitedLineAggregator support string quote characters [BATCH-2463]
**[Doug Breaux](https://jira.spring.io/secure/ViewProfile.jspa?name=breaux)** opened **[BATCH-2463](https://jira.spring.io/browse/BATCH-2463?redirect=false)** and commented

I can't imagine why org.springframework.batch.item.file.transform.DelimitedLineAggregator doesn't already support a quote character like the delimited tokenizer does. I can see that the StringUtils class being used doesn't provide this capability either, so I'm guessing it just wasn't trivially easy to add, but it seems like a necessary capability for working with delimited files.

I know I need it and am going to have to implement it some other way instead. (As far as I can tell, I can't simply extend an existing Batch class to add a quote delimiter around an individual field value as needed.)

---

2 votes, 9 watchers

cppwfs:
resolves #1139

We may need to discuss escaping the quotes embedded in the elements.
I chose the triple quote method to handle resolve this.   An example would be fo"o would be replaced with "fo"""o"

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineAggregator.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineAggregator.java
@@ -15,18 +15,27 @@
  */
 package org.springframework.batch.item.file.transform;
 
+import java.util.StringJoiner;
+
+import org.springframework.lang.Nullable;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 /**
  * A {@link LineAggregator} implementation that converts an object into a
- * delimited list of strings. The default delimiter is a comma.
+ * delimited list of strings. The default delimiter is a comma.  An optional
+ * quote value can be set to add surrounding quotes for each of elements of the list.
+ * Default is null, which means not quotes.
  * 
  * @author Dave Syer
+ * @author Glenn Renfro
  * 
  */
 public class DelimitedLineAggregator<T> extends ExtractorLineAggregator<T> {
 
 	private String delimiter = ",";
+
+	private String quote;
 
 	/**
 	 * Public setter for the delimiter.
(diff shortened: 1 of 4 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DelimitedLineAggregator.java
```diff
@@ -38,7 +47,52 @@ public void setDelimiter(String delimiter) {
...
+		if (ObjectUtils.isEmpty(arr)) {
+			return "";
+		} else if (arr.length == 1) {
+			return quote(handleEmbeddedQuote(ObjectUtils.nullSafeToString(arr[0])));
+		} else {
+			StringJoiner sj = new StringJoiner(delim);
+			Object[] var3 = arr;
+			int var4 = arr.length;
+
+			for(int var5 = 0; var5 < var4; ++var5) {
```
giuliopulina, line 63: variable naming could be improved :) is this decompiled code from StringUtils?

fmbenhassine:
Thank you for your PR! I added a couple inline notes. Could you please check? If you agree on the changes, I can take care of them on merge.

My plan is to include this change in the upcoming `5.1.0-M3`.

cppwfs:
Cool on the merge!   I didn't see the changes though.

fmbenhassine:
I did not introduce any change or merge the PR yet. I was waiting for your feedback if you agree on the suggestions: https://github.com/spring-projects/spring-batch/pull/3906#issuecomment-1720953191

If you agree, I can take care of the changes when merging the PR.

cppwfs:
Sounds good to me!    Thank you!

fmbenhassine:
Rebased and merged as cc233e6b350ab8894c669bbf01dec39e135e09a2. Thank you for your contribution!

PR #3908, opened 2021-05-16
Remove Reflection from StepScopeTestExecutionListener

marschall:
`StepScopeTestExecutionListener` uses reflection to call methods on `TestContext`. This seems to have been introduced to support Spring 3 and 4 at the same time. This is no longer required as we currently require Spring 5.

This can be seen as `JobScopeTestExecutionListener` does not use reflection and calls methods on `TestContext` directly.


benas:
Rebased and merged as 6d5a4d5352fcc3694f6ddd6025c277629f32c057. Thank you for your PR!

PR #3909, opened 2021-05-16
Use default methods in TestExecutionListener

marschall:
`TestExecutionListener` has default methods since Spring 5 (SPR-14432). As we require Spring 5 we can delete the empty implementation methods.


benas:
Rebased and merged as 1480dc5fd22bc0e6ff29fad1fbbdf8fd41f1e541. Thank you for your contribution.

PR #3910, opened 2021-05-17
FlatFileItemWriter now uses charset to determine default encoding

### LINKED ISSUE #1154: FlatFileItemReader and FlatFileItemWriter don't have the same default encoding [BATCH-2448]
**[Ant Kutschera](https://jira.spring.io/secure/ViewProfile.jspa?name=ant@maxant.co.uk)** opened **[BATCH-2448](https://jira.spring.io/browse/BATCH-2448?redirect=false)** and commented

We have a batch component which writes to an interjob file and another which reads from that file. They write and read using FlatFileItemWriter and FlatFileItemReader respectively.

Unfortunately the data which is read is no longer correct (our umlaut characters are corrupt), but the data in the interjob file is correct.  This happens because the default encoding for FlatFileItemReader is not the same as the default encoding for FlatFileItemWriter. Looking at the source, they are:

org.springframework.batch.item.file.FlatFileItemReader#encoding = #DEFAULT_CHARSET = Charset.defaultCharset().name()

org.springframework.batch.item.file.FlatFileItemWriter#encoding = #OutputState.DEFAULT_CHARSET = "UTF-8"

I.e. the writer is hard coded to UTF-8 but the reader is environment specific.

Ideally the reader and the writer would use the same default regardless of the environment.

The work around is to set the encoding specifically to UTF-8 on the reader.

Affects:
\<groupId>org.springframework.batch\</groupId>
\<artifactId>spring-batch-infrastructure\</artifactId>
\<version>3.0.4.RELEASE\</version>

---

**Affects:** 3.0.4

cppwfs:
resolves #1154

It now matches the encoding scheme of FlatFileItemReader

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.batch.item.file;
 
+import java.nio.charset.Charset;
 import java.util.List;
 
 import org.springframework.batch.item.file.transform.LineAggregator;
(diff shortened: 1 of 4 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/FlatFileItemWriter.java
```diff
@@ -39,11 +40,15 @@
...
  * @author Michael Minella
  * @author Mahmoud Ben Hassine
+ * @author Glenn Renfro
  */
 public class FlatFileItemWriter<T> extends AbstractFileItemWriter<T> {
 
 	protected LineAggregator<T> lineAggregator;
 
+	// default encoding for writing to flat files - set to charset of this Java virtual machine.
+	public static final String DEFAULT_CHARSET = Charset.defaultCharset().name();
```
benas, line 50: There is already a DEFAULT_CHARSET constant in [AbstractFileItemWriter](https://github.com/spring-projects/spring-batch/blob/d8fc58338d3b059b67b5f777adc132d2564d7402/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/AbstractFileItemWriter.java#L75). It is that value that should be changed instead of adding a new constant here.

cppwfs:
I will update this in the PR.  The issue mentions `FlatfileItemWriter` specifically  and  Ididn't want to affect other classes that inherit from `AbstractFileItemWriter ` for example `JDBCFileItemWriter`.

benas:
Rebased and merged as 6c619b9f9465709359af515e62ba0da39aa498be. Thank you @cppwfs !

PR #3912, opened 2021-05-18
Added  protected access methods for getTargetMethod and getTargetObject

### LINKED ISSUE #1786: Expose targetObject and targetMethod in AbstractMethodInvokingDelegator
**[Kyrill Alyoshin](https://jira.spring.io/secure/ViewProfile.jspa?name=kyrill007)** opened **[BATCH-1803](https://jira.spring.io/browse/BATCH-1803?redirect=false)** and commented

I have a child class of org.springframework.batch.item.adapter.AbstractMethodInvokingDelegator, and I'd love to log what method of which service bean I am invoking on a particular object. I think it would be appropriate to expose those properties with 'protected' access, so we don't have to resort to reflection.

---
No further details from [BATCH-1803](https://jira.spring.io/browse/BATCH-1803?redirect=false)

cppwfs:
Resolves #1786


fmbenhassine:
LGTM. Rebased and merged as 058dd3c1b06ee338b951c60f6b8982fa1f5863b1. Thank you!

Issue #3913, opened 2021-05-18
JobParameter must not accept null values


benas:
As of v4.3, it is possible to create a `JobParameter` with a `null` value. Unless I'm missing something, I see no valid use case for that. Moreover, this might cause issues like #834 and makes the code defensive for no added value. Like other domain objects where passed parameters are validated at creation time, job parameter values should be validated at construction time to be non null.

This is breaking change, so I will schedule it for v5.

GirirajSinghRathore:
I just upgraded to SpringBatch 5 and I got error like

"value must not be null"

it does not tell which field is set to null or the stack trace,

could we not add checks in .toParamerters constructor and throw some meaningful message

GirirajSinghRathore:
just upgraded to SpringBatch 5 and I got error like

"value must not be null"

it does not tell which field is set to null or the stack trace,

could we not add checks in .toParamerters constructor and throw some meaningful message

fmbenhassine:
@GirirajSinghRathore Can you please open a separate issue with an example? I agree the error message should at least mention the parameter name

GirirajSinghRathore:
Created the issue
@fmbenhassine 
[issue](https://github.com/spring-projects/spring-batch/issues/4581)

PR #3914, opened 2021-05-18
Add assertions to reject null values in JobParameter

benas:
This is a WIP. Tests should be updated accordingly.

Issue #3913


benas:
Updated and merged as 497438a318299cebe2500e6be2616c14fdfdcc88.

PR #3915, opened 2021-05-18
Add URL to error message when StaxEventItemReader does not exist

### LINKED ISSUE #1171: Provide better error message in StaxEventItemReader if Resource does not exist
**[Gunnar Hillert](https://jira.spring.io/secure/ViewProfile.jspa?name=hillert)** opened **[BATCH-2431](https://jira.spring.io/browse/BATCH-2431?redirect=false)** and commented

Depending on the type of resource we should provide the name/URI as part of the exception description. A message like

```
Caused by: java.lang.IllegalStateException: Input resource must exist (reader is in 'strict' mode)
```

does not seem helpful.

---

**Affects:** 3.0.4

cppwfs:
resolves #1171

### DIFF
--- a/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/xml/StaxEventItemReaderTests.java
+++ b/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/xml/StaxEventItemReaderTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008-2020 the original author or authors.
+ * Copyright 2008-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/test/java/org/springframework/batch/item/xml/StaxEventItemReaderTests.java
```diff
@@ -110,6 +113,41 @@ public void testAfterPropertiesSet() throws Exception {
 		source.afterPropertiesSet();
 	}
 
+	/**
+	 * Regular usage scenario. ItemReader should pass XML fragments to unmarshaller wrapped with StartDocument and
+	 * EndDocument events.
+	 */
+	@Test
+	public void testNoResource() throws Exception {
```
fmbenhassine, line 121: There is already a test named `testStrictness` that uses a `NonExistentResource` for the use case related to the original issue. So I think there is no need to add this new test and its related `getBadResource()` method. We can update the existent test to assert that the exception's message now contains the URL of the resource.

I will take care of this on merge.

fmbenhassine:
Rebased and merged as 4444e465a65381a57978f1bbd86350d07317f816. Thank you for your contribution!

PR #3919, opened 2021-05-21
Update README.md

Buzzardo:
Add a note to tell people to generate the project before generating the documentation. Running `./mvnw site` first causes a build failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ To generate the reference documentation, run:
 
     $ ./mvnw site
 
+NOTE: Generate the project before generating the reference documentation. The reference documentation build picks up content from the project build.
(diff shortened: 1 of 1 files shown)


benas:
There is actually no need to build the entire project to generate the docs, we only need to generate the javadocs first so they are included in the distribution when running `mvn site`. The README was implicit with regard to the order of commands, so I made that more explicit in 6b4c91632195dff3a5ac6059c7c69ac1063fad60.

Closing this in favour 6b4c91632195dff3a5ac6059c7c69ac1063fad60 since there is no need to build the entire project to generate the docs. Thank you for your PR anyway!

Issue #3917, opened 2021-05-20
Duplicate entries in the documentation distribution file


benas:
The distribution file for reference documentation seems to contain duplicate PDF entries which prevents the HTML docs from being correctly deployed during a release:

```
$ ./gradlew clean dist
...
$ unzip -l build/distributions/spring-batch-4.3.4-SNAPSHOT-docs.zip
Archive:  build/distributions/spring-batch-4.3.4-SNAPSHOT-docs.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  05-20-2021 15:59   api/
...
     4564  05-20-2021 15:59   api/package-list
        0  05-20-2021 15:59   reference/
        0  05-20-2021 15:59   reference/pdf/
  8281601  05-20-2021 15:59   reference/pdf/spring-batch-reference.pdf
  8281601  05-20-2021 15:59   reference/pdf/spring-batch-reference.pdf
        0  05-20-2021 15:59   reference/html/
...
---------                     -------
 52235708                     1438 files
```

Trying to unzip the archive requires a confirmation to override the file:

```
$ unzip build/distributions/spring-batch-4.3.4-SNAPSHOT-docs.zip
...
  inflating: api/package-list
   creating: reference/
   creating: reference/pdf/
  inflating: reference/pdf/spring-batch-reference.pdf
replace reference/pdf/spring-batch-reference.pdf? [y]es, [n]o, [A]ll, [N]one, [r]ename: ^C
```

On the docs server, this command fails with the following error:

```
$ unzip spring-batch-4.3.4-SNAPSHOT-docs.zip
error: invalid zip file with overlapped components (possible zip bomb)
```

This leads to a situation where the PDF version is deployed but the HTML version is skipped. Obviously this could be fixed by using the `unzip -o` flag, but the gradle task should be fixed instead to not include duplicates in the distribution file (for more efficiency in terms of file size when transferring the zip over the network).

benas:
Resolved with fa8063f.

PR #3921, opened 2021-05-24
Exclude duplicate output paths from asciidoctor

hpoettker:
Fixes https://github.com/spring-projects/spring-batch/issues/3917

Root cause seems to be https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/471


benas:
> An alternative solution seems to be upgrading to Gradle 6.6 or higher

Indeed, upgrading to Gradle 6.9 seems to fix the issue. Closing this in favour of fa8063fa74dbe398648805819456dcfdaeb2e3c0.

Thank you for looking into this!

PR #3933, opened 2021-06-08
Renamed setJobIncrementer to setJobInstanceIncrementer in JdbcJobInst…

### LINKED ISSUE #3929: Rename setJobIncrementer to setJobInstanceIncrementer in JdbcJobInstanceDao
The `DataFieldMaxValueIncrementer jobIncrementer` field in `JdbcJobInstanceDao` is actually an incrementer for job instance IDs and not job IDs (which does not make much sense as there is no `Job` entity as part of the batch domain classes).

This setter is better renamed to `setJobInstanceIncrementer` (similar to `jobExecutionIncrementer` and `stepExecutionIncrementer` in other DAOs) to be more explicit about the entity IDs it is incrementing. The old setter should be deprecated in favour of the new one.

parikshitdutta:
Closes #3929

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobInstanceDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobInstanceDao.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2019 the original author or authors.
+ * Copyright 2006-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobInstanceDao.java
```diff
@@ -301,11 +302,26 @@ public int getJobInstanceCount(@Nullable String jobName) throws NoSuchJobExcepti
...
+	/**
+	 * Setter for {@link DataFieldMaxValueIncrementer} to be used when
+	 * generating primary keys for {@link JobInstance} instances.
+	 *
+	 * @param jobIncrementer the {@link DataFieldMaxValueIncrementer}
+	 *
+	 * @since 5.0
+	 */
+	public void setJobInstanceIncrementer(DataFieldMaxValueIncrementer jobIncrementer) {
+		this.jobIncrementer = jobIncrementer;
```
benas, line 322: The parameter and field could also be renamed to `jobInstanceIncrementer` here. I will do that on merge.

benas:
Rebased and merged as 49134ea5338dca40dcf2b6a918e77e966dc7c46b. Thank you for your contribution.

PR #3934, opened 2021-06-09
Added default methods in listener interfaces

### LINKED ISSUE #3924: Add default methods in interfaces
Several core interfaces could benefit from using Java 8's default methods in interfaces feature. This has a couple of advantages:

* It makes client code more readable as it does not require it to implement all methods of a given interface.
* It reduces the code on the framework side since support classes (like `JobExecutionListenerSupport`, `StepExecutionListenerSupport`, `ChunkListenerSupport`, etc) become obsolete in favour of default methods.

This feature is similar to what has been done in other projects ([Here is an example](https://github.com/spring-projects/spring-framework/blob/main/spring-tx/src/main/java/org/springframework/transaction/support/TransactionSynchronizationAdapter.java#L32-L33) from Spring Framework) and the goal is to make sure the Spring (Batch) developer has a consistent experience across the portfolio.

parikshitdutta:
Closes #3924

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/ChunkListener.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/ChunkListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2018 the original author or authors.
+ * Copyright 2006-2021 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


benas:
Rebased and merged as fecf8a5c44919ed37ff63cfad1518e8640960e39. Thank you for your contribution!

PR #3938, opened 2021-06-09
BATCH-3936 StepScope TARGET_NAME_PREFIX scopedTarget

farnetto:
StepScope.TARGET_NAME_PREFIX changed to "scopedTarget." to match ScopedProxyUtils.TARGET_NAME_PREFIX and check in BatchScopeSupport#postProcessBeanFactory.


farnetto:
Possibly better just to reference constant in ScopedProxyUtils instead of copying the value? JobScope should probably be similarly changed if this change is ok.

fmbenhassine:
LGTM. Rebased and merged as 5fae33185fb92ac46f27d150404cc6be581b4749. Thank you for your contribution.

PR #3958, opened 2021-07-02
Make ScopeConfiguration publicly accessible

hpoettker:
Splitting the batch infrastructure and the actual job configurations into parent and child contexts, respectively, can be beneficial. Interestingly, this is not mentioned in the Spring Batch but [Spring Framework documentation](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#testcontext-ctx-management-ctx-hierarchies).

Unfortunately, custom scopes such as `StepScope` and `JobScope` are not visible in child contexts (https://github.com/spring-projects/spring-framework/issues/11749). Thus, the two scopes are only visible in the context that contains the annotation `@EnableBatchProcessing`, which would be the parent context in the above scenario. This leads to failing jobs in the child contexts if they use step or job scoped beans.

With this change, it is possible for users to import `ScopeConfiguration` into child contexts and use job and step scope within them as usual.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/AbstractBatchConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


benas:
Thank you for your PR. I agree with you, we should be able to use step/job scoped beans defined in child contexts without any issue. We only need to ensure that the changes suggested here are consistent/coherent with those that will be introduced in #3942 . So I will review your PR in details when addressing #3942 .

benas:
Rebased and merged as fce9a480801bbe211b66b4d20013b6bf4b87366c. Thank you for your contribution.

Issue #3963, opened 2021-07-12
Spring batch metrics is not aware of JobExecutionListener that modify job status


qwazer:
I have a Spring Boot 2.5.2 app with spring-batch-4.3.3
The app has a multi-step job and a custom JobStatusReportingListener 

```java
@Component
public class JobStatusReportingListener implements JobExecutionListener {

    private final Logger logger = LoggerFactory.getLogger(getClass());

    @Override
    public void beforeJob(JobExecution jobExecution) {
    }

    @Override
    public void afterJob(JobExecution jobExecution) {

        List<StepExecution> failedSteps = jobExecution
                .getStepExecutions()
                .stream().filter(v -> !v.getExitStatus().equals(COMPLETED))
                .collect(Collectors.toList());

        if (!failedSteps.isEmpty()) {
            jobExecution.setStatus(BatchStatus.FAILED);
            jobExecution.setExitStatus(ExitStatus.FAILED);
            logger.error("Job failed on steps {}", failedSteps);
        }
    }
}
```

The problem that metrics at `actuator/metrics/spring.batch.job` doesn't contain tags with status="FAILED" that is set by JobExecutionListener

fmbenhassine:
Thank you for opening this issue. By looking at [this snippet](https://github.com/spring-projects/spring-batch/blob/7babb341f5da9be00eff27b388268f77eb5eeaa0/spring-batch-core/src/main/java/org/springframework/batch/core/job/AbstractJob.java#L365-L374), I think this is a valid issue. In fact, the observation is stopped and the status is recorded *before* listeners are called. Therefore, any status change done in the listener is not reflected in the final metric.

That said, I would like to validate that with a failing test or a minimal complete example that reproduces the issue. This failing test or minimal example will be used to validate the bug as well as the fix. Please take some time to create a minimal example that reproduces the problem. To help you in reporting your issue, we have prepared a project template that you can use as a starting point. Please check the [Issue Reporting Guidelines](https://github.com/spring-projects/spring-batch/blob/main/ISSUE_REPORTING.md) for more details about this.

Thank you for your collaboration.

Issue #3846, opened 2021-02-08
StepExecution.getEndTime() returns null in StepExecutionListener.afterStep()


HomoEfficio:
**Bug description**

StepExecution.getEndTime() invoked inside StepExecutionListener.afterStep() returns null.

It should return a valid time, according to the [API docs](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/StepExecutionListener.html#afterStep-org.springframework.batch.core.StepExecution-) which says below:

>(StepExecutionListener.afterStep() is) Called after execution of step's processing logic (both successful or failed).

I already checked the similar isssue, https://github.com/spring-projects/spring-batch/issues/2710, but the comments in that issue do not seem to be consistent with the API docs.

**Environment**

Java version: AdoptOpenJDK 11
Spring Batch version: 4.3.1
Spring Boot version: 2.4.2

**Steps to reproduce**

Download and execute this spring boot application: https://github.com/HomoEfficio/scratchpad-spring-batch

The console result is like below:

```
[  restartedMain] i.h.s.s.s.SpringBatchSampleApplication   : Started SpringBatchSampleApplication in 0.812 seconds (JVM running for 1.259)
[  restartedMain] o.s.b.a.b.JobLauncherApplicationRunner   : Running default command line with: []
[  restartedMain] o.s.b.c.l.support.SimpleJobLauncher      : Job: [SimpleJob: [name=Sample Job]] launched with the following parameters: [{}]
[  restartedMain] o.s.batch.core.job.SimpleStepHandler     : Executing step: [Sample Step]
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : Sample TaskLet executed
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : stepName: Sample Step
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : stepStartedAt: Mon Feb 08 11:36:12 KST 2021
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : stepEndedAt: null  ===============> SHOULD BE A VALID TIME
[  restartedMain] o.s.batch.core.step.AbstractStep         : Step: [Sample Step] executed in 10ms
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : jobName: Sample Job
[  restartedMain] i.h.s.springbatch.sample.BatchConfig     : jobStartedAt: Mon Feb 08 11:36:12 KST 2021
[...]

PR #3964, opened 2021-07-12
Stop metrics timers after job listeners invocation

qwazer:
Stop metrics timers and set "end time of job execution" 1) after job listeners invocation and 2) immediately before persisting a job execution object.


fmbenhassine:
Thank you for the PR. But with this change, the listener is now called before setting the end time on the job execution, which will lead to an issue like https://github.com/spring-projects/spring-batch/issues/3846

fmbenhassine:
Closing this as explained in my previous comment: this will lead to issues like #3846 where the end time is `null` in `JobExecutionListener#afterJob` which would be a regression. Thank you for your time anyway 🙏 Feel free to open another PR if you find another way to fix #3963 (I will check on my side as well).

PR #3972, opened 2021-08-12
Add `CommandRunner` and `JvmCommandRunner`

### LINKED ISSUE #3955: Improve testability of SystemCommandTasklet
Currently, the `SystemCommandTasklet` is not so test-friendly, mainly due to the usage of `Runtime.getRuntime().exec(...)`:

https://github.com/spring-projects/spring-batch/blob/6b4c91632195dff3a5ac6059c7c69ac1063fad60/spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/SystemCommandTasklet.java#L103

For example, testing a job or a step with this tasklet defined does not work properly if the platform running the test does not have the target command available.

Would it make sense to decouple the `Runtime` dependency from this tasklet, so that a test can have more control over it? I would propose a pattern similar to the [SystemExiter](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SystemExiter.java) and [JvmSystemExiter](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/JvmSystemExiter.java).

I am happy to submit a proposal in case the idea is accepted.

scordio:
Fixes #3955.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/SystemCommandTasklet.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/SystemCommandTasklet.java
@@ -64,6 +64,8 @@ public class SystemCommandTasklet extends StepExecutionListenerSupport implement
 
 	protected static final Log logger = LogFactory.getLog(SystemCommandTasklet.class);
 
+	private static CommandRunner commandRunner = new JvmCommandRunner();
+
 	private String command;
 
 	private String[] environmentParams = null;
@@ -100,7 +102,7 @@ public class SystemCommandTasklet extends StepExecutionListenerSupport implement
 
 			@Override
 			public Integer call() throws Exception {
-				Process process = Runtime.getRuntime().exec(command, environmentParams, workingDirectory);
+				Process process = commandRunner.exec(command, environmentParams, workingDirectory);
 				return process.waitFor();
 			}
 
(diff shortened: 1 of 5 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/SystemCommandTasklet.java
```diff
@@ -142,6 +144,26 @@ else if (stopped) {
...
+
+	/**
+	 * Injection setter for the {@link CommandRunner}.
+	 *
+	 * @param commandRunner {@link CommandRunner} instance to be used by SystemCommandTasklet instance.
+	 */
+	public void setCommandRunner(CommandRunner commandRunner) {
+		SystemCommandTasklet.commandRunner = commandRunner;
+	}
+
```
scordio, line 166: Here I followed the same pattern used in the `CommandLineJobRunner`, please let me know if there is anything to change about it.

spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/SystemCommandTasklet.java
```diff
@@ -64,6 +64,8 @@
 
 	protected static final Log logger = LogFactory.getLog(SystemCommandTasklet.class);
 
+	private static CommandRunner commandRunner = new JvmCommandRunner();
```
scordio, line 67: I would like to write some tests that use a mocked `CommandRunner`. Would it make sense to add them in `SystemCommandTaskletIntegrationTests`, or should I create a separate `SystemCommandTaskletTests`?

spring-batch-core/src/test/java/org/springframework/batch/core/step/tasklet/SystemCommandTaskletIntegrationTests.java
```diff
@@ -180,6 +183,34 @@ public void testInterruption() throws Exception {
...
+	@Test
+	public void testCommandRunnerNotSet() throws Exception {
+		SystemCommandTasklet.presetCommandRunner(null);
+		try {
+			tasklet.afterPropertiesSet();
+			fail();
+		}
+		catch (IllegalArgumentException e) {
+			// expected
+		} finally {
```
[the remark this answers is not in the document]
fmbenhassine: We moved to JUnit Jupiter since then. I will update assertions as needed on merge.

fmbenhassine:
Rebased and merged as eccea800fb70cba8c9894e1e5dc4585f3016fd4a. Revisited in 4eefb1d2875ce3c3dc187af72e5c8f695611f190. Thank you for your contribution!

scordio:
Thanks for taking care of the updates!

PR #3975, opened 2021-08-15
Change collection and delete field access modifiers and Add getters f…

santfirax:
- Change access modifiers for collection and delete fields from private to protected in order to be access those when extending a class.
- Add Getter for collection and delete field in order to access those two fields.


fmbenhassine:
Rebased and merged as 5e515473de5b05b286ca2ec62380c3b380fde525. Please note that the `delete` flag was replaced with an enum in #4351. Therefore, I updated the getter accordingly. Moreover, since we provide getters now, there is no need to make the fields `protected` (to be consistent with the `template` field).

PR #3979, opened 2021-08-23
Optimize ExitStatus#addExitDescription

marschall:
Optimize ExitStatus#addExitDescription by reducing string allocations
through:

- avoid string allocation when the current description is empty
- avoid string allocation when the given description is empty
- avoid intermediate string allocation by allocating the correct buffer size

When profiling the string allocations in `ExitStatus#addExitDescription` where showing up. They were mostly showing up through `ExitStatus#and(ExitStatus)`.

ExitStatus-addExitDescription


fmbenhassine:
I love this kind of optimisations! LGTM 👍  Rebased and merged as b38226e810e5ba255c41c93cc79f7e997d8ddb5a. Thank you for your contribution!

Issue #816, opened 2019-01-29
PlatformTransactionManager bean is exposed unconditionally  [BATCH-2788]


spring-projects-issues:
**[Stéphane Nicoll](https://jira.spring.io/secure/ViewProfile.jspa?name=snicoll)** opened **[BATCH-2788](https://jira.spring.io/browse/BATCH-2788?redirect=false)** and commented

While this can't be changed easily as it will break backward compatibility, Spring Batch should not expose a `PlatformTransactionManager` unconditionally (see [SimpleBatchConfiguration](https://github.com/spring-projects/spring-batch/blob/42596920e64288717f6f24f90647dce3c9fe364a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/SimpleBatchConfiguration.java#L89)).

Spring Boot 2.1 has a mode enabled by default that throws an exception on startup if a bean is overridden. If a user defines its own transaction manager with Spring Batch, they will have to disable the flag as the application will fail. This isn't really expected as Spring Boot is supposed to handle and detect such scenario.

Arguably, Spring Boot should provide whatever `PlatformTransactionManager` and Spring Batch should be configured to use this. We can't implement this mode at the moment as Spring Batch exposes a `PlatformTransactionManager` itself.

This problem has been raised in several places (i.e. [BATCH-2294](https://jira.spring.io/browse/BATCH-2294)). IMO, the resolution is incorrect: this is not about customizing the transaction manager to use, this is about exposing a bean in the context that may not be Spring Batch's responsibility in the first place. Besides, it is very likely that the transaction manager instance the user provides via the `BatchConfigurer` is already a bean so exposing it again feels wrong in the first place.

---

**Affects:** 4.1.1

benas:
Related/Similar issue: https://github.com/spring-projects/spring-batch/issues/838.

marcingrzejszczak:
With this issue in Sleuth (https://github.com/spring-cloud/spring-cloud-sleuth/issues/1941) we've added support for instrumentation of `PlatformTransactionManager`s . I've ran the batch sample https://github.com/spring-cloud-samples/spring-cloud-sleuth-samples/tree/3.1.x/batch and the instrumented `PTM` wasn't called at all (even though it got instrumented). That means that currently Batch is ignoring any `PTM` wrapping and creates its beans outside of the standard Spring lifecycle (e.g. `BeanPostProcessors` can't warp its components).

PR #3981, opened 2021-08-26
Remove the unconditional exposure of the transaction manager as a bean

benas:
This PR addresses Issue #816. It removes the unconditional exposure of the transaction manager as a bean in the application context. The transaction manager is still taken from the `BatchConfigurer` and set where needed (ie on `JobRepository` and `StepBuilderFactory`) as previously done, but is not exposed anymore as a bean to prevent any clash with a user-defined transaction manager if any.

If no transaction manager is provided, a `DataSourceTransactionManager` will be configured by default as required by batch (without being exposed as a bean). Note that a `DataSource` will be mandatory starting from v5, so this seems a reasonable default to me.

Moreover, this change makes the Java config style consistent with the XML config style, where the transaction manager is not exposed as a bean by the `<batch:*>` namespace.


benas:
Rebased and merged as 0d3282554c153af112fc0ad75f5ad94152a196dd.

Issue #1098, opened 2016-04-22
StepBuilderFactory Only Supports Listener Annotations, Not Listener Interfaces [BATCH-2501]


spring-projects-issues:
**[Jeff](https://jira.spring.io/secure/ViewProfile.jspa?name=jefff)** opened **[BATCH-2501](https://jira.spring.io/browse/BATCH-2501?redirect=false)** and commented

When adding listeners to a step using the step builder, the step builder only broadcasts to listeners that use listener annotations, such as `@OnReadError`. Listeners that implement interfaces, such as "public class ExceptionListener implements ItemWriteListener", never receive broadcasted events.

> stepBuilderFactory.get("somethingFlatIdStep")
> .<TheObjectType, TheObjectType> chunk(25000)
> .reader(somethingFlatIdReader)
> .processor(new PassThroughItemProcessor\<TheObjectType>())
> .writer(somethingFlatIdWriter)
> .transactionAttribute(ConfigUtil.makeTransAttribute())
> .listener(new ExceptionListener())
> .listener(new AnnotationEventListener())
> .build();

Two files attached exemplifying a listener class that doesnt work (uses spring batch listener interfaces) and a listener clas that does work (uses listener annotations).

---

**Affects:** 3.0.6

**Attachments:**
- [AnnotationEventListener.java](https://jira.spring.io/secure/attachment/23377/AnnotationEventListener.java) (_1.95 kB_)
- [ExceptionListener.java](https://jira.spring.io/secure/attachment/23378/ExceptionListener.java) (_2.52 kB_)

fmbenhassine:
Resolved with #3989 .

Issue #3686, opened 2020-03-27
lost transactionAttribute when using chaining StepBuilder


hoyeol:
lost transactionAttribute when using chaining StepBuilder.

according to below code, there will be lost "transactionAttribute" already set.
`DefaultTransactionAttribute readOnlyTxAttribute = new DefaultTransactionAttribute();
readOnlyTxAttribute.setReadOnly(true);

stepBuilderFactory.get("stepName")
.transactionManager(txManager)
.<InputType, OutputType>chunk(CHUNK_SIZE)
.reader(reader)
.writer(writer)
.transactionAttribute(readOnlyTxAttribute)
.faultTolerant()
.backOffPolicy(backOffPolicy())
.retryLimit(3)
.retry(SomeException.class)
.build();`

next below code, does not lost transactionAttribute. (just only difference is *"where is transactionAttribute method call"*)
`
stepBuilderFactory.get("stepName")
.transactionManager(txManager)
.<InputType, OutputType>chunk(CHUNK_SIZE)
.reader(reader)
.writer(writer)
.faultTolerant()
.backOffPolicy(backOffPolicy())
.retryLimit(3)
.retry(SomeException.class)
.transactionAttribute(readOnlyTxAttribute)
.build();
`

it seems to broken compatibility between SimpleStepBuilder and FaultTolerantStepBuilder (and both classes AbstractTaskletStepBuilder what have transactionAttribute property.

there transactionAttributes will be missing on method calls. (there methods does not support copying "transactionAttribute")
- org.springframework.batch.core.step.builder.SimpleStepBuilder.faultTolerant
- org.springframework.batch.core.step.builder.FaultTolerantStepBuilder constructor
- org.springframework.batch.core.step.builder.SimpleStepBuilder constructor
- org.springframework.batch.core.step.builder.AbstractTaskletStepBuilder constructor
- org.springframework.batch.core.step.builder.StepBuilderHelper constructor

acktsap:
I can't reproduce it by a syntax error.

![Screen Shot 2021-09-30 at 4 47 06](https://user-images.githubusercontent.com/24649991/135338171-ee113441-16d7-4838-9aa5-0be5537281a7.png)

It looks because the transactionAttribute returns AbstractTaskletStepBuilder<B>, not SimpleStepBuilder.

[...]

PR #3989, opened 2021-09-05
Let fluent setters of SimpleStepBuilder return proper type

### LINKED ISSUE #773: ItemReadListener not being correctly registered after adding a StepExecutionListener [BATCH-2841]
**[Benne Otten](https://jira.spring.io/secure/ViewProfile.jspa?name=bolemeus)** opened **[BATCH-2841](https://jira.spring.io/browse/BATCH-2841?redirect=false)** and commented

I've stumbled upon this bug and it gave me quite a headache.
Consider the following Step configuration

```java
@Bean
public Step step1() {
    return steps.get("step1")
        .<SomeDto, SomeDto>chunk(10)
        .reader(reader())
        .processor(processor())
        .writer(writer())
        .listener(stepExecutionListener())
        .listener(itemReadListener())
        .build();
}
```

I would expect the last listener to register the ItemListener correctly, using the listener(ItemReadLiistener listener) method in the SimpleStepBuilder class, but the listener(Object listener) method in the AbstractTaskletStepBuilder gets invoked instead.

The reason for this is that the listener(StepExecutionListener listener) method resides in the StepBuilderHelper class, and that method returns an AbstractTaskletStepBuilder, and not a SimpleStepBuilder, what it should really return. In other words, the inheritance of SimpleStepBuilder > AbstractTaskletStepBuilder > StepBuilderHelper is messed up.

To illustrate, consider the following tweaked example...

```java
@Bean
public Step step1() {
    return steps.get("step1")
        .<SomeDto, SomeDto>chunk(10)
        .reader(reader())
        .processor(processor())
        .writer(writer())
        .listener(stepExecutionListener())
        .listener(itemReadListener())
        .listener(itemReadListener())
        .build();
}
```

Here, the first listener(itemReadListener) gets invoked in the AbstractTaskletStepBuilder class (which does nothing for an ItemReadListener) and the second listener(itemReadListener) gets invoked in the SimpleStepBuilder class (as it should).

---
No further details from [BATCH-2841](https://jira.spring.io/browse/BATCH-2841?redirect=false)

hpoettker:
Let fluent setters of `SimpleStepBuilder` return `this` with type `SimpleStepBuilder`. This makes the required order of some fluent setters a bit more lenient and prevents that `AbstractTaskletStepBuilder::listener` for parameters of type `Object` is invoked although an overloaded method specific for `SimpleStepBuilder` is intended.

Fixes #773 and #1098.


hpoettker:
I rebased the original commit on the current main branch.

I also added a commit that is functionally the same but uses less unchecked casts. It's a bit cleaner but also a bit noisier as it touches more classes.

SebastianGoeb:
Is this PR still active? I've also encountered the listener issue #773 , and would love to see this change merged.

hpoettker:
From my side, I'd definitely consider this PR active. If it needs work, I'll be happy to do it.

The `listener` method on the DSL can be confusing. And I think this PR would really help in this regard.

hpoettker:
I fixed the merge conflict that arised mainly from #4118 for the main branch.

There are now at least three open issues that are addressed by this PR: #773, #1098, and #3686. And I'd also consider it as having votes from the community.

It would be great to have a target release assigned to this.

fmbenhassine:
Rebased, squashed and merged as d280556d0202124df6c8f4a8fcbddee2b123dd73. Thank you for your contribution!

PR #3999, opened 2021-09-13
Devdynam0507/modify schema appendix docs

devdynam0507:
I use the mysql database. and I was trying to use Spring Batch
so I read the documentation and I was follow steps create Spring Batch Meta Table

The problematic part is:

~~~
CREATE TABLE BATCH_JOB_EXECUTION_CONTEXT  (
  JOB_EXECUTION_ID BIGINT PRIMARY KEY,
  SHORT_CONTEXT VARCHAR(2500) NOT NULL,
  SERIALIZED_CONTEXT CLOB,
  constraint JOB_EXEC_CTX_FK foreign key (JOB_EXECUTION_ID)
  references BATCH_JOB_EXECUTION(JOB_EXECUTION_ID)
) ;
~~~

~~~
CREATE TABLE BATCH_STEP_EXECUTION_CONTEXT  (
  STEP_EXECUTION_ID BIGINT PRIMARY KEY,
  SHORT_CONTEXT VARCHAR(2500) NOT NULL,
  SERIALIZED_CONTEXT CLOB,
  constraint STEP_EXEC_CTX_FK foreign key (STEP_EXECUTION_ID)
  references BATCH_STEP_EXECUTION(STEP_EXECUTION_ID)
) ;
~~~

'CLOB' is not exists type in mysql
so I did modified  'CLOB' -> 'BLOB' for mysql user

# Related references
https://dev.mysql.com/doc/refman/8.0/en/blob.html

### DIFF
--- a/spring-batch-docs/src/main/asciidoc/schema-appendix.adoc
+++ b/spring-batch-docs/src/main/asciidoc/schema-appendix.adoc
@@ -319,7 +319,7 @@ belongs. There may be more than one row associated with a given execution.
 * `SHORT_CONTEXT`: A string version of the `SERIALIZED_CONTEXT`.
 * `SERIALIZED_CONTEXT`: The entire context, serialized.
 
-> If you did create database table, but can't create table follow try something like below
+> If you did create mysql database table, but can't create table follow try something like below!
 ----
 CREATE TABLE BATCH_JOB_EXECUTION_CONTEXT  (
   JOB_EXECUTION_ID BIGINT PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


benas:
Thank you for opening this PR. However, have you checked the [Example DDL scripts](https://docs.spring.io/spring-batch/docs/4.3.x/reference/html/schema-appendix.html#exampleDDLScripts) section? This is where it is mentioned to look for the specific schema `schema-*.sql` of the target platform. So in your case, you should be looking at [schema-mysql.sql](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/resources/org/springframework/batch/core/schema-mysql.sql) where the mentioned type is `TEXT` and not `CLOB`.

The samples in the rest of the appendix are generic and used only to explain the purpose of each column.

fmbenhassine:
As mentioned in my previous comment, the samples in the appendix are generic. For platform-specific scripts, the documentation already mentions where to find the DDL files: [Example DDL Scripts](https://docs.spring.io/spring-batch/docs/4.3.x/reference/html/schema-appendix.html#exampleDDLScripts).

PR #4003, opened 2021-09-14
Instead of creating unnecessary collection objects, change them to immutable or static collections.

sjh836:
If a collection is created but not added, it can be refactored into an immutable (static empty) collection.
But in this case, I excluded it. For example, when using new ArrayList to convert Set to List ..

Is there something I missed?

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/scope/context/StepContext.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/scope/context/StepContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2019 the original author or authors.
+ * Copyright 2006-2021 the original author or authors.
  *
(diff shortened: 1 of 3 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/scope/context/StepContext.java
```diff
@@ -160,7 +160,7 @@ public Map<String, Object> getJobParameters() {
 
 	@SuppressWarnings({"rawtypes", "unchecked"})
 	public Map<String, Object> getPartitionPlan() {
```
fmbenhassine, line 162: This method was removed in e5c752b4b82057a80ed2a049770223201c309666, I will omit this change and its related test on merge.

fmbenhassine:
Rebased and merged as d9bf57e57288be6147c921c05ca9a4bb56735a13. Thank you for your contribution!

PR #4005, opened 2021-09-14
Add the method to expose the internal map as read-only in ExecutionContext

### LINKED ISSUE #4004: Consider exposing the internal map as read-only in ExecutionContext
The current code is exposing `this.map.entrySet()` instead of `this.map`.
JobContext and StepContext are converting to `Map` using `EntrySet`.
Also, the entrySet is being exposed as modifiable.

I think, If the map being exposed is read-only, there is no problem.
This does not violate this comment. (`It is a thin wrapper for a map that allows optionally for type safety on reads.`)

I want to add a method that exposes the inner map. if so no longer need to create maps unnecessarily.

sjh836:
Resolves #4004


fmbenhassine:
LGTM. Rebased and merged as 06e6bdbc98799bb100ba6fedd4e6ba876fd1be07. Thank you for your contribution!

PR #4007, opened 2021-09-27
Add '!items.isEmpty()' conditional expression in AsyncItemWriter 

ParadiseCHOI:
1. Modify unnecessary parameter expression, list.add(future.get()) to list.add(item)
2. Modify unnecessary conditional expression, f (cause != null  && cause instanceof Exception) to if (cause instanceof Exception)
3. Add conditional expression when items are empty
```
1. Unwrapped future's value have alreay been assigned to the item variable. So, 'item' variable should be delivered instead of future.get() in the list.add method.

2. The getCause method of the Throwable Object returns null if the cause is nonexistent or unknown. And instanceOf operator always returns false if the object is null. So, 'cause != null' expression is unnecessary.

3. While looking at Future related to asynchronous processing, I found that it was in use in AsyncItem (Processor, Writer) of spring-batch.
Looking at the internal implementation, it was written in violation of the processor's default behavior, null return, so I modified the code so that if items are empty in AsyncItemWriter, nothing is returned.
```


ParadiseCHOI:
I will open it again after updating the code.

PR #4034, opened 2021-12-09
Add @Nullable to StepExecution::endTime

acktsap:
StepExecution::endTime is initialized as null at creation. But it's not marked as `@Nullable`.
When step is in execution, endTime is null. So i used [null safe calls](https://kotlinlang.org/docs/null-safety.html#safe-calls) but it give me a warning like this.

Screen Shot 2021-12-09 at 7 59 18 PM

Marking`@Nullable` can remove the warning.

Screen Shot 2021-12-09 at 8 07 05 PM


benas:
LGTM. Rebased and merged as dce1525e87b9f9c30c0d0103c70f6709792400bf. Thank you!

Issue #757, opened 2019-11-12
Simple CompositeItemReader analagous to CompositeItemWriter and CompositeItemProcessor


spring-projects-issues:
**[Elliot Korte](https://jira.spring.io/secure/ViewProfile.jspa?name=ekcrisp)** opened **[BATCH-2857](https://jira.spring.io/browse/BATCH-2857?redirect=false)** and commented

There may be a reason for this, but why doesn't there exist a composite item reader? It would be useful to chain different readers in the same way as CompositeItemWriter and CompositeItemProcessor. There's a MultipleResourceItemReader but I'm wondering why does the only reader which uses this delegate pattern have the restriction of being "ResourceAware"?

---
No further details from [BATCH-2857](https://jira.spring.io/browse/BATCH-2857?redirect=false)

fmbenhassine:
> There may be a reason for this, but why doesn't there exist a composite item reader? It would be useful to chain different readers in the same way as CompositeItemWriter and CompositeItemProcessor. 

I couldn't find any context about why there is no composite item reader, but I guess the main reasons are how to respect the "forward-only" contract of an item reader and how to deal with restartability. Similar to the composite writer, a composite reader could delegate reading items to a list of delegates in their registration order (once a delegate returns `null`, it moves to the next delegate, and so on), but making this kind of composition restartable is quite challenging.

An example of such a composite item reader can be found [here](https://github.com/fmbenhassine/spring-batch-lab/tree/main/issues/gh757). I would be grateful for anyone who gives it a try and share feedback here. A typical use case for this is reading the same data from different datasources for aggregation (the example shows a composite reader that delegates to 3 readers, reading the same data from two files and a database table).

> There's a MultipleResourceItemReader but I'm wondering why does the only reader which uses this delegate pattern have the restriction of being "ResourceAware"?

I see no obvious reason for that, but since items could come from different resources, one should probably be able to know the original resource from which an item was read.

PR #4039, opened 2021-12-21
BATCH-2857 Added CompositeItemReader

1grzyb1:
As mentioned in BATCH-2857 there was a need for CompositeItemReader. For scenarios like reading from multiple data sources or reading from different tables without using `GROUP BY` in a query as it decreases performance.
My solution passes data from a reader which could be for example Cursor reader and passes it to chained fetchers for fetching additional values.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ItemFetcher.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ItemFetcher.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


andrev321:
Very poor quality code, pozdrawiam

romchellis (approved):
This is useful

fmbenhassine:
Thank you for this PR! I appreciate your time and effort, but for this kind of features, we expect to discuss how to introduce the feature before start working on it.

For example, I don't see any composition between delegate readers here (there is only one reader and multiple `ItemFetcher`). I think the OP in #757 was asking for a simple composite reader similar to the composite writer (ie expecting a list of delegate readers, without introducing a new API like `ItemFetcher`).

There is no plan to introduce new APIs for this feature. The goal is to agree on how to orchestrate delegate readers and how to deal with restartability. For this, I invite you to join the discussion in #757 and share feedback or ideas. Many thanks upfront!

PR #4043, opened 2022-01-01
Adjust h2 schema to work with v2.0.x

hpoettker:
Spring Batch is currently not compatible with 2.0.x of H2. This creates problems for users that want to upgrade. However, the DDL for H2 can be adjusted such that it works with both the currently used 1.4 and 2.0.

When H2 is upgraded to 2.0.204, the test `H2JobRepositoryIntegrationTests` still fails as Spring JDBC is not compatible with 2.0.x. But this can be fixed in a similar way: https://github.com/spring-projects/spring-framework/pull/27870


hpoettker:
As the required change in Spring JDBC has been merged, I amended the PR by upgrading the H2 dependency for the tests.

The new test is inspired by this commit by @sbrannen: https://github.com/spring-projects/spring-framework/commit/75e18114700b8e8cc0b95a3732ffaa2ceedf0a48

The same test also passes for the currently used H2 version 1.4.200.

I took the liberty to use some Java 17 features in the test. If you want to backport the PR to Spring Batch 4.x, please let me know and I'll change it to be Java 8-compliant.

benas:
Thank you for this PR and for all your efforts across the portfolio! Much appreciated.

LGTM. Rebased and merged as e267e1f7886347a3a6d28ac82aeb0edb047a4a5a.

> I took the liberty to use some Java 17 features in the test. If you want to backport the PR to Spring Batch 4.x, please let me know and I'll change it to be Java 8-compliant.

The adjustment should be backported to 4.x, I will take care of the required changes. Thank you!

Issue #4033, opened 2021-12-08
Remove JUnit4 dependency from spring-batch-test


murdos:
**Expected Behavior**

spring-batch-test should not have dependency to JUnit4, or should have an optional dependency to JUnit4.

**Current Behavior**

spring-batch-test has a compile dependency to JUnit4.

**Context**

In a project using only JUnit5, JUnit4 is in the test classpath when using spring-batch-test.

danilopiazza:
Hi, I opened pull request #4054 to address this issue.

PR #4054, opened 2022-02-06
Make JUnit Jupiter a required dependency in test 

danilopiazza:
JUnit Jupiter is now a required dependency in spring-batch-test, whereas
JUnit 4 has become optional.

Issue #4033

Spring Batch Test's own unit tests are still using JUnit 4; they have not been migrated to JUnit Jupiter in this pull request.


benas:
Thank you for this PR. I think junit-jupiter should remain optional as well, as there are teams using TestNG only. With this PR, they will have junit-jupiter in their classpath. If you can update the PR to only make junit4 optional, it would be good to merge. Otherwise, please let me know and I will take care of it. Thank you.

danilopiazza:
Sorry for being late. I'm working on it.

danilopiazza:
[`AssertFile`](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-test/src/main/java/org/springframework/batch/test/AssertFile.java) requires either JUnit 4 or JUnit Jupiter, because it calls `assertEquals`. Should it call Spring Framework's [`Assert`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/Assert.html) instead?

danilopiazza:
> If you can update the PR to only make junit4 optional, it would be good to merge.

Done! All the unit tests pass (I ran `./mvnw clean test`). I updated both the commit message and the PR description. Thank you!

Issue #2244, opened 2009-07-14
IllegalArgumentException thrown from FlatFileItemReader where IllegalState would be more appropriate [BATCH-1337]


spring-projects-issues:
**[Iwein Fuld](https://jira.spring.io/secure/ViewProfile.jspa?name=iwein)** opened **[BATCH-1337](https://jira.spring.io/browse/BATCH-1337?redirect=false)** and commented

```
public void afterPropertiesSet() throws Exception {
		Assert.notNull(lineMapper, "LineMapper is required");
	}
```

Should use Assert.state instead imo.

---

**Affects:** 2.0.1

danilopiazza:
Hi, I opened PR #4055 to address this issue.

PR #4055, opened 2022-02-07
Throw IllegalStateException from afterPropertiesSet

danilopiazza:
Consistently use Assert.state in the afterPropertiesSet() methods to
throw IllegalStateException instead of IllegalArgumentException when
some properties are missing and/or invalid.

Issue #2244


fmbenhassine:
Rebased and merged as fc0ec01ff806973010562b8d2ce20b0c1e5c8364. Thank you for your contribution!

Issue #4049, opened 2022-01-24
Best to distinguish CreateTime and StartTime in StepExecution


lcmarvin:
I need to record create time of StepExecution, but I found it only has a filed `startTime`, though this field has been assigned value when StepExecution is created.
```
private volatile Date startTime = new Date(System.currentTimeMillis());
```
But its value will be updated when StepExecution gets started in AbstractStep. 
```
stepExecution.setStartTime(new Date());
stepExecution.setStatus(BatchStatus.STARTED);
```

So, maybe an individual `createTime` field is more suitable, just like what it is in JobExecution.
```
private volatile Date startTime = null;
private volatile Date createTime = new Date(System.currentTimeMillis());
```

lcmarvin:
This requires following changes:
1. Schema updates, table `bath_step_execution` needs a new column `create_time`.
2. Class `StepExecution `needs a new filed `createTime`.
3. SQL in `org.springframework.batch.core.repository.dao.JdbcStepExecutionDao#SAVE_STEP_EXECUTION`, `org.springframework.batch.core.repository.dao.JdbcStepExecutionDao#GET_RAW_STEP_EXECUTIONS`, and `org.springframework.batch.core.repository.dao.JdbcStepExecutionDao#GET_LAST_STEP_EXECUTION` need to be updated to include `create_time`.

lcmarvin:
#4057

lcmarvin:
Could you please link another pull request #4099 to this issue? Since a wrong rebase was done by mistake in #4057, so I close it.

Sorry for that, Thanks！

PR #4057, opened 2022-02-08
Distinguish CreateTime and StartTime in StepExecution.

lcmarvin:
- Add field createTime for StepExecution.
- Add column create_time for batch_step_execution and make start_time default null.
- Make getLastStepExecution order by createTime instead of startTime.


spring-batch-core/src/main/java/org/springframework/batch/core/StepExecution.java
```diff
@@ -243,6 +245,24 @@ public void setRollbackCount(long rollbackCount) {
 		this.rollbackCount = rollbackCount;
 	}
 
+	/**
+	 * Gets the time this execution created
```
benas, line 249:
```suggestion
	 * Gets the time this execution was created
```

spring-batch-core/src/main/java/org/springframework/batch/core/StepExecution.java
```diff
@@ -243,6 +245,24 @@ public void setRollbackCount(long rollbackCount) {
...
+	 *
+	 * @return the time when this execution was created.
+	 */
+	public Date getCreateTime() {
+		return createTime;
+	}
+
+	/**
+	 * Sets the time this execution created
```
benas, line 258:
```suggestion
	 * Sets the time this execution was created
```

benas:
Thank you for opening #4049 and for contributing this PR! Can you please add/amend migration scripts in `src/main/resources/org/springframework/batch/core/migration/5.0` to alter the `BATCH_STEP_EXECUTION` table and add the new column? This is required for existing apps to upgrade their schema to v5. Thank you upfront.

Other than that, the change set LGTM 👍

lcmarvin:
Thanks for adding it to milestone，I am trying to add scripts for migration.

PR #4063, opened 2022-02-11
Remove SQLLite Batch database tables before starting tests

cppwfs:
SQL Lite integration tests sporadically fails because tables already exist.   This removes those tables before starting.


benas:
LGTM. Rebased and merged. Thank you!

PR #4066, opened 2022-02-14
Remove benign [WARNINGS] from batch build

cppwfs:
This is done by:
1) Enabling quiet for doc build so that it, "Shuts off non-error and non-warning messages," so that builders can focus on warnings.
2) Disable missing javadoc warnings.   We will handle missing javadoc warnings in subsequent PRs.
3) Fix existing warnings that are in the batch code base.

This reduces the number of [WARNING]s from 7171 down to 52


benas:
LGTM. Rebased and merged as 710d89ce9dc6dd9c9e56bec1960154aa37c29a1b. Thank you!

Issue #4087, opened 2022-04-03
ResultSet.getLong(), ResultSet.getDouble() return primitive value make restart job work incorrect.


cnlin:
**Bug description**
In SpringBatch 1.x~4.x , JobParameter supports null value , so save in database will be null. While `java.sql.ResultSet#getLong`  or `java.sql.ResultSet#getDouble` return primitive type, a round trip from database will finally get 0 or 0.0. It work uncorrectly in `jobOperator.restart`. 
That may cause:
1. `jobOperator.restart` will create new JobInstance, not act like restart.
2. If null is meaningful in user project, that could cause the different result, because it restart an JobParameter with 0 or 0.0 value, not null anymore.

**Environment**
Spring Batch version: 1.x ~ 4.x, using `JdbcJobExecutionDao` (JobParameter unsupported null value in 5.x )

**Steps to reproduce**
1.  Creating an JobParameter with null value that in `java.lang.Long` or  `java.lang.Double` type.
2. Creating an job will failed in process.
3. Using `jobOperator.restart` to restart that JobInstance.
4. Check the count of JobInstance

**Expected behavior**
The `jobOperator.restart` should work right.

benas:
That's a valid issue against v4, thank you for raising it. This is exactly the type of issues behind the decision of not accepting `null` job parameters anymore in v5. The fix will be merged in v4.3.6.

benas:
Resolved with #4089 .

PR #4089, opened 2022-04-03
BATCH-4087 Return null by ResultSet#getLong or ResultSet#getDouble when the actual value that is null save in database

cnlin:
Return null by ResultSet#getLong or ResultSet#getDouble when the actual value that is null save in database.

Issue #4087

I think it should be merge into 4.x at least, while 5.x is not supported null value in JobParameter at all.


benas:
LGTM 👍  Rebased, squashed and merged as fdd253a7e6e68d02b5f4c4d36751d132e4821ac2. Thank you for reporting the issue and for contributing a fix!

PR #4099, opened 2022-04-16
Distinguish CreateTime and StartTime in StepExecution.

lcmarvin:
- Add field createTime for StepExecution.
- Add column create_time for batch_step_execution and make start_time default null.
- Make getLastStepExecution order by createTime instead of startTime.
- Add migration scripts for 5.0.

This is to replace #4057, sorry for that a wrong rebase was done by mistake in that PR. 
Thanks.


benas:
Thank you for the updates in this new PR. Rebased and merged as 464985bf0a61d0f94d1616949ee84a8d97ac5b0b.

I noticed that a migration script for SQLite was missing, so I added one in a722cd2f53c25d278b9df93c127cf3f59eb352f6.

PR #4112, opened 2022-05-10
Updated AssertFile to use Jupiter package names

### LINKED ISSUE #4111: ClassNotFound Exception when using AssertFile
When using `AssertFile.assertLineCount` with Spring Boot 3.0.0-SNAPSHOT a CNFE is thrown because org.junit.Assert package name has been changed to `org.junit.jupiter.api.Assertions`.

cppwfs:
resolves #4111


spring-batch-samples/src/test/java/org/springframework/batch/sample/AMQPJobFunctionalTests.java
```diff
@@ -34,7 +35,7 @@
  * queue for inspection.</p>
 */
 
-@RunWith(SpringJUnit4ClassRunner.class)
+@ExtendWith(SpringExtension.class)
 @ContextConfiguration(locations = { "/simple-job-launcher-context.xml", "/jobs/amqp-example-job.xml", "/job-runner-context.xml" })
```
hpoettker, line 39: Not really important, but if `@ContextConfiguration` is replaced with `@SpringJUnitConfig`, the `@ExtendWith(SpringExtension.class)` can be omitted, which makes the code a bit leaner.

`@SpringJUnitConfig` is basically `@ConfigurationConfiguration` but meta-annotated with `@ExtendWith(SpringExtension.class)`.

fmbenhassine: +1 to using `@SpringJUnitConfig` here and in every test class changed in this PR. Another PR could be opened to update all tests cases to use this annotation where appropriate.

hpoettker: The tests have been migrated to JUnit 5 and `@SpringJUnitConfig` in #4173. When the PR is rebased, the tests in the sample module shouldn't require much change.

fmbenhassine:
I like how we removed the dependency to any testing library in [e90330b](https://github.com/spring-projects/spring-batch/pull/4112/commits/e90330ba762332e3dff65cf6d11175575ec87c18)! 👍 I'm not sure what would be the best exception to throw when the file comparison fails, but if you think `IllegalStateException` is not ideal, we can create a custom exception provided by `spring-batch-test`.

In hindsight, everything in that `AssertFile` class is already provided by modern testing libraries (JUnit 5, AssertJ, etc) today, so I wonder if that class should be deprecated for removal from Batch in favor of what is provided by other libraries. Wdyt?

fmbenhassine:
Thank you for the update. Rebased, squashed and merged as 8ca9802da47fe3506a742f71d5feba6ca998d8d4. I noticed that the usage of `@SpringJUnitConfig` was reverted compared to `main`, so I've put it back in 7e535afb7316c51e974bbbdba9a5ae2ef8d7527e. Thank you for this contribution!

Issue #3987, opened 2021-09-01
Sorting in JdbcJobExecutionDao.GET_RUNNING_EXECUTIONS makes no sense


marschall:
`JdbcJobExecutionDao#GET_RUNNING_EXECUTIONS` currently has `order by E.JOB_EXECUTION_ID desc` however `JdbcJobExecutionDao#findRunningJobExecutions(String)` returns a `HashSet` therefore the iteration order is undefined. We should either

* remove the `order by`
* return a `TreeSet`

benas:
Thank you for raising this. The contract of `JobExecutionDao#findRunningJobExecutions` does not specify the order in which running executions are returned (hence the `Set` return type). So clients should not be expecting any order or any sorted data structure. I think the `order by` clause is irrelevant here and could be removed.

PR #4113, opened 2022-05-11
Make the meaning of running status consistent

### LINKED ISSUE #1483: BatchStatus#isRunning() is not consistent with JobExplorer#findRunningJobExecutions(String) or JobExecution#isRunning()
**[Alexis SOUMAGNE](https://jira.spring.io/secure/ViewProfile.jspa?name=almex)** opened **[BATCH-2114](https://jira.spring.io/browse/BATCH-2114?redirect=false)** and commented

In STOPPING status should be considered as a "running" status.
The endTime is null until job status is STOPPED. Which means that the job did not recieved interuption signal yet.

Simply the code should be :

```java
public boolean isRunning() {
   return this == STARTING || this == STARTED || this == STOPPING;
}
```

---

**Affects:** 2.2.2

lcmarvin:
- A running status is STARTING, STARTED, or STOPPING.

 Resolves #1483


lcmarvin:
I noticed in #3987 that the order by cause could be removed, so I reverted the change of HashSet to LinkedHashSet.

fmbenhassine:
LGTM. Rebased and merged as bbf35b79f7f830e50545de2bbbf191382e3e7dd4. Thank you for your contribution!

PR #3945, opened 2021-06-17
refactor: simplify boolean expression

gregorriegler:
Co-authored-by: Moderne <team@moderne.io>

Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Sign the [contributor license agreement](https://cla.pivotal.io/sign/spring)
* Rebase your changes on the latest `master` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission

For more details, please check the [contributor guide][1].
Thank you upfront!

[1]: https://github.com/spring-projects/spring-batch/blob/master/CONTRIBUTING.md

PR #4124, opened 2022-06-07
Migrate Spring Batch Integration to JUnit Jupiter

hpoettker:
As of now, most of Spring Batch's tests are written with JUnit 4 which works mostly fine but lacks some niceties (e.g. parameterized tests are a bit cumbersome), and can lead to issues when integrating with test frameworks of other projects. 😉 

This PR migrates the tests of Spring Batch Integration fully to JUnit 5 such that JUnit 4 is no longer on the classpath of the module.

The changes are optimized for reviewability in the sense that I tried to keep them minimal. E.g. all the `public` modifiers are still there as their removal would be very noisy.

The most part of the PR has been done by automation. Only `AsyncItemProcessorMessagingGatewayTests` and `PollingAsyncItemProcessorMessagingGatewayTests` were migrated manually for the most part.

Please let me know if you're interested in the same for the other modules.

### DIFF
--- a/spring-batch-integration/pom.xml
+++ b/spring-batch-integration/pom.xml
@@ -111,16 +111,10 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-			<groupId>junit</groupId>
-			<artifactId>junit</artifactId>
-			<version>${junit.version}</version>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<version>${junit-jupiter.version}</version>
 			<scope>test</scope>
-			<exclusions>
-				<exclusion>
-					<groupId>org.hamcrest</groupId>
-					<artifactId>hamcrest-core</artifactId>
-				</exclusion>
-			</exclusions>
 		</dependency>
 		<dependency>
 			<groupId>org.slf4j</groupId>
--- a/spring-batch-integration/src/test/java/org/springframework/batch/integration/SmokeTests.java
+++ b/spring-batch-integration/src/test/java/org/springframework/batch/integration/SmokeTests.java
@@ -1,10 +1,24 @@
+/*
+ * Copyright 2008-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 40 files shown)


benas:
This is awesome! Thank you for contributing this update, REALLY appreciated 👍 

> The changes are optimized for reviewability in the sense that I tried to keep them minimal. E.g. all the public modifiers are still there as their removal would be very noisy.

That noise is not issue for the review. If we decide to move to JUnit 5, let's do it in a clean way from the beginning. Please go ahead and apply all the best practices recommended by JUnit 5 in this PR (and others).

> The most part of the PR has been done by automation. 

Nice! Just curious, which automation tool have you used? We had an internal discussion about this kind of migrations and we were planning to use [open-rewrite](https://docs.openrewrite.org/tutorials/migrate-from-junit-4-to-junit-5) for it. We had a successful PoC by the team at [moderne.io](https://www.moderne.io) in #3945.

> Please let me know if you're interested in the same for the other modules.

Yes please! a PR per module should work.

Please note that we are planning to review testing support in Spring Batch in [5.0.0-M5](https://github.com/spring-projects/spring-batch/milestone/134), so those PRs would be reviewed only when we start working on that milestone. Many thanks upfront.

hpoettker:
Thanks for the detailed and encouraging feedback! I'm also quite intrigued by the promises of OpenRewrite and Spring Boot Migrator. This PR is basically a by-product of me having a look at them.

To explore the problem space and to create a baseline for a fair comparison, I approached the problem with `sed` commands, quick-and-dirty scripting, and manual finishing touches. That's what I euphemistically referred to as "automation". 😄 

I set out with the expectation to give up at some point but got surprisingly far as the programming style is very consistent for a code base of this size and age. The only pain was the migration from expected exceptions at the `@Test` level to `assertThrows` around the specific throwing line. I didn't want to wrap the whole body of the functions and the heuristic that it's most likely the last line didn't really work out. But that's a problem that I don't expect automated tools to solve based on static code analysis.

I'll try out the fully automated solutions in the next days. It would be fun to discuss this kind of PR on the level of recipes to be selected. I'll keep you posted.

hpoettker:
I've replaced the explicit catches of expected exceptions with `assertThrows`. Furthermore, I've removed unnecessary public modifiers and `throws` clauses, used static imports for assertions, and replaced the remaining explicit catching of expected exceptions.

I think the PR now has a nice scope for a JUnit 5 migration. Please let me know whether you agree or if you'd increase or decrease the scope.

Something that could be considered is a migration to AssertJ. But I think this PR would get unmanageable, if it were included here.

hpoettker:
As 5.0.0-M4 has been released, I've squashed and rebased the latest state on the commit with the tag `5.0.0-M4`.

PR #4125, opened 2022-06-07
Migrate Spring Batch Infrastructure to JUnit Jupiter

hpoettker:
This is a follow-up on #4124. It's quite a high number of touched files but I don't think there is a single change that's in itself interesting. If it helps, I can also split this up e.g. in tests from JUnit 3 and from JUnit 4 or something similar.

JUnit 4 is no longer on the classpath of the module.

Together with #4124 and #4112, it feels like the biggest part of migrating all tests to JUnit 5 is already done. The biggest open part is Spring Batch Core which shouldn't take long either. So far this migration is not on any roadmap but I think it would be nice to do it. With the release of Spring Batch 5, there will also be 5 years of JUnit 5. Perfect moment! 🥳

### DIFF
--- a/spring-batch-infrastructure/pom.xml
+++ b/spring-batch-infrastructure/pom.xml
@@ -170,16 +170,16 @@
 
 		<!-- test dependencies -->
 		<dependency>
-			<groupId>junit</groupId>
-			<artifactId>junit</artifactId>
-			<version>${junit.version}</version>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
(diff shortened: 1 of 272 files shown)


fmbenhassine:
Rebased and merged as 83d01fbfca6554ad609cf502a810b4da4da6f391. I noticed that some assertion updates were missed, I fixed them in 932e93428e7badfb3bb39df50e1dd23e1e811d19.

Other PRs will follow. I created #4166 to track the progress of the migration of all modules.

Thank you very much for this huge update!! Really appreciated 👍

PR #4136, opened 2022-06-17
Fix tests catching nested exceptions

hpoettker:
This PR adjusts some tests that are currently failing due to the upstream change spring-projects/spring-framework#25162.

I also added a `generateUniqueName(true)` in `ChunkMessageItemWriterIntegrationTests` as without it `RemoteChunkFaultTolerantStepJdbcIntegrationTests` stalls on my machine when running subsequently.


benas:
Rebased and merged as 451db9a968f440665c592ff8df1b1dc1b4cd6a1f. Thank you for your contribution!

PR #4143, opened 2022-06-30
add final field to APIs

brick0123:
Since these fields are not reassigned, added final.
- `JobBuilderFactory`, `StepBuilderFactory` and `RunIdIncrementer`

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/JobBuilderFactory.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/JobBuilderFactory.java
@@ -24,11 +24,11 @@ import org.springframework.batch.core.repository.JobRepository;
  *
  * @author Dave Syer
  * @author Mahmoud Ben Hassine
- *
+ * @author Jinho Han
  */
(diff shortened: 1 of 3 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/configuration/annotation/JobBuilderFactory.java
```diff
@@ -24,11 +24,11 @@
  *
  * @author Dave Syer
  * @author Mahmoud Ben Hassine
- *
+ * @author brick0123
```
fmbenhassine, line 27: Please use your real name here and in git commits, see https://github.com/spring-projects/spring-batch/blob/main/CONTRIBUTING.md#use-your-real-name-in-git-commits. Thank you upfront.

brick0123: I modified it. thank you

fmbenhassine:
Rebased and merged as fa70bcd775e02a79a2b53a768cda0b1361a4bba7. Thank you for your contribution!

PR #4145, opened 2022-07-04
Fix non-nullable columns in MySQL migration for Spring Batch 4.3

hpoettker:
The database migration for MySQL to Spring Batch 4.3 currently misses `NOT NULL` constraints from `BATCH_JOB_EXECUTION.CREATE_TIME` and `BATCH_STEP_EXECUTION.START_TIME`.

If the current migration is applied to a Spring Batch database < 4.3, it effectively removes the existing `NOT NULL` constraints although they are still contained in `schema-mysql.sql`.


fmbenhassine:
LGTM. Rebased and merged as 6bbd9d8379213b500d95993ec9ecd907ac490f23. Thank you for your contribution!

PR #4171, opened 2022-07-30
Fix `RecordFieldExtractorTests`

hpoettker:
The `RecordFieldExtractorTests` in Spring Batch Infrastructure have been failing since the migration to JUnit 5. Two `assertThrows` are missing or misplaced, respectively.


fmbenhassine:
The build passed on our internal CI infrastructure since the migration to JUnit 5. I don't see what change in this PR would make it pass instead of failing. But the changes make sense to me, they make tests more consistent with the rest of the suite. Thank you for the PR anyway.

hpoettker:
I am surprised that it passed in the CI.

It did not when I executed `package` locally. And this is also expected: The test `testSetupWithNonRecordTargetType` tests that an `IllegalArgumentException` is thrown. Before this commit the test only contained the throwing statement without any assertion or `catch`.

Issue #4166, opened 2022-07-27
Migrate tests to JUnit Jupiter


fmbenhassine:
This issue is to track the migration of the test suite from JUnit 4 to JUnit 5:

- [x] spring-batch-infrastructure: #4125
- [x] spring-batch-core: #4172 
- [x] spring-batch-test: #4174 
- [x] spring-batch-integration: #4124
- [x] spring-batch-samples: #4173

fmbenhassine:
Refinements like those done for `spring-batch-infrastructure` in #4175 can be done later for other modules. Improving test suites is a non-ending task, so I'm closing this for now as it is already a huge improvement to Spring Batch's test suite.

A BIG thank you to @hpoettker for this absolutely amazing contribution! REALLY appreciated 👍

PR #4172, opened 2022-07-31
Migrate Spring Batch Core to JUnit Jupiter

hpoettker:
Migrates the tests of Spring Batch Core to JUnit Jupiter as part of #4166.


hpoettker:
The PR is closed but the change is not committed to `main`. In fact, the test of Core on `main` are still on JUnit 4.

Is this intended?

fmbenhassine:
No, my mistake. I just pushed the commit. Thank you.

PR #4175, opened 2022-07-31
Polish Spring Batch Infrastructure tests

hpoettker:
This PR is in scope of #4166 polishes the tests in Spring Batch Infrastructure that have been migrated to JUnit 5 in #4125.

The goal of the PR is to improve the readability of the tests by reducing visual clutter and using more concise test constructs where directly available, i.e.

- test classes and methods are now package-private where possible
- not-required `throws` clauses are removed from test method signatures
- all `Assertions` are changed to static imports
- `assertThrows` and `assertDoesNotThrow` are used where previoulsy `try`, `catch`, and `fail` were used in a much more verbose style
- `@MockitoSettings` is used to set strictness of mocking at the class level

Some `final` modifiers have been added to class variables to make clearer whether they are expected to change or not during the tests, respectively.


hpoettker:
I rebased on `main` to fix the merge conflict with #4171. `RecordFieldExtractorTests` is no longer changed in this PR.

fmbenhassine:
> I rebased on main to fix the merge conflict with https://github.com/spring-projects/spring-batch/pull/4171. RecordFieldExtractorTests is no longer changed in this PR.

Great! I was about to do it locally. Thank you for the update!

PR #4176, opened 2022-08-15
Let every table have a primary key for MySQL

### LINKED ISSUE #4146: Improve replication friendliness of the MySQL schema
### Status quo

The table `BATCH_JOB_EXECUTION_PARAMS` currently has no primary key.

For a single node MySQL deployment, this does not cause any issue. For a setup of multiple nodes with replication, this is however not ideal as it causes performance problems with row-based replication and leads to ambiguity when calculating check sums over tables to validate the integrity of the replication.

To prevent such issues, MySQL 8 offers the system variable [`sql_require_primary_key`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_sql_require_primary_key). If set to `ON`, it is currently not possible to deploy the Spring Batch schema.

### Suggestion

As job parameters are handled by Spring Batch internally as maps, it already works with the assumption that job parameter keys are unique for each job execution. This allows to declare a primary key on `BATCH_JOB_EXECUTION_PARAMS` consisting of `JOB_EXECUTION_ID` and `KEY_NAME`.

It would be nice if this (or any other) primary key were defined on `BATCH_JOB_EXECUTION_PARAMS` in the MySQL schema.

The tables `BATCH_JOB_SEQ`, `BATCH_JOB_EXECUTION_SEQ`, `BATCH_STEP_EXECUTION_SEQ` also do not have a primary key. As they contain only a single row, respectively, this is not a problem in practice. But if the column `UNIQUE_KEY` were to be declared as primary key, this would allow a seamless deployment to a MySQL instance with `sql_require_primary_key=ON`.

hpoettker:
Resolves #4146.

The PR sets a primary key for every table of the Spring Batch schema for MySQL.

Currently, the table `BATCH_JOB_EXECUTION_PARAMS` has no unique key. With MySQL, this has disadvantages in replication setups typically used in production: For example, it can cause performance problems with row-based replication, prohibits the use of parts of the Percona Toolkit, and limits the use of Galera cluster.

Setting not only unique keys but primary keys has additional advantages. In particular, `sql_require_primary_key` can be set to `ON`.


cppwfs:
Hello @hpoettker,
Thank you for all your contributions to this and other projects.
The batch team had a discussion on this issue and it was determined that `JobParameters` are value objects and not entities.    You can read more about the foundation for this decision in the following documentation:

* [JobParameters javadoc](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/JobParameters.html) 
* The last paragraph of the `JobParameters` section of the [reference documentation](https://docs.spring.io/spring-batch/docs/current/reference/html/schema-appendix.html#metaDataBatchJobParams)

The documentation does state that you can add a primary key with a database generated key without causing any issues to the framework itself.

Again thank you for opening this issue.

hpoettker:
Hi @cppwfs, thanks for the pointer to the documentation. I wasn't aware that the issue is discussed there.

I think a more opinionated take on indexes by the framework would be nice but the reasoning in the documentation also makes complete sense. I'll close both the PR and the issue.

Issue #4151, opened 2022-07-14
Add ExecutionContextSerializer implementation based on Gson


fmbenhassine:
The goal of this issue is to add an implementation of `ExecutionContextSerializer`  based on [Google's Gson](https://github.com/google/gson), to give our users the choice between at least two implementations (ie Jackson and Gson) out-of-the-box, as mentioned in #4140.

fmbenhassine:
For some types, a serialization round trip with `gson` does not work OOTB. The following tests are failing with `gson` `2.9.1`:

```
@Test
void testGsonSerializationRoundTripForDate() {
    Gson gson = new Gson();
    Date src = new Date();
    String s = gson.toJson(src);
    Date date = gson.fromJson(new StringReader(s), Date.class);
    Assertions.assertEquals(date, src);
}

@Test
void testGsonSerializationRoundTripForLocalDate() {
    Gson gson = new Gson();
    LocalDate src = LocalDate.now();
    String s = gson.toJson(src);
    LocalDate date = gson.fromJson(new StringReader(s), LocalDate.class);
    Assertions.assertEquals(date, src);
}
```

ssjf409:
Hi. @fmbenhassine

Could I handle this issue? If you are okay, I want to try to implement it.
I already took a look at your [draft](https://github.com/spring-projects/spring-batch/pull/4185).

Do I have to make my branch from [GH-4151](https://github.com/fmbenhassine/spring-batch/tree/GH-4151) of your repository?

fmbenhassine:
@ssjf409 Sure! Thank you for your offer to help. You can start from the latest `main` branch if you want to contribute a new implementation.

My attempt was almost done, except that I couldn't figure out how to make `gson` serialize the type information (similar to jackson which writes an attribute `@class: fully.qualified.class.name`). I believe this should be configured on the `gson` instance, but I could not find how at the time. Do you know how to do that? I prefer digging further and try to delegate this to the serializarion library than handling it manually in the execution serializer implementation of Spring Batch.

ssjf409:
Thanks for your guide.👍

Actually. I didn't think about that you mentioned (`I couldn't figure out how to make gson serialize the type information (similar to jackson which writes an attribute @class: fully.qualified.class.name)`). I just it is enough to meet to condition for roundtrip.
However, I try to find a way.

fmbenhassine:
[...]

PR #4185, opened 2022-09-01
Add ExecutionContextSerializer implementation based on Gson

fmbenhassine:
This is a draft PR. Gson seems to deserialize custom types to `Map` instances instead of the original type (it looks like type information is not serialized).

TODO: Need to figure out if this should be done in the implementation of the serializer or as a configuration of the gson instance.

Issue #4151


fmbenhassine:
I am closing this PR as gson is now in maintenance mode: https://github.com/spring-projects/spring-batch/issues/4151#issuecomment-3422668467.

PR #4186, opened 2022-09-04
Upgrade Derby to version 10.16.1.1

hpoettker:
As of version 10.15.x, the artifact `derbytools` needs to be included to use Derby as an embedded `DataSource`: https://db.apache.org/derby/releases/release-10.15.1.3.html

The class `DerbyDataSourceFactoryBean` is deleted from the module Spring Batch Core as it is not used in that module.


fmbenhassine:
LGTM. Rebased and merged. Thank you for your contribution!

Issue #1028, opened 2017-01-31
Upgrade Jettison to 1.3.x [BATCH-2575]


spring-projects-issues:
**[Derek Lewis](https://jira.spring.io/secure/ViewProfile.jspa?name=dlewis1788)** opened **[BATCH-2575](https://jira.spring.io/browse/BATCH-2575?redirect=false)** and commented

Spring Batch currently depends on Jettison 1.2, which conflicts with other Spring Boot modules, like Spring Cloud Netflix. Specifically, Eureka has a dependency for Jettison 1.3.x. If an application utilizes both Eureka and Spring Batch, then XStream exceptions are generated when utilizing JobExplorer.

Jettison 1.2 is also very old. It was released in 2010.

---

**Affects:** 3.0.7

**Referenced from:** commits https://github.com/spring-projects/spring-batch/commit/4ad00166f84eb6df1e0a3bf1a3900ed70e09858d

**Backported to:** [4.0.0.M2](https://github.com/spring-projects/spring-batch/milestone/85?closed=1)

0 votes, 5 watchers

PR #4196, opened 2022-09-21
Declare dependency on jettison as optional

hpoettker:
Spring Batch depends on jettison only for `XStreamExecutionContextStringSerializer` which is deprecated and optional to use. This PR declares the dependency as optional in the published pom, which is already the case for XStream.

See also the discussion in #1028.


fmbenhassine:
Rebased and merged as 7b0e00e8a500f99a72e94e364ce8f7777af3768e. Just in time for 4.3.7 😉 . Thank you for your contribution!

PR #4199, opened 2022-09-25
Fix minor example in job.adoc

IvanAgafonov:
Minor fix example for
https://docs.spring.io/spring-batch/docs/current/reference/html/job.html#JobOperator
forgot jobLauncher


fmbenhassine:
Thank you for your contribution! Rebased and merged.

I created https://github.com/spring-projects/spring-batch/issues/4205 to include code samples from sources that are compiled and verified to make sure all snippets in the docs are correct.

PR #4203, opened 2022-10-03
Check the return value of File.delete()

marschall:
Check the return value of File.delete() and throw an exception if deleting a file fails.

This fixes a well known bug pattern

[RV: Method ignores exceptional return value (RV_RETURN_VALUE_IGNORED_BAD_PRACTICE)](https://findbugs.sourceforge.net/bugDescriptions.html#RV_RETURN_VALUE_IGNORED_BAD_PRACTICE)

> This method returns a value that is not checked. The return value should be checked since it can indicate an unusual or unexpected function execution. For example, the `File.delete()` method returns false if the file could not be successfully deleted (rather than throwing an Exception). If you don't check the result, you won't notice if the method invocation signals unexpected behavior by returning an atypical return value.


fmbenhassine:
Rebased and merged as 8c7fb5ca02a1c46d27b27b09a0d724b30c25c7a7. Thank you!

PR #422, opened 2016-06-05
remove dead code related to PagingQueryProvider#generateJumpToItemQuery

jpraet:

PR #4207, opened 2022-10-05
Remove un-used code for jump-to-item queries

hpoettker:
The PR is based on the current main branch and has the same scope as #422.


fmbenhassine:
LGTM. Rebased and merged as 220de61f752908957286f017007123a92e0ce4b6. Thank you for your contribution!

Issue #4135, opened 2022-06-17
Improve performance when polling replies in MessageChannelPartitionHandler


lcmarvin:
Currently a for each loop is used in `MessageChannelPartitionHandler` to poll the partition StepExecutions.
```
for (Iterator<StepExecution> stepExecutionIterator = split.iterator(); stepExecutionIterator.hasNext();) {
	StepExecution curStepExecution = stepExecutionIterator.next();

	if (!result.contains(curStepExecution)) {
		StepExecution partitionStepExecution = jobExplorer
            .getStepExecution(managerStepExecution.getJobExecutionId(), curStepExecution.getId());

		if (!partitionStepExecution.getStatus().isRunning()) {
			result.add(partitionStepExecution);
		}
	}
}
```
When there are lots of partition StepExecutions, the method `org.springframework.batch.core.explore.JobExplorer#getStepExecution` will cause lots of repeated sql query on database. Take a look at the implementation of this method.
```
	@Nullable
	@Override
	public StepExecution getStepExecution(@Nullable Long jobExecutionId, @Nullable Long executionId) {
		JobExecution jobExecution = jobExecutionDao.getJobExecution(jobExecutionId);
		if (jobExecution == null) {
			return null;
		}
		getJobExecutionDependencies(jobExecution);
		StepExecution stepExecution = stepExecutionDao.getStepExecution(jobExecution, executionId);
		getStepExecutionDependencies(stepExecution);
		return stepExecution;
	}
```
The JobInstance, JobExecution, JobParameters, JobExecutionContext, and StepExecutions of the JobExecution are queried again and again, which can be simplified in my option. 

So can we only query the needed StepExecution in this loop? I believe there will be performance improvement if we can simplify the query.

fmbenhassine:
This is a valid point, thank you for raising it!

The performance improvement will be addressed in #3790, which reports the same issue.

PR #4208, opened 2022-10-05
Performance improvement for step execution retrieval

hpoettker:
This PR is for versions 4.3.x and addresses #3790 and #4135.

Using `JobExplorer::getStepExecution` for multiple step executions is not ideal. Each invocation will retrieve the corresponding job execution which in turn triggers the retrieval of all step executions of the job execution. If `JobExplorer::getStepExecution` is used for all step executions, this leads to efforts and memory consumption that scale quadratically in the number of step executions.

The idea of the PR is to only call `JobExplorer::getJobExecution` and get the step executions from the returned job execution. The PR is intended to be conservative in its changes due to the late stage of the 4.3.x releases. That's why the filtering on the step execution ids are implemented although I'm not 100% sure that they are actually needed in every case.

I tested the change locally with the reproducing project provided here: https://github.com/spring-projects/spring-batch/issues/3790#issuecomment-810761639 With 250 MB of heap, the application no longer crashes and the memory profile looks much better.

For Spring Batch 5, it makes sense to forward port the change, I think. It should also be considered to deprecate `JobExplorer::getStepExecution` and provide a way to retrieve a `StepExecution` from a `JobExecution` by id instead.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/launch/JsrJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/launch/JsrJobOperator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 the original author or authors.
+ * Copyright 2013-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


vitenkov:
do you guys have any expectation when the fix be pushed to spring-batch?

fmbenhassine:
The idea of this PR is a clever trick and a nice middle ground between having the performance issue and fixing it with a single query for polling + shallow copies of step executions as mentioned in https://github.com/spring-projects/spring-batch/issues/3790#issuecomment-1439615319.

I appreciate the fact that you tried to fix the problem with 4.3.x in mind! This PR has been rebased and merged in v4.3.x and forward ported to v5 (without the JSR related changes, see 93800c6bae679fdafdced45f1c22b2c150e33ed6). Thank you for your contribution!

Issue #3699, opened 2020-04-20
Multiple Job unit testing with @SpringBatchTest


tiparega:
Hi all,

From documentation, it seems impossible to unit test a Spring Batch with more than one job, problem is JobLauncherTestUtils autowiring the Job. This can be resolved as in #1237 but, as docs state to use @SpringBatchTest (#889), a default JobLauncherTestUtils bean is created in BatchTestContextCustomizer so, even if you don't use any JobLauncherTestUtils at all, it tries to create one and fails on no single Job bean.

Not sure if this has a simple resolution, may be deleting autowired Job in JobLauncherTestUtils, or allowing some kind of @SpringBatchTest customization, but I think that it could be, at least, mentioned in unit testing docs, so to not use @SpringBatchTest annotation if you have more than one Job bean.

mminella:
The utility in question `JobLauncherTestUtils` is for testing **a** job.  That implies that the context it is wired into only has one (or is marked via `@Primary` for Spring's autowiring functionality).  I'd be interested in hearing the use case you are exploring where a test of a job ends up with more than one job that cannot be marked as `@Primary`.

jblayneyXpanxion:
@mminella 

So I might chime in here. I'm pretty good at spring and setting up integration tests, and setting up spring batch tests has been quite the nightmare. 

I have a spring boot web application that is also enabled for spring batch.

A few issues I've run into:
- If you add `@SpringBatchTest` to any test that uses a context configuration that uses your main app's main class.. eg: Application.java that may have multiple jobs in the component scan (common), then this explodes. There is no perfect way to exclude certain sets of jobs from the main app's context for certain tests so there is only a single job bean. Overriding the bean or defining the jobLauncherTestUtils explicitly doesn't work either, as apparently it still tries to autowire just one job and explodes before the overriding can kick in.
[...]

PR #4218, opened 2022-10-23
Improve `@SpringBatchTest` to autowire the job under test in `JobLauncherTestUtils` if it is unique 

### LINKED ISSUE #1237: Do not autowire the Job inside JobLauncherTestUtils
**[Petar Tahchiev](https://jira.spring.io/secure/ViewProfile.jspa?name=ptahchiev)** opened **[BATCH-2366](https://jira.spring.io/browse/BATCH-2366?redirect=false)** and commented

Hello,

I have several jobs defined. Then I want to test them so I create a bean of type `JobLauncherTestUtils` inside my config:

```
@Bean(name = "ncsvjobLauncherTestUtils")
public JobLauncherTestUtils defaultJobLauncherTestUtils() {
   JobLauncherTestUtils result =  new JobLauncherTestUtils();
   result.setJob(context.getBean("ncsvImportJob", Job.class));
   return result;
}
```

However when I run my tests I get this exception:

```
ption is org.springframework.beans.factory.BeanCreationException: Could not autowire method: public void org.springframework.batch.test.JobLauncherTestUtils.setJob(org.springframework.batch.core.Job); nested exception is org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type [org.springframework.batch.core.Job] is defined: expected single matching bean but found 3: ncsvImportJob,syncEntityJob,syncCatalogJob
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor.postProcessPropertyValues(AutowiredAnnotationBeanPostProcessor.java:334)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.populateBean(AbstractAutowireCapableBeanFactory.java:1202)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:537)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:476)
```

This is because the `JobLauncherTestUtils` the `setJob` method has the `@Autowired` annotation so it will try to autowire the job, but as I said I have several.

---

2 votes, 3 watchers

hpoettker:
To resolve #1237, the feature to always autowire a job into `JobLauncherTestUtils` was removed as the autowiring leads to issues if there is no unique candidate in the test context.

This PR adds the flag `autowireJob` to the annotation `@SpringBatchTest`. If set to `true`, the developer opts in to the old behavior.

Using narrowly defined test contexts when writing unit tests for jobs is a good practice in my opinion. Similar to using test slices for a Spring Boot application, it can have a strong influence on the run-time of build pipelines and thus tighten feedback cycles during development.

The proposed change is intended to be an incentive for test contexts that contain only the job to be tested. Additionally, it simplifies migration to Spring Batch 5 a bit for developers that relied on the old behavior.

### DIFF
--- a/spring-batch-samples/src/test/java/org/springframework/batch/sample/AMQPJobFunctionalTests.java
+++ b/spring-batch-samples/src/test/java/org/springframework/batch/sample/AMQPJobFunctionalTests.java
@@ -17,7 +17,6 @@ package org.springframework.batch.sample;
 
 import org.junit.jupiter.api.Test;
 
-import org.springframework.batch.core.Job;
 import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.test.JobLauncherTestUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -50,9 +49,8 @@ class AMQPJobFunctionalTests {
 	private JobExplorer jobExplorer;
 
 	@Test
-	void testLaunchJob(@Autowired Job job) throws Exception {
+	void testLaunchJob() throws Exception {
 		// given
-		this.jobLauncherTestUtils.setJob(job);
 		this.jobLauncherTestUtils.launchJob();
 
 		// when
--- a/spring-batch-samples/src/test/java/org/springframework/batch/sample/BeanWrapperMapperSampleJobFunctionalTests.java
+++ b/spring-batch-samples/src/test/java/org/springframework/batch/sample/BeanWrapperMapperSampleJobFunctionalTests.java
@@ -18,7 +18,6 @@ package org.springframework.batch.sample;
 
 import org.junit.jupiter.api.Test;
 
-import org.springframework.batch.core.Job;
 import org.springframework.batch.test.JobLauncherTestUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
@@ -31,10 +30,7 @@ class BeanWrapperMapperSampleJobFunctionalTests {
 	private JobLauncherTestUtils jobLauncherTestUtils;
 
 	@Test
-	void testJobLaunch(@Autowired Job job) throws Exception {
-		// given
-		this.jobLauncherTestUtils.setJob(job);
-
+	void testJobLaunch() throws Exception {
 		// when
 		this.jobLauncherTestUtils.launchJob();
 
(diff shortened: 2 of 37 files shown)


fmbenhassine:
> To resolve https://github.com/spring-projects/spring-batch/issues/1237, the feature to always autowire a job into JobLauncherTestUtils was removed as the autowiring leads to issues if there is no unique candidate in the test context.

We had an internal discussion about this, and one of the ideas that came out to address the case where multiple jobs are defined in the test context was to add an attribute with the name of the job to autowire in `JobLauncherTestUtils`. We decided to not implement this feature as such an information is statically defined in an annotation (and would also "encourage" a bad practice that we want avoid in the first place).

> Using narrowly defined test contexts when writing unit tests for jobs is a good practice in my opinion.

I couldn't agree more! See my comment here: https://github.com/spring-projects/spring-batch/issues/3699#issuecomment-702966663

> The proposed change is intended to be an incentive for test contexts that contain only the job to be tested. Additionally, it simplifies migration to Spring Batch 5 a bit for developers that relied on the old behavior.

For the case where a single job is defined, I think there is a room for improvement with a sensible default. What you suggest is a good idea IMO. However, do we really need a flag for that? What about autowiring the job in `JobLauncherTestUtils` if there is only one? This removes the need for a flag and is nearly the same behaviour as in v4. wdyt?

hpoettker:
Thanks for the feedback! I've implemented the proposal to only inject a job into `JobLauncherTestUtils` if it's unique.

For me personally this solution would work very well. I'm wondering however if it's not too "magical" for developers who are not aware of the underlying mechanism. To them it might seem that the injection process is flaky as "sometimes" a job is injected and "sometimes" not.

The original commit with the flag has the advantage that it's clearer how it works: If the flag is not set, then no job will be injected. Otherwise, a job will be injected or the test fails while trying to do so.

fmbenhassine:
Thank you for your feedback and for the quick update!

> For me personally this solution would work very well.

I believe this is because you are applying the best practice of isolating tests for each job. And this would work very well too for anyone who is applying this best practice. My thinking with the suggested default is as follows: It is unfair to make users who apply the best practice (of isolating tests) to pay the price of autowiring the single job in test utilities.

> I'm wondering however if it's not too "magical" for developers who are not aware of the underlying mechanism. To them it might seem that the injection process is flaky as "sometimes" a job is injected and "sometimes" not.

There is no magic. Magical is only magical if it is not documented. I will refer anyone who would complain about this being magical to this [tweet](https://twitter.com/starbuxman/status/708579187087511552) by Josh 😄 . That said, I see no note about this new default in the Javadoc of `SpringBatchTest` in this PR. I will take care of that on merge.

Thanks again for all your contributions! REALLY appreciated 👍

fmbenhassine:
Rebased and merged as 3bb7ce532b14151bb14b4823c5cad87a8265637d. Revised in 1527593a3676483fa6aaa7808c1453500dec7964. Thank you for your contribution!

PR #4220, opened 2022-10-25
Change StepBuilerHelper#enhance parameter type to AbstractStep

acktsap:
This commit moves `StepBuilerHelper#enhance` method to `AbstractStep`. It removes unnecessary casting to `AbstractStep`.

Currently, changing argument type to `AbstractStep` is enough. But until 4.3.x, there were [another casting code](https://github.com/spring-projects/spring-batch/blob/84b1e1f7027e4dfa480a87f3f092633d64a1e559/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/StepBuilderHelper.java#L160-L163) (It’s been removed by https://github.com/spring-projects/spring-batch/commit/7c8fb172a7cfb8d61d54f530c564b836d83d0a21). But there may be similar cases in future. 

By moving enhance method to AbstractStep, we can use like followings for those cases.

```java
// in TaskletStep
@Override
public void enhance(StepProperties properties) {
    super.enhance(properties);
    this.transactionManager = properties.getTransactionManager();
}
```

This also can be applied to [JobBuilderHelper](https://github.com/spring-projects/spring-batch/blob/5da133ac3669343b05187aec7469e07ff37f5d36/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilderHelper.java#L158-L186).

I also extracted `StepProperties` to prevent dependency cycle (org.springframework.batch.core.step.builder <-> org.springframework.batch.core.step)

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
@@ -21,12 +21,12 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 
 import org.springframework.batch.core.ChunkListener;
-import org.springframework.batch.core.Step;
 import org.springframework.batch.core.StepExecutionListener;
 import org.springframework.batch.core.annotation.AfterChunk;
 import org.springframework.batch.core.annotation.AfterChunkError;
 import org.springframework.batch.core.annotation.BeforeChunk;
 import org.springframework.batch.core.listener.StepListenerFactoryBean;
+import org.springframework.batch.core.step.AbstractStep;
 import org.springframework.batch.core.step.tasklet.Tasklet;
 import org.springframework.batch.core.step.tasklet.TaskletStep;
 import org.springframework.batch.item.ItemStream;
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Thank you for the PR. Could you please rebase it in on the latest `main`? Please note that we added an `ObservationRegistry` and `MeterRegistry` properties in `CommonStepProperties`, please make sure to include them in your rebase.

Any reason you renamed `CommonStepProperties` to `StepProperties`? I think we should keep the name `CommonStepProperties` to be consistent with the naming of `CommonJobProperties` for jobs.

> This also can be applied to [JobBuilderHelper](https://github.com/spring-projects/spring-batch/blob/5da133ac3669343b05187aec7469e07ff37f5d36/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilderHelper.java#L158-L186).

Please open a separate PR for this enhancement. Thank you upfront.

acktsap:
Rebased and add `ObservationRegistry`, `MeterRegistry`. Renamed to `CommonStepProperties` back.
I'll make a pr for `JobBuilderHelper` soon.

acktsap:
> That method is actually configuring/enhancing the step with some properties, and I believe this is the role of the builder, not the step itself. The step can provide setters for its properties, but not a public method enhance (currently, that method is protected in the builder)

Make sense. I thought changing parameter type to `AbstractStep` at first. Then can i make the change (just change the parameter type) in this pr and https://github.com/spring-projects/spring-batch/pull/4231?

fmbenhassine:
> Then can i make the change (just change the parameter type) in this pr and https://github.com/spring-projects/spring-batch/pull/4231?

Yes please. It should be good to merge if all tests pass. Thank you upfront!

acktsap:
Done

fmbenhassine:
Rebased and merged as 3385a19cb4ebc518652a1aa05d0cd306487feb47. Thank you for your contribution!

Issue #4230, opened 2022-11-11
Calling `JobExplorer` outside of a transaction logs warnings about the isolation Level not being applied


kzander91:
Not sure if this is a bug or just a documentation issue.

**Bug description**
As of Batch 5, calling any `JobExplorer` method without an active transaction produces the following warning:
```
2022-11-11T23:33:19.650Z  WARN o.s.jdbc.support.JdbcTransactionManager  : Custom isolation level specified but no actual transaction initiated; isolation level will effectively be ignored: PROPAGATION_SUPPORTS,ISOLATION_READ_COMMITTED
```

The cause seems to be the changes made with #4195. As of these changes, `JobExplorer` is transactional with `PROPAGATION_SUPPORTS,ISOLATION_READ_COMMITTED`, causing the warning being logged if called outside of a transaction.

**Environment**
Spring Batch 5.0.0-RC2

**Steps to reproduce**
1. Call any `JobExplorer` method without an active transaction.

**Expected behavior**
The call works without issuing warnings.

**Minimal Complete Reproducible example**
[demo1.zip](https://github.com/spring-projects/spring-batch/files/9993667/demo1.zip)
1. Unzip
1. Run `mvn spring-boot:run`

The app calls `JobExplorer#getJobInstanceCount()` from a `CommandLineRunner`, which issues the warning mentioned above.

hpoettker:
As far as I can tell, the warning is harmless, and can be ignored. But I think it should still be dealt with before Spring Batch 5 goes GA.

It really doesn't take much to trigger it:
```
@SpringBootApplication
public class DemoApplication {

  public static void main(String[] args) {
    SpringApplication.run(DemoApplication.class, args);
  }

  @Bean
  Job job(JobRepository jobRepository, Step step) {
    return new JobBuilder("job", jobRepository)
        .start(step)
        .incrementer(new RunIdIncrementer())
        .build();
  }

  @Bean
  Step step(JobRepository jobRepository, PlatformTransactionManager transactionManager) {
    return new StepBuilder("step", jobRepository)
        .tasklet((contribution, chunkContext) -> RepeatStatus.FINISHED, transactionManager)
        .build();
  }

}
```

Here, the `RunIdIncrementer` leads to a call of `JobParametersBuilder::getNextJobParameters`, which calls `JobExplorer::getLastJobInstance`.

fmbenhassine:
[...]

PR #4236, opened 2022-11-17
Fix `JdbcJobExecutionDaoTests`

hpoettker:
The commit bbf35b79f7f830e50545de2bbbf191382e3e7dd4 broke the build, and the test `JdbcJobExecutionDaoTests::testFindRunningExecutions` is currently failing.

It's a minor thing in the test, and this PR fixes it.

Issue #4195, opened 2022-09-21
Add support to customize transaction attributes


fmbenhassine:
While the `JobRepositoryFactoryBean` provides a way to customize the transaction manager to use in the transactional proxy around the `JobRepositoy` it creates, it does not however allow the customization of transaction attributes.

The goal of this feature request is to make the transaction attributes configurable as well, without changing the current default attributes.

This should also be applied to the `JobExplorerFactoryBean` (#1307) and `JobOperatorFactoryBean` (#1078).

PR #4237, opened 2022-11-17
Do not proxy the `JobExplorer` by default

### LINKED ISSUE #1307: Convenient transaction support in JobExplorerFactoryBean [BATCH-2295]
**[Erwin Vervaet](https://jira.spring.io/secure/ViewProfile.jspa?name=klr8)** opened **[BATCH-2295](https://jira.spring.io/browse/BATCH-2295?redirect=false)** and commented

The JobRepositoryFactoryBean will automatically wrap a transactional proxy around the created JobRepository instance to enforce its specialized transactional requirements (i.e. REQUIRES_NEW for create methods, ...).

It would be **convenient** to have a similar feature in the JobExplorerFactoryBean: it also needs a transaction to access the job data in a database. In this case it would be optional since the JobExplorer is read-only and has no special transactional requirements like the JobRepository, or might be used inside an existing transactional scope.

---

**Affects:** 2.2.7, 3.0.1

hpoettker:
To resolve #1307 and #4195, the possibility to wrap the `SimpleJobExplorer` in a transactional proxy was added. In addition, a new behavior was introduced to use such a proxy also by default with transaction propagation `SUPPORTS` and isolation level `READ_COMMITTED`.

This default behavior causes the issue #4230, i.e. it produces warnings that are harmless but might confuse many.

This PR proposes to return the **default** behavior to the previous behavior, i.e. no transactional proxy for the `JobExplorer`. As the `JobExplorer` is read-only, the difference between no transaction and the combination of `SUPPORTS` and `READ_COMMITTED` seems negligible to me for most cases. And the possibility for the user to add a transactional proxy still exists if required.

The commit for this PR does not build successfully. But this is independent of the PR and dealt with in #4236.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/explore/support/AbstractJobExplorerFactoryBean.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/explore/support/AbstractJobExplorerFactoryBean.java
@@ -16,8 +16,6 @@
 
 package org.springframework.batch.core.explore.support;
 
-import java.util.Properties;
-
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.core.repository.dao.ExecutionContextDao;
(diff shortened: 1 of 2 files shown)


hpoettker:
I've rebased after #4236 has been merged. The build of the branch is now successful.

fmbenhassine:
Thank you for your PR. The suggested change introduces an inconsistency between how the `JobRepository` and `JobOperator` are configured compared to the `JobExplorer`. With this change, one would need to provide a transaction attributes instance to create a transactional proxy around the `JobExplorer`, while it is not the case with the `JobRepository` and `JobOperator`. So for consistency, I think the provided `JobExplorer` should remain transactional by default (so that users are not required to make it transactional), but we need to review the defaults.

I think the issue is that the current default isolation level of `READ_COMMITTED` is not correct. It should rather be `DEFAULT`. With this new default, the warning is not logged.

Do you agree? If yes, I can take care of the change to fix #4230.

hpoettker:
Thanks for the detailed feedback!

In fact, I think that it's most important that these warnings are not logged in standard use cases. If isolation level `DEFAULT` achieves this, then it's a good solution. Please feel free to decline this PR.

fmbenhassine:
Closing this in favour of changing the default isolation level to `DEFAULT`, which fixes the issue. Thank you for your contribution anyway!

Issue #4263, opened 2022-12-19
Incorrect notNull assertion on parameter type in JobParameter


parksjin01:
In JobParameter constructor there are duplicated assertion which check `value` parameter is not null.
https://github.com/spring-projects/spring-batch/blob/d3bb0ded413f947a396735e0ad9d5edc851b7490/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameter.java#L50

```
public JobParameter(@NonNull T value, @NonNull Class<T> type, boolean identifying) {
	Assert.notNull(value, "value must not be null");
	Assert.notNull(value, "type must not be null");
	this.value = value;
	this.type = type;
	this.identifying = identifying;
}
```

With annotation of construction parameter and message in second `Assert.notNull`,
I guess second assertion might be intended to check that type parameter value is not null.

If my guess is correct, then how about change second assertion to check `type` parameter instead of `value`

PR #4266, opened 2022-12-25
Add notNull assertion in JobParameter to ensure type is not null

drow724:
Add notNull assertion in JobParameter to ensure type is not null #4263


fmbenhassine:
Rebased and merged as 6bea505f1505e9f69a46e60bc7464b88272d463d. Thank you for your contribution!

PR #4269, opened 2023-01-03
Updated architecture adoc so that it batch-asciidoc variable is set

### LINKED ISSUE #4268: Architecture html docs are missing diagrams
The diagrams in the html (index.html) are not displayed.   However they do show up in the html-single (index-single.hmtl) and PDF docs.

cppwfs:
* Updated all docs that had `batch-asciidoc` declared to use attributes.adoc instead.   This way we can set the variable in one location and all adocs can take advantage of it.
* Also made sure that `batch-asciidoc` was set to use a relative path vs. the link in the UI.  This is because the PDF can't use a URL that contains the `-` character.

Resolves #4268


fmbenhassine:
Rebased and merged as e4de63c2ebfaef4d5ae2540b535b2818c110eb6d. Thank you for the fix!

PR #4285, opened 2023-01-17
Add default component name in constructor of AvroItemReader

drow724:
```java
public class AvroItemReaderBuilder<T> {

	private String name = AvroItemReader.class.getSimpleName();
```

AvroItemReaderBuilder's name field has defalut value (AvroItemReader.class.getSimpleName)

``` java
if (this.saveState) {
	Assert.state(StringUtils.hasText(this.name), "A name is required when saveState is set to true.");
}
```

so that assert code is never throw error because name field has default value

and i'm adding code that initializing name field to AvroItemReader.class.getSimpleName

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/builder/AvroItemReaderBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/builder/AvroItemReaderBuilder.java
@@ -22,7 +22,6 @@ import org.springframework.batch.item.avro.AvroItemReader;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.core.io.Resource;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 
 /**
  * A builder implementation for the {@link AvroItemReader}.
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/avro/builder/AvroItemReaderBuilder.java
```diff
@@ -170,10 +170,6 @@ public AvroItemReader<T> build() {
 
 		avroItemReader.setSaveState(this.saveState);
 
-		if (this.saveState) {
-			Assert.state(StringUtils.hasText(this.name), "A name is required when saveState is set to true.");
-		}
-
```
fmbenhassine, line 176: > remove useless assert state

This assertion is not useless. Here is a case where the name could end up being null or empty even with a default name in the constructor:

```java
AvroItemReader avroItemReader = new AvroItemReaderBuilder()
		.name(getNameFromPropertiesFile())
		// set other properties
		.build();
```

In this example, if `getNameFromPropertiesFile()` returns null or an empty string, the reader will be created in an inconsistent state. That's why it is better to fail fast at the builder level.

fmbenhassine:
Rebased, squashed and merged as 89e76164af832268ab5bb76e0b9b653b495ed274. I just reverted the assertion removal as I believe it is needed as explained in the comment above. Thank you for your contribution!

PR #4323, opened 2023-03-10
Add cursor-based ItemReader for MongoDB

juchanei:
The MongoItemReader currently uses pagination with MongoDB's skip operation, which can cause unnecessary document scans when requesting next pages.

> The `skip()` method requires the server to scan from the beginning of the input results set before beginning to return results. As the offset increases, `skip()` will become slower.
> https://www.mongodb.com/docs/manual/reference/method/cursor.skip/#using-skip--

Therefore, I have implemented a new cursor-based MongoDB `ItemReader` that accesses the next documents without skipping.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoCursorItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoCursorItemReader.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.batch.item.data;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import org.bson.Document;
+import org.bson.codecs.DecoderContext;
+
+import org.springframework.batch.item.support.AbstractItemCountingItemStreamItemReader;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.mongodb.core.MongoOperations;
+import org.springframework.data.mongodb.core.query.BasicQuery;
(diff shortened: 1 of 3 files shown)


MinJunKweon:
related #3824

fmbenhassine:
Awesome! Thank you for contributing this cursor-based implementation! I will plan this feature for Spring Batch 5.1.

Just curious, have you run a benchmark to compare the performance of this reader with the paging one?

juchanei:
> Just curious, have you run a benchmark to compare the performance of this reader with the paging one?

@fmbenhassine 

As you know, pagination-based item readers perform poorly in terms of performance due to the amount of documents that need to be skipped. This amount is related to the total number of documents and the batch size, making it difficult to quantify the performance improvement of cursor-based item readers in numbers.

In my case, when benchmarking with a batch size of 10k on 0-3.5 million documents, the elapsed time was reduced by about 20-40%. While different results may occur depending on the test conditions, I am attaching the scatter chart results from my test.

image

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoCursorItemReader.java
```diff
@@ -0,0 +1,297 @@
...
+    public void setHint(String hint) {
+        this.hint = hint;
+    }
+
+    /**
+     * The size of batches to use when iterating over results.
+     * @param batchSize size the batch size to apply to the cursor
+     */
+    public void setBatchSize(Integer batchSize) {
+        this.batchSize = batchSize;
```
hmhuan, line 164: Nitpicking: Should we take care of `batchSize` when it is 0 or negative?
As I checked the spec of jpa-mongo

* Use {@literal 0 (zero)} for no limit. A <strong>negative limit</strong> closes the cursor after returning a single
* batch indicating to the server that the client will not ask for a subsequent one.

ref: https://github.com/spring-projects/spring-data-mongodb/blob/main/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Query.java#L539-L546

juchanei: @hmhuan I'm sorry for the delay in responding.
Users would expect that this code is implemented using spring-data-mongodb, given the explicit use of `MongoOperations` as an argument in `MongoCursorItemReader`. Therefore, if `setBatchSize` handles 0 or negative values, I think it could reduce predictability.

hmhuan (approved):

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoCursorItemReader.java
fmbenhassine: The type of the parameter should be the same as expected by the `Query` API. The setter should add the javadoc tag `@see Query#cursorBatchSize(int)` so the user knows that the value will be set on that field, and therefore knows the (default) values to use. I will take care of this change when merging the PR.

fmbenhassine:
Excellent! Thank you for sharing those benchmarks. Great to see up to 40% improvement!

Rebased and merged as b1039004bf364bf78a64495183e9dbc9b8022e34. Refined in d84057c2b1f077c4d48b41305f00178f4abbf138.

PR #4331, opened 2023-03-21
Change visibility of properties in JdbcPagingItemReaderBuilder

Ditscheridou:
There are cases, when you want to extend the JdbcPagingItemReader by adding some custom fields and behaviours, but want to stay in the normal way of creating this class via a some kind of derived JdbcPagingItemReaderBuilder class. If you want to do it, you currently have to reimplement the whole class just for adding some fields because there is no non-ugly way to access the fields. My proposal here is to make them protected, so you can override the default build() method behaviour in your derived class and still have access to the properties.


fmbenhassine:
LGTM. Rebased and merged as bb337c6b2fad4f76b33525863d341e44adbf076a. Thank you for your contribution!

Issue #3829, opened 2021-01-11
Access to update counts in JdbcBatchItemWriter


Jeff-Walker:
**Expected Behavior**

I'd like to have the ability to access the `updateCounts` array in the `JdbcBatchItemWriter<>`.
This could be easily done by extracting the "assert updates" portion of the `write()` method into a `protected` method.

```java
protected void postProcess(int[] updateCounts) throws Exception {
  if (assertUpdates) {
    for (int i = 0; i < updateCounts.length; i++) {
      int value = updateCounts[i];
         if (value == 0) {
           throw new EmptyResultDataAccessException("Item " + i + " of " + updateCounts.length
               + " did not update any rows: [" + items.get(i) + "]", 1);
         }
       }
  }
}
```

Then, my class, using a class that tracks the facilities (my business object) that were updated and inserted could look like this.

```java
@Override
protected void postProcess(int[] updateCounts) throws Exception {
  facilityTracker.addUpdated(Arrays.stream().sum());
  super.postProcess(updateCounts);
}
```

**Current Behavior**

Currently, the `updateCounts` array is a local variable in the `write()` method, and I'd have to totally reimplement the entire method to have access to it. Having a parallel implementation that could become out of date in subsequent releases of Spring Batch.

**Context**

I realize that this might be niche ask, but my batch program updates a table of "facilities" based on data received from another system. I have to keep track of how may were updated and inserted, etc., but right now I can only estimate based on the result from my `ItemProcessor`. My only other option is to totally reimplement the `write()` method, including the complex JDBC code.

**Note**

I can provide a PR soon for this if it has any chance of being committed.

fmbenhassine:
@Jeff-Walker  Thank you for opening this feature request. The update count is indeed inaccessible as of v5.0.2. We can make it accessible if this helps extending the class with such features and without having to reimplement the entire `write` method.

However, the suggested change might break the `assertUpdate` feature. Indeed, if someone extends the class, overrides `postProcess` and forgets to call `super.postProcess(updateCounts);`, the assertion is not applied if it is set to `true`. Moreover, the suggested method signature requires passing the chunk of items as parameter to be able to raise an `EmptyResultDataAccessException` and specify which item did not update any row, something like the following:

```java
protected void postProcess(int[] updateCounts, final Chunk<? extends T> chunk) throws Exception {
  if (assertUpdates) {
    for (int i = 0; i < updateCounts.length; i++) {
      int value = updateCounts[i];
[...]

PR #4339, opened 2023-03-28
Allow access to update counts in JdbcBatchItemWriter

robertmcnees:
Fix for #3829


fmbenhassine:
Thank you for contributing this PR! I think the change set here was based on the initial suggestion in #3829. However, as explained in https://github.com/spring-projects/spring-batch/issues/3829#issuecomment-1640401795, I believe we need to minimize the scope of the extension point and call it outside the `if (assertUpdates) { ... }` block. Do you agree?

If you agree, could you please update the PR accordingly? Thank you upfront.

fmbenhassine:
LGTM 👍  Rebased, squashed and merged as 2c979743667a7754e5c43ffbe4d42ed036db7f53. Thank you for your contribution!

PR #4343, opened 2023-04-03
Add option to provide custom JobKeyGenerator in JdbcJobInstanceDao

### LINKED ISSUE #3926: Add option to provide a custom JobKeyGenerator in JdbcJobInstanceDao
The `JdbcJobInstanceDao` uses a default `JobKeyGenerator` but provides no way to supply a custom implementation. Even though it is unlikely to implement a custom `JobKeyGenerator`, there should be a way to provide a custom one.

robertmcnees:
Fixes #3926


fmbenhassine:
Thank you for this PR, @robertmcnees ! LGTM.

One thing though is that we need to update the `JobExplorerFactoryBean` as well to set the custom key generator on the `JdbcJobInstanceDao` (similar to what is done for the `JobRepositoryFactoryBean`). We need to use the same key generator in both the `JobRepository` and `JobExplorer`.

With that in place, I think the PR would be good to merge. Let me know if you need help on addressing the update.

fmbenhassine:
LGTM. Thank you for the updates! Rebased, squashed and merged as e36a44788d25d14a2faefc5aea5ecd9abeb35c39. Thank you for your contribution!

PR #4351, opened 2023-04-13
Add Operation Mode in MongoItemWriter

### LINKED ISSUE #4149: Add support for bulk inserts in MongoItemWriter
Currently there is support for the upsert and remove operations, I would like to suggest to add support for insert operation.

It would be nice to define an enum based set of operations (INSERT, UPSERT, REMOVE).

UPSERT will be the default operation (same behaviour as it is now) and the 'setDelete' method can be modified to set the REMOVE operation.

I am happy to submit a proposal in case the idea is accepted.

cgrings:
Fixes #4149


fmbenhassine:
Great! Thank you. Planning that for 5.1.

PR #4355, opened 2023-04-17
Fix for the XML/Java switch regression

Buzzardo:
and some other problems.

Some problems cannot be fixed, because we use the same file
for both the single-file output and the multi-file output. That
makes some of the anchors impossible to find in the multi-file
output. I'll look into a solution for that.


fmbenhassine:
Great! Thank you for restoring the toggle! Rebased and merged as 851bf706d84780e4cce9e034bd7cd0cc3c63d216.

I also created an improvement request related to that, see #4357.

Issue #4360, opened 2023-04-27
Disable Docker-based tests when Docker is not available


fmbenhassine:
Currently, a full build runs Docker-based (TestContainers) integration tests no matter if Docker is available on the platform or not. This leads to a build failure when Docker is not available. This behaviour can be improved by disabling Docker-based tests when Docker is not available.

The TestContainers library provides the attribute `@TestContainers(disabledWithoutDocker = true)` for that case. The goal of this issue is to add that attribute to all tests annotated with `@TestContainers`.

Contributions are welcome!

CNJingo:
I'd like to try it.

sukanth:
@fmbenhassine  can you assign this to me ? i can take this one

sukanth:
@fmbenhassine  Thank you for assigning it to me . here is the PR https://github.com/spring-projects/spring-batch/pull/4367

PR #4361, opened 2023-04-28
Disable Docker-based tests when Docker is not available

CNJingo:
fix for #4360


CNJingo:
Didn't recognize the label 'first-timers-only'. So this PR is should be closed.

PR #4364, opened 2023-05-02
Fix to get parameters correctly when multiple job executions exist

### LINKED ISSUE #4246: SimpleJobExplorer getJobExecutions returns job parameters of first job execution
**Bug description**
The following method `List<JobExecution> getJobExecutions(JobInstance jobInstance)` in the `SimpleJobExplorer` class returns the wrong job parameters when a job instance is started multiple times with different job parameters (identifying job parameters are the same but the non-identifying parameters were  changed)

**Environment**
Versions:
* Spring Batch: 4.3.7
* Java: 17.0.3
* Database: Postgres

**Steps to reproduce**
* Define a job with identifying parameters and non-identifying parameters
* Launch a job and stop it
* Restart the job with the same identifying parameters but changed non-identifying parameters
* Fetch the job executions with `List<JobExecution> getJobExecutions(JobInstance jobInstance)`
* Returned job executions will have the same job parameters even though the second execution has changed non-identifying parameters

**Expected behavior**
Either document this behavior or change the implementation to fetch the job parameters of each job execution.

**Implementation hint**

I think this behavior is because the `JdbcJobExecutionDao` only fetches the job parameters once for a job instance and then reuses them for each job execution.
`JdbcJobExecutionDao`
```java
public List<JobExecution> findJobExecutions(final JobInstance job) {

Assert.notNull(job, "Job cannot be null.");
Assert.notNull(job.getId(), "Job Id cannot be null.");

return getJdbcTemplate().query(getQuery(FIND_JOB_EXECUTIONS), new JobExecutionRowMapper(job), job.getId());
}
```

A fixed version would fetch the job parameters for each execution because each job execution has its own job parameters, see `batch_job_execution_params` table.
This would impact the performance of the method, because more data would need to be fetched.

CNJingo:
fixes #4246

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobExecutionDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobExecutionDao.java
@@ -71,6 +71,7 @@ import org.springframework.util.Assert;
  * @author Mahmoud Ben Hassine
  * @author Dimitrios Liapis
  * @author Philippe Marschall
+ * @author Jinwoo Bae
  */
 public class JdbcJobExecutionDao extends AbstractJdbcBatchMetadataDao implements JobExecutionDao, InitializingBean {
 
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Great! LGTM. Rebased and merged as e329b61d18a44cd12f2e5ed2555008889db25d06. I also added an integration test for the case, see 7f304a1a7904bfa9d7ac289e9c9371a50abb4eeb.

PR #4366, opened 2023-05-03
Upgrade to Mockito 5

hpoettker:
This PR upgrades Mockito to version 5.3.1.

I refactored the test that previously stalled. The creation of the mocks seems to interfere with the proxying of the configuration class as the problem can also be solved by setting `proxyBeanMethods = false`.

I also replaced some of the `ArgumentCaptor`s with `assertArg` that is available since Mockito 5.3.


fmbenhassine:
Thank you for looking into this! LGTM 👍  Rebased and merged as 4faa842bb1fb7a341e6dfedd3b31b7b885a3161b.

Issue #4373, opened 2023-05-13
Update SQLServer DDL script to drop sequences instead of tables


pranavjandu:
**Bug description**
We use sequences instead of tables in SQL Server. Example drop script should be updated to drop sequence instead of table.

PR #4374, opened 2023-05-13
GH-4373 Update SQLServer DDL script to drop sequences instead of tables

pranavjandu:
Updating SQL server example drop script to drop sequence instead of table.


fmbenhassine:
Good catch! LGTM 👍  Rebased and merged as 384b8a095ccf67a30c995ce849a5b8cf7c4af575. Thank you for your contribution!

PR #4376, opened 2023-05-16
Add option to clear the persistence context in jpaItemWriter

### LINKED ISSUE #804: Add option to clear the persistence context in JpaItemWriter
**[Mahmoud Ben Hassine](https://jira.spring.io/secure/ViewProfile.jspa?name=mbenhassine)** opened **[BATCH-2797](https://jira.spring.io/browse/BATCH-2797?redirect=false)** and commented

The `HibernateItemWriter` provides a parameter named `clearSession` to clear the session after writing items. However, the `JpaItemWriter` does not provide such an option, which may lead to some OOM errors (see SO thread in reference URL).

For the record, both `JpaItemWriter` and `HibernateItemWriter` used to clear the persistent context but this has been removed in [BATCH-1635](https://jira.spring.io/browse/BATCH-1635). However, this has been re-added and made configurable in the `HibernateItemWriter` in [BATCH-1759](https://jira.spring.io/browse/BATCH-1759) through the `clearSession` parameter but not in the `JpaItemWriter`.

For consistency with `HibernateItemWriter`, the `JpaItemWriter` should provide a parameter called `clearPersistenceContext` to clear the persistence context after writing items.

---

**Reference URL:** https://stackoverflow.com/questions/54746793/spring-batch-memory-leak-csv-to-database-using-jpaitemwriter

CNJingo:
Resolves #804

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaItemWriter.java
@@ -54,7 +54,7 @@ public class JpaItemWriter<T> implements ItemWriter<T>, InitializingBean {
 
 	private boolean usePersist = false;
 
-	private boolean clearSession = true;
+	private boolean clearEntityManager = true;
 
 	/**
 	 * Set the EntityManager to be used internally.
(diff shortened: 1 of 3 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/database/JpaItemWriter.java
```diff
@@ -91,6 +103,9 @@ public void write(Chunk<? extends T> items) {
 		}
 		doWrite(entityManager, items);
 		entityManager.flush();
+		if (Boolean.TRUE.equals(clearSession)) {
```
mdeinum, line 106: Why not `if (clearSession)`? It is a primitive boolean, so no need for a more complex if statement IMHO.

CNJingo: Since the clearSession variable can't be null, I think it would be better to change it as you suggest. Thanks for the review.

fmbenhassine:
LGTM! Rebased and merged as 34eda1fb0d31841fa13ce1a8c70d0657420bef69. I just renamed the variable `clearEntityManager` to `clearPersistenceContext` to be consistent with the Javadoc of `EntityManager#clear` (we do actually clear the persistence context, not the entity manager itself), see 6fa4e453560543682f4fa07c6b0ccdf747902593.

Thank you for your contribution 👍

Issue #4388, opened 2023-05-28
ItemReaderBuilder<> targetType requirement not enforced


casimk:
Please do a quick search on Github issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
ItemReaderBuilder<Type> does not set targetType and also there is no error if targetType is not set explicitly leading to an exception being thrown in reading the line but no error at the ItemReaderBuilder.  

```
    @Bean(name = "itemReader")
    public ItemReader<Establishment> itemReader() {
        return new FlatFileItemReaderBuilder<Establishment>()
                .name("itemReader")
                .linesToSkip(1)
//                .targetType(Establishment.class)
                .resource(new ClassPathResource(FILE_NAME))
                .delimited()
                .names(FIELD_NAMES)
                .fieldSetMapper(new BeanWrapperFieldSetMapper<Establishment>() {
                    {
                        setDistanceLimit(0);
                        setStrict(false);
                    }
                }).build();
    }
```

**Environment**
Spring Batch Core 5.02
Java 17

**Steps to reproduce**
Commented line in snippet

**Expected behavior**
The builder should validate and/or enforce the setting of the target type as required. 

**Minimal Complete Reproducible example**

woniper:
Hello @casimk .
I think it should work like the code below.

```
    @Bean(name = "itemReader")
    public ItemReader<Establishment> itemReader() {
        return new FlatFileItemReaderBuilder<Establishment>()
                .name("itemReader")
                .linesToSkip(1)
                .resource(new ClassPathResource(FILE_NAME))
                .delimited()
                .names(FIELD_NAMES)
                .fieldSetMapper(new BeanWrapperFieldSetMapper<Establishment>() {
                    {
                        setTargetType(Establishment.class);
                        setDistanceLimit(0);
                        setStrict(false);
                    }
                }).build();
    }
```
setting targetType

OR

```
        @Bean(name = "itemReader")
	public ItemReader<Establishment> itemReader() {
		BeanWrapperFieldSetMapper<Establishment> mapper = new BeanWrapperFieldSetMapper<Establishment>() {
			{
				setTargetType(Establishment.class);
				setDistanceLimit(0);
				setStrict(false);
			}
		};
                // validate target type class
		mapper.afterPropertiesSet();

		return new FlatFileItemReaderBuilder<Establishment>()
				.name("itemReader")
[...]

PR #4393, opened 2023-06-07
Call afterPropertiesSet when set InitializingBean

woniper:
- PR for https://github.com/spring-projects/spring-batch/issues/4388


fmbenhassine:
Thank you for your PR. However, as explained in https://github.com/spring-projects/spring-batch/issues/4388#issuecomment-1590838036, it is not Spring Batch's responsibility to enforce bean lifecycle manually by checking if every collaborator is `instanceof InitializingBean` and calling `afterPropertiesSet` manually. It is up to the user to honor the `InitializingBean` contract if she/he provides an instance of a collaborator that is not managed by Spring.

Thank you for your contribution anyway.

PR #4413, opened 2023-07-06
Make RowMapper impl of batch domain public

### LINKED ISSUE #3928: Make RowMapper implementations of batch domain objects public
Currently, as of v4.3.3, `RowMapper` implementations of batch domain objects (`JobInstance`, `JobExecution`, `StepExecution`, and `ExecutionContext`) are [private inner classes](https://github.com/spring-projects/spring-batch/blob/193f82005a373819e5921c7e2255607cf239d14e/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcJobExecutionDao.java#L400) in their respective JDBC DAOs.

These mappers can be made public so they can be (re)used in custom DAOs implementations.

cgrings:
This change fixes #3928.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcExecutionContextDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcExecutionContextDao.java
@@ -363,7 +363,7 @@ public class JdbcExecutionContextDao extends AbstractJdbcBatchMetadataDao implem
 		return results;
 	}
 
-	private class ExecutionContextRowMapper implements RowMapper<ExecutionContext> {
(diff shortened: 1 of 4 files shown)


fmbenhassine:
Thank you for your PR. I was thinking of going a step further and extracting those mappers as separate classes. But even when doing this, some mappers won't be possible to make configurable, see https://github.com/spring-projects/spring-batch/issues/3928#issuecomment-1709962788. It would be possible to do it for `ExecutionContext` mapper (as it only depends on `serializer` and `charset`) and `JobInstance` mapper (as it is stateless), but not for job/step execution mappers as they are created dynamically based on method parameters.

For this reason, I am closing this PR. Thank you for your contribution anyway!

PR #4422, opened 2023-07-21
Antora

rwinch:
Migrate to Antora based documentaiton


fmbenhassine:
Thank you for opening the PR! I tested it locally and seems to correctly generate the HTML version of the docs. However, a couple notes:

* Is it intended that resources are duplicated in `modules/ROOT/assets/images` and `src/main/asciidoc/images`?
* I noticed that the `asciidoctor-maven-plugin` was removed. Therefore the pdf and epub3 versions are not generated anymore. Is it possible to generate them with the `antora-maven-plugin`?
* The `maven-deploy-plugin` was removed as well. This was required to skip deploying the docs as part of the release process as we have a separate [action](https://github.com/spring-projects/spring-batch/blob/main/.github/workflows/documentation-upload.yml) for that (and that is on purpose). Anything needed to update here?

The goal is to be able to keep the distribution generation process as described in the [REAME](https://github.com/spring-projects/spring-batch#using-the-command-line). If we have that, it won't be necessary to modify the release process (which I would prefer). Otherwise I am open to update the instructions if needed as well.

rwinch:
> Is it intended that resources are duplicated in modules/ROOT/assets/images and src/main/asciidoc/images?

Thanks for noticing this. It is not intentional and I've added a commit that corrects this.

> I noticed that the asciidoctor-maven-plugin was removed. Therefore the pdf and epub3 versions are not generated anymore. Is it possible to generate them with the antora-maven-plugin?

At the moment we are not trying to do any other formats. We will likely investigate using https://gitlab.com/antora/antora-assembler[Antora Assembler] to output PDFs in a later iteration.

> The maven-deploy-plugin was removed as well. This was required to skip deploying the docs as part of the release process as we have a separate [action](https://github.com/spring-projects/spring-batch/blob/main/.github/workflows/documentation-upload.yml) for that (and that is on purpose). Anything needed to update here?

The PR has a GitHub action that takes care of pushing your documentation for you. The build is structured in a way that rsync is preferred because it can copy only the files that have changed and sync a single version if that is all that is needed to be built. See https://github.com/spring-projects/spring-batch/pull/4423/files#diff-87e860af4b9db6c503ed680b6edb6333c6f8d7659f7d1a779a2487466bbc50f6R41

fmbenhassine:
Rebased, squashed and merged as 2e8d5063f7a68e45f25cd9aaf3e5075a6f250c30. Thank you for doing this!

Issue #4411, opened 2023-07-04
Add ability to start a job flow definition with a decider


fmbenhassine:
### Discussed in https://github.com/spring-projects/spring-batch/discussions/4359

<div type='discussions-op-text'>

<sup>Originally posted by **acktsap** April 12, 2023</sup>
In xml based job configuration, we can set decider first like this.

```xml
<job id="decisionJob1" >
    <decision id="decider1" decider="stepFlowDecider">
        <end on="EXIT" exit-code="COMPLETED"/>
        <next on="*" to="job1Step1" />
    </decision>

    <step id="job1Step1">
        <tasklet ref="createFileTasklet"/>
    </step>
</job>
```

But when using code based job configuration, no way to start with decider. `JobBuilder` only provides start methods taking `Step`, `Flow` only.

https://github.com/spring-projects/spring-batch/blob/972951a40af03c9c0553fb4c438f9aaeb1c9f1b5/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilder.java#L53-L78

It there any way to set decider first on code based job configuration?</div>

seonwoo960000:
Hi! @fmbenhassine. I've created a draft PR about this issue. Please feel free to check it anytime.

PR #4424, opened 2023-07-25
Allow `SimpleJobBuilder` to start with `JobExecutionDecider`

seonwoo960000:
### Modification 
- Add below method to allow users to specify `JobExecutionDecider` in the `start(...)` method as an argument. 

`JobBuilder.java` 
```java 
public JobFlowBuilder start(JobExecutionDecider decider) {
  return new SimpleJobBuilder(this).start(decider);
}
```

I found that we already have `SimpleJobBuilder` method that can start with `JobExecutionDecider`. So I think it's possible to add a method that utilizes `SimpleJobBuilder`'s method(above mentioned) which will allow users to specify `JobExecutionDecider` as an argument of `start` method. 

Resolves: https://github.com/spring-projects/spring-batch/issues/4411


seonwoo960000:
@fmbenhassine Thank you for the detailed information. I've added missing method(`FlowJobBuilder#start(JobExecutionDecider)`, javadoc and tests.

fmbenhassine:
Great 👍  Thank you for the updates. Rebased and merged as 3fe3b7fc5d0719ef3a83a74a71aa1e3c85c0712e.

PR #4430, opened 2023-08-10
Remove javaContent of Configuring a JobRepository

### LINKED ISSUE #4333: Doc issue: ambiguous description of Configuring a JobRepository
**Bug description**

The description of required configuration options of Configuring a JobRepository [in this chapter](https://docs.spring.io/spring-batch/docs/current/reference/html/job.html#configuringJobRepository) is ambiguous. We don't know whether `id` is required or `dataSource` and `transactionManager` are required.

image

cgrings:
Fixes #4333

### DIFF
--- a/spring-batch-docs/src/main/asciidoc/job.adoc
+++ b/spring-batch-docs/src/main/asciidoc/job.adoc
@@ -535,7 +535,7 @@ configuration options available, as the following example shows:
     transaction-manager="transactionManager"
     isolation-level-for-create="SERIALIZABLE"
     table-prefix="BATCH_"
-	max-varchar-length="1000"/>
+    max-varchar-length="1000"/>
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Thank you for this PR! The section reported in #4333 was confusing because the toggle of configuration styles was set on "Both", as described in https://github.com/spring-projects/spring-batch/issues/4333#issuecomment-1711364705.

However, this PR completely removes the Java section, which should rather be updated with an example that matches the one in XML, rather than be removed completely. 

The update was done in b2d76c7329e22ae50a50d8dc8b0b7272435b294e, which is now visible [here](https://docs.spring.io/spring-batch/reference/job/configuring-repository.html). Thank you for your contribution anyway!

Issue #867, opened 2018-06-05
Add Automatic-Module-Name to MANIFEST.MF in all components


spring-projects-issues:
**[Serban Iordache](https://jira.spring.io/secure/ViewProfile.jspa?name=siordache)** opened **[BATCH-2729](https://jira.spring.io/browse/BATCH-2729?redirect=false)** and commented

To avoid module hell when using spring-batch in modular applications, a stable [automatic module name](http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html) should be provided for all components.

---

**Affects:** 4.1.0

sukanth:
I can take this .. can you assign this to me ?

sukanth:
> I can take this .. can you assign this to me ? @fmbenhassine

here is the PR https://github.com/spring-projects/spring-batch/pull/4372

PR #4444, opened 2023-09-09
Added Automatic module name

sukanth:
fix for issue https://github.com/spring-projects/spring-batch/issues/867

image


fmbenhassine:
Thank you for the updates! LGTM now 👍 

I just wanted to mention that there is no need to add the `maven-jar-plugin` in all modules as it is already defined in the parent pom. I will update that on merge.

fmbenhassine:
Rebased and merged as 3eb7e09faa88bf4ed3954963551b914cd5c119a9. Thank you for your contribution!

sukanth:
> Thank you for the updates! LGTM now 👍
> 
> I just wanted to mention that there is no need to add the `maven-jar-plugin` in all modules as it is already defined in the parent pom. I will update that on merge.

Oh cool. , thank you.

PR #4445, opened 2023-09-09
Fix image link in README.md

acktsap:
Image link in `README.md` is broken.

Screen Shot 2023-09-10 at 3 28 27

It seems moved in https://github.com/spring-projects/spring-batch/pull/4422.


fmbenhassine:
Good catch! Thank you for the PR.

PR #4452, opened 2023-09-21
Use more `SynchronizedItemReader` in tests

hpoettker:
The test `AsyncTaskletStepTests::testStepExecutionUpdates` is flaky at the moment. At least on my machine, it fails rather reliably a handful of times with `@RepeatedTest(1000)`.

This PR fixes the test by using a `SynchronizedItemReader`. It also replaces the fix from #4291 with `SynchronizedItemReader`, which wasn't available previously. Furthermore, it polishes the two tests a bit with more recent Java features.


fmbenhassine:
Rebased and merged as cf1f34b4f9c7f3e405aa573b7f3ac167c0dca375. Thank you for your contribution!

PR #4454, opened 2023-09-21
Take care of some deprecations

hpoettker:
The PR takes care of some deprecations. In particular,

- it removes references to the Security Manager,
- it replaces a method of Spring Framework's `StopWatch` that will be deprecated with 6.1,
- it replaces Spring Framework's `ConcurrentTaskExecutor` whose zero-arguments constructor will be deprecated with 6.1


fmbenhassine:
Rebased and merged as f7ec69e891868811e9dd17faa67c6dae5a605760. Thank you for your contribution!

Issue #1213, opened 2015-06-08
Stop job execution if an exception is thrown in an after step [BATCH-2391]


spring-projects-issues:
**[Guillaume LE FLOCH](https://jira.spring.io/secure/ViewProfile.jspa?name=glefloch)** opened **[BATCH-2391](https://jira.spring.io/browse/BATCH-2391?redirect=false)** and commented

Currently, if an exception is thrown during a `beforeStep` or a step, the exception is catched and added to job failure exceptions.
If the exception is thrown during an `afterStep`, the exception is just catched and logged as an error but the job continue.
This can lead to a weird state for next steps... I think exception thrown in `afterStep` should be catched and added to job failure exception. This way if an exception is thrown the job is stopped and this is the same behavior for `beforeStep` and `afterStep`. If the developer want to the job to continue he just have to catch the execption and do what he want to.

---

**Affects:** 3.0.1, 3.0.2, 3.0.4

**Referenced from:** pull request https://github.com/spring-projects/spring-batch/pull/367

1 votes, 3 watchers

PR #4455, opened 2023-09-23
Terminates the job if a flag is set on afterStep

umbum:
As-is : Even if the terminateOnly flag is set on afterStep, It is ignored and the next step is executed. There is no way to terminate a job gracefully without using the `on(ExitStatus)` method when building the job.

To-be : If the terminateOnly flag is set, the job terminates gracefully.

It looks similar to https://github.com/spring-projects/spring-batch/issues/1213 but It's different.
If developers want to **explicitly** terminate a job, they should be able to do so.


fmbenhassine:
Thank you for opening this PR. I am not sure "isTerminateOnly" was designed to be used in a step execution listener. My thinking is that in the `afterStep`, the step has already ended successfully or with a failure, so it is too late to set that flag at that point. The issue more is about stopping the surrounding job (as mentioned by the javadocs). And this is where things become tricky, which I admit can be confusing.

There is a key difference between a `SimpleJob` and a `FlowJob`: Unlike `FlowJob`, `SimpleJob` is designed only for sequential executions of steps until the last step succeeds or one of the steps fail. So this implementation was not designed to be stopped programmatically.

Here is a quick example:

```java
import org.springframework.batch.core.ExitStatus;
import org.springframework.batch.core.Job;
import org.springframework.batch.core.JobParameters;
import org.springframework.batch.core.Step;
import org.springframework.batch.core.StepExecution;
import org.springframework.batch.core.StepExecutionListener;
import org.springframework.batch.core.job.builder.JobBuilder;
import org.springframework.batch.core.launch.JobLauncher;
import org.springframework.batch.core.step.builder.StepBuilder;
import org.springframework.batch.repeat.RepeatStatus;
import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
import org.springframework.jdbc.support.JdbcTransactionManager;

import javax.sql.DataSource;

@Configuration
@EnableBatchProcessing
public class HelloWorldJobConfiguration {

		ApplicationContext context = new AnnotationConfigApplicationContext(HelloWorldJobConfiguration.class);
		JobLauncher jobLauncher = context.getBean(JobLauncher.class);
		jobLauncher.run(job, new JobParameters());
	}

	@Bean
	public Step step1(JobRepository jobRepository, JdbcTransactionManager transactionManager) {
		return new StepBuilder("step1", jobRepository)
				.tasklet((contribution, chunkContext) -> {
			System.out.println("Hello");
			return RepeatStatus.FINISHED;
		}, transactionManager)
				.listener(new StepExecutionListener() {
					@Override
					public ExitStatus afterStep(StepExecution stepExecution) {
						stepExecution.setTerminateOnly();
						return stepExecution.getExitStatus();
					}
				})
				.build();
	}

	@Bean
	public Step step2(JobRepository jobRepository, JdbcTransactionManager transactionManager) {
		return new StepBuilder("step2", jobRepository).tasklet((contribution, chunkContext) -> {
			System.out.println("world!");
			return RepeatStatus.FINISHED;
		}, transactionManager).build();
	}

	@Bean
	public Job job(JobRepository jobRepository, Step step1, Step step2) {
		return new JobBuilder("job", jobRepository)
				.start(step1)
				.next(step2)
				.build();
	}

	// infra beans

	@Bean
				.addScript("/org/springframework/batch/core/schema-hsqldb.sql")
				.addScript("/org/springframework/batch/samples/common/business-schema-hsqldb.sql")
				.generateUniqueName(true)
				.build();
	}

	@Bean
	public JdbcTransactionManager transactionManager(DataSource dataSource) {
		return new JdbcTransactionManager(dataSource);
	}

}
```

In this case, step2 runs even if step1 was flagged as terminateOnly. However, if I only change the job type to `FlowJob`, things behave as expected:

```
@Bean
public Job job(JobRepository jobRepository, Step step1, Step step2) {
   return new JobBuilder("job", jobRepository)
				.flow(step1)
				.next(step2)
				.build()
				.build();
}
```

Do you see the difference?

> If developers want to explicitly terminate a job, they should be able to do so.

BTW, I find that stopping a job with a listener is implicit rather than explicit. The more explicit way would be to create a clear execution flow through the Java DSL or the XML namespace.

fmbenhassine:
Closing this as explained in my previous comment:

```
I am not sure "isTerminateOnly" was designed to be used in a step execution listener. My thinking is that in the afterStep, the step has already ended successfully or with a failure, so it is too late to set that flag at that point
```

But feel free to add a comment or open a PR if you think we can improve things in the documentation.

PR #4462, opened 2023-10-09
Add documentation to setSort method

farazahmadk:
Add documentation to setSort method regarding sort order when multiple sort values are given.


fmbenhassine:
Rebased and merged as 89db774509c2f4e5f6c447f8a168fe37fe89b594. I just amended the same change on the javadoc of the builder as well. Thank you for your contribution!

PR #4465, opened 2023-10-12
Add clear method to ExecutionContext

### LINKED ISSUE #4464: Add clear method to ExecutionContext
## **Expected Behavior**

I want to clear the information of [ExecutionContext](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java) after Job or Step execution.

## **Current Behavior**

However, there is currently no method to empty the entire [ExecutionContext](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java).

## **Context**

An [ExecutionContext](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java) can be used to address information that is common to tasks between the Steps that make up a Job, or within a Step.

Currently, the information contained in an `ExecutionContext` is stored in the `BATCH_JOB_EXECUTION_CONTEXT` table or the `BATCH_STEP_EXECUTION_CONTEXT` table when the Job ends.
Sometimes, this information may no longer be needed after the job runs. For example, if you use it for caching purposes.

However, there is currently no method to empty the entire `ExecutionContext`.

So if we had a public method to empty the entire `ExecutionContext`, we could prevent unnecessary data from being stored in the DB table after the batch job execution.

Since we are actually using `ConcurrentHashMap` under the hood, I think we can use [Map.clear()](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/util/Map.java#L336C18-L336C18) method.

doljae:
## Description

This PR adds a clear method to [ExecutionContext](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java). 
With this change, we can now remove all mappings from the map at once, improving code maintenance.

## Changes

- Added clear method to [ExecutionContext](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java)
- Added JavaDoc for the clear method.

Resolve #4464

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
@@ -372,6 +372,16 @@ public class ExecutionContext implements Serializable {
 	}
 
 	/**
+	 * Removes all mappings from this context.
+	 * After this method call, the context will be empty.
+	 *
+	 * @see java.util.Map#clear()
+	 */
+	public void clear() {
+		this.map.clear();
+	}
+
+	/**
 	 * Indicates whether or not a value is represented in this context.
 	 * @param value Value to check existence for
 	 * @return True if value is represented in context, false otherwise


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
```diff
@@ -371,6 +371,16 @@ public Object remove(String key) {
...
 	}
 
+	/**
+	 * Removes all mappings from this context.
+	 * After this method call, the context will be empty.
+	 *
+	 * @see java.util.Map#clear()
+	 */
+	public void clear() {
+		this.map.clear();
```
doljae, line 381: I found that this code base was worked on in 2008, and there's a few places where I could do some code cleansing, such as removing `this` keyword where it's clear that it doesn't need to be used.

However, code cleansing has a slightly different purpose than the current PR work, so I'll work on it as a separate PR if needed.

doljae, line 381: I found that this code base was worked on in 2008, and there's a few places where I could do some code cleansing, such as removing `this` keyword where it's clear that it doesn't need to be used.

However, code cleansing has a slightly different purpose than the current PR work, so I'll work on it as a separate PR if needed.

fmbenhassine:
I am closing this PR since the corresponding issue (#4464) was closed.

PR #4471, opened 2023-10-23
Add AbstractTaskletStepBuilder copy constructor

### LINKED ISSUE #4438: Incorrect step configuration when setting the taskExecutor before faultTolerant()
Hello,

I think I spotted an issue related to the StepBuilder. Depending on the order a developer build its step, the `taskExecutor` field will have different values.

Here is a minimal example to reproduce the issue:

`step1` and `step2` are exactly the same, except in one thing.
```java
public class SpringBatchTest {

    private PlatformTransactionManager transactionManager = mock(PlatformTransactionManager.class);
    private JobRepository jobRepository =  mock(JobRepository.class);
    private ItemReader itemReader = mock(ItemReader.class);
    private ItemProcessor itemProcessor = mock(ItemProcessor.class);
    private ItemWriter itemWriter = mock(ItemWriter.class);

    private SimpleAsyncTaskExecutor taskExecutor  = new SimpleAsyncTaskExecutor();

    @Test
    void test_inconsistent_behavior() {
        TaskletStep step1 = new StepBuilder("step-name", jobRepository)
                .chunk(10, transactionManager)
                .reader(itemReader)
                .processor(itemProcessor)
                .writer(itemWriter)
                .faultTolerant()
                .taskExecutor(taskExecutor)
                .build();
        TaskletStep step2 = new StepBuilder("step-name", jobRepository)
                .chunk(10, transactionManager)
                .taskExecutor(taskExecutor)// The task executor is set before faultTolerant()
                .reader(itemReader)
                .processor(itemProcessor)
                .writer(itemWriter)
                .faultTolerant()
                .build();

        RepeatTemplate stepOperations1 = (RepeatTemplate) ReflectionUtils
                .readFieldValue(TaskletStep.class, "stepOperations", step1).get();// TaskExecutorRepeatTemplate
        RepeatTemplate stepOperations2 = (RepeatTemplate) ReflectionUtils
                .readFieldValue(TaskletStep.class, "stepOperations", step2).get();// RepeatTemplate

        assertEquals(stepOperations1.getClass(), stepOperations2.getClass());// This fails
    }
}
```

In this example, `step2` will use the default task executor instead of the provided one.

Looking at the code, the `taskExecutor` value is lost after setting `faultTolerant()`.

I did not go too deep into the solution, but I think the issue comes from this constructor (see link). The `taskExecutor` value could be kept from there.

https://github.com/spring-projects/spring-batch/blob/75025a24701eafb2387a2cc8c937d585f576401a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java#L73

Ilpyo-Yang:
## Description
This PR adds a copy constructor to the AbstractTaskletStepBuilder to address an issue arising during the property setting process with taskExecutor. 

## Changes
+ Add AbstractTaskletStepBuilder copy constructor
+ Remove unnecessary transactionManager in SimpleStepBuilder copy constructor
+ Add testcode

Resolve #4438

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
@@ -49,6 +49,7 @@ import org.springframework.transaction.interceptor.TransactionAttribute;
  * @author Dave Syer
  * @author Michael Minella
  * @author Mahmoud Ben Hassine
+ * @author Ilpyo Yang
  * @since 2.2
  * @param <B> the type of builder represented
  */
@@ -72,6 +73,23 @@ public abstract class AbstractTaskletStepBuilder<B extends AbstractTaskletStepBu
 
 	public AbstractTaskletStepBuilder(StepBuilderHelper<?> parent) {
 		super(parent);
+	}
+
+	/**
+	 * Create a new builder initialized with any properties in the parent. The parent is
+	 * copied, so it can be re-used.
+	 * @param parent a parent helper containing common step properties
+	 */
+	public AbstractTaskletStepBuilder(AbstractTaskletStepBuilder<?> parent) {
+		super(parent);
+		this.chunkListeners = parent.chunkListeners;
+		this.stepOperations = parent.stepOperations;
+		this.transactionManager = parent.transactionManager;
+		this.transactionAttribute = parent.transactionAttribute;
+		this.streams.addAll(parent.streams);
+		this.exceptionHandler = parent.exceptionHandler;
+		this.throttleLimit = parent.throttleLimit;
+		this.taskExecutor = parent.taskExecutor;
 	}
 
 	protected abstract Tasklet createTasklet();
--- a/spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
+++ b/spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
@@ -1,0 +1,114 @@
+package org.springframework.batch.test;
+/*
+ * Copyright 2020-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.batch.core.repository.JobRepository;
+import org.springframework.batch.core.step.builder.AbstractTaskletStepBuilder;
+import org.springframework.batch.core.step.builder.SimpleStepBuilder;
+import org.springframework.batch.core.step.builder.StepBuilderHelper;
+import org.springframework.batch.item.ItemProcessor;
+import org.springframework.batch.item.ItemReader;
+import org.springframework.batch.item.ItemWriter;
+import org.springframework.batch.test.context.SpringBatchTest;
+import org.springframework.core.task.SimpleAsyncTaskExecutor;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * Test cases for verifying the {@link AbstractTaskletStepBuilder} and faultTolerant() functionality.
+ *
+ * @author Ilpyo Yang
+ */
+@SpringBatchTest
+@SpringJUnitConfig
+public class AbstractTaskletStepBuilderTests {
+	private final JobRepository jobRepository =  mock(JobRepository.class);
+	private final int chunkSize = 10;
+	private final ItemReader itemReader = mock(ItemReader.class);
+	private final ItemProcessor itemProcessor = mock(ItemProcessor.class);
+	private final ItemWriter itemWriter = mock(ItemWriter.class);
+	private final SimpleAsyncTaskExecutor taskExecutor  = new SimpleAsyncTaskExecutor();
+	SimpleStepBuilder simpleStepBuilder;
+
+	private <T> T accessPrivateField(Object o, String fieldName) throws ReflectiveOperationException {
+		Field field = o.getClass().getDeclaredField(fieldName);
+		field.setAccessible(true);
+		return (T) field.get(o);
+	}
+
+	private <T> T accessSuperClassPrivateField(Object o, String fieldName) throws ReflectiveOperationException {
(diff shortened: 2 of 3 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
```diff
@@ -74,6 +74,25 @@ public AbstractTaskletStepBuilder(StepBuilderHelper<?> parent) {
 		super(parent);
 	}
 
+	/**
+	 * Create a new builder initialized with any properties in the parent. The parent is
+	 * copied, so it can be re-used.
+	 * @param parent a parent helper containing common step properties
+	 *
+	 * @author Ilpyo Yang
```
injae-kim, line 82: Maybe `@author` should be located on `L51`, not here?

spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+		assertEquals(chunkSize, copyChunkSize);
+		assertEquals(itemReader, copyItemReader);
+		assertEquals(itemProcessor, copyItemProcessor);
+		assertEquals(itemWriter, copyItemWriter);
+	}
+
+	@Test
+	void faultTolerantMethodTest() throws ReflectiveOperationException {
+		simpleStepBuilder.taskExecutor(taskExecutor);	// The task executor is set before faultTolerant()
```
injae-kim, line 91:
```
assertEquals(taskExecutor, afterTaskExecutor);
```
Can we check taskExecutor is copied well after faultTolerant()?

spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
Ilpyo-Yang: change to ```L51``` completed!

spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
```diff
@@ -74,6 +75,23 @@ public AbstractTaskletStepBuilder(StepBuilderHelper<?> parent) {
...
+	 * copied, so it can be re-used.
+	 * @param parent a parent helper containing common step properties
+	 */
+	public AbstractTaskletStepBuilder(AbstractTaskletStepBuilder<?> parent) {
+		super(parent);
+		this.chunkListeners = parent.chunkListeners;
+		this.stepOperations = parent.stepOperations;
+		this.transactionManager = parent.transactionManager;
+		this.transactionAttribute = parent.transactionAttribute;
+		this.streams.addAll(parent.streams);
```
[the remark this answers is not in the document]
Ilpyo-Yang: > Hmm maybe we can just use `parent.streams` instead of `addAll()`?

To creating a copy constructor, it seems like the second option, ```this.streams = parents.streams```, is appropriate. Thank you for your comment @injae-kim!

But just curious.. was there any specific reason for initially declaring ```streams``` as a ```final``` variable?

spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
```diff
@@ -74,6 +75,23 @@ public AbstractTaskletStepBuilder(StepBuilderHelper<?> parent) {
 		super(parent);
 	}
 
+	/**
+	 * Create a new builder initialized with any properties in the parent. The parent is
+	 * copied, so it can be re-used.
+	 * @param parent a parent helper containing common step properties
+	 */
+	public AbstractTaskletStepBuilder(AbstractTaskletStepBuilder<?> parent) {
```
acktsap, line 83: How about making it protected?

spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
```diff
@@ -0,0 +1,114 @@
+package org.springframework.batch.test;
```
acktsap, line 1: I think package line should be followed by copyright comments. ([see also](https://github.com/spring-projects/spring-batch/blob/main/CONTRIBUTING.md#add-apache-license-header-to-all-new-classes))

spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
+
+/**
+ * Test cases for verifying the {@link AbstractTaskletStepBuilder} and faultTolerant() functionality.
+ *
+ * @author Ilpyo Yang
+ */
+@SpringBatchTest
+@SpringJUnitConfig
+public class AbstractTaskletStepBuilderTests {
```
acktsap, line 44: I think this test code should be placed somewhere in [spring-batch-core/src/test/java...](https://github.com/spring-projects/spring-batch/tree/main/spring-batch-core/src/test/java/org/springframework/batch/core).

spring-batch-test module is for providing test utils for spring batch users.

spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/AbstractTaskletStepBuilder.java
Ilpyo-Yang: oh I will try thanks!

spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
Ilpyo-Yang: umm that is why this not merged.. I missed few things in contribute regulation. Thank you for kindness review. I will update all in this holiday

Ilpyo-Yang: @acktsap thanks for the detailed review, and I've incorporated the suggested changes.

fmbenhassine:
Thank you for the PR! It fixes the issue 👍 

However, for the test, there is no need to test the copy constructor itself. I think something like the test suggested in #4438 is enough to validate the fix and cover the code path for any future regression. 

Could you please update the test and rebase/squash the changes? It should then be good to merge. Thank you upfront!

Ilpyo-Yang:
Hello @fmbenhassine! 
I've updated the commits as per your guidance. The test code now matches the issue details. I've stuck with the TaskletStep approach, finding it most suitable. Please review at your convenience. Thank you!

spring-batch-test/src/test/java/org/springframework/batch/test/AbstractTaskletStepBuilderTests.java
fmbenhassine: I still see this class in the `spring-batch-test` module. No problem, I will move it to the core module on merge.

fmbenhassine:
Thank you for the updates, @Ilpyo-Yang ! I noticed that the tests pass even without the changes in the PR. I think we can simplify tests by checking that the type of the repeat template is correct whether the task executor is set before or after `.faultTolerant()`. I will take care of that on merge.

fmbenhassine:
LGTM 👍 Rebased and merged as b9ba8ffb861119924cbb4e17069ddd9e911aaa91. I just updated the tests as mentioned before in 4b8e504972a59a99e33e634a12c6ceebbad5fca8. Thank you for your contribution!

PR #4475, opened 2023-10-25
Fix `FlowBuilder.next().end()` infinite loop

### LINKED ISSUE #4432: The first step in "Flow" is execute twice
**Bug description**
If you create a "Flow" and use the "next" method instead of the "start" method to add the first step, then this step will be executed twice.

The problem is in the line `addTransition("COMPLETED", next);`. It is she who makes Flow perform the step a second time.

https://github.com/spring-projects/spring-batch/blob/e6c27273fa2b3713c6f2d472bf3de1b18f8e5eba/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java#L249

Current version:
```java

	private void doNext(Object input) {
		if (this.currentState == null) {
			doStart(input);
		}
		State next = createState(input);
		addTransition("COMPLETED", next);
		addTransition("*", failedState);
		this.currentState = next;
	}

```

Fixed version:
```java

	private void doNext(Object input) {
		if (this.currentState == null) {
			doStart(input);
		} else {
			State next = createState(input);
			addTransition("COMPLETED", next);
                        addTransition("*", failedState);
			this.currentState = next;
		}
		
	}

```

**Environment**
```
	<groupId>org.springframework.boot</groupId>
 	<artifactId>spring-boot-starter-batch</artifactId>
 	<version>3.1.2</version>
```

**Steps to reproduce**
```java
	FlowBuilder<Flow> flowBuilder = new FlowBuilder<>("someFlowName");
	flowBuilder.next(step);
	return flowBuilder.build();
```

**Expected behavior**
Created "Flow" with one step. Step executed once.

The step was to transfer data from one table to another identical table. The second run of the step failed only because a unique primary key constraint was created.
image

injae-kim:
related issue #4432

### Motivation:
```java
builder.next(createCompleteStep("stepA"))
       .end()
       .start();
--
[main] INFO org.springframework.batch.core.job.SimpleStepHandler - Executing step: [stepA]
[main] INFO org.springframework.batch.core.job.SimpleStepHandler - Executing step: [stepA]
[main] INFO org.springframework.batch.core.job.SimpleStepHandler - Executing step: [stepA]
... "stepA" never finishes
```
- On #4432, we found that `FlowBuilder.next().end()` never finishes
- Problem is that `addTransition("COMPLETED", next);` on `next()` although there's no step registered yet

### Modification:
- If there's no step registered yet on `FlowBuilder.next().end()`, handle same as ``FlowBuilder.start().end()`` to avoid infinite loop

### Result:
- `FlowBuilder.next().end()` works same as `FlowBuilder.start().end()` and finishes well
- Close #4432 issue

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
@@ -48,6 +48,7 @@ import org.springframework.core.task.TaskExecutor;
  * @author Dave Syer
  * @author Michael Minella
  * @author Mahmoud Ben Hassine
+ * @author Injae Kim
  * @since 2.2
  * @param <Q> the type of object returned by the builder (by default a Flow)
  *
@@ -107,7 +108,7 @@ public class FlowBuilder<Q> {
 
 	/**
 	 * Transition to the next step on successful completion of the current step. All other
-	 * outcomes are treated as failures.
+	 * outcomes are treated as failures. If no steps are registered yet just a synonym for {@link #start(Step)}.
 	 * @param step the next step
 	 * @return this to enable chaining
 	 */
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
```diff
@@ -247,29 +248,32 @@ protected Flow flow() {
...
+		} else {
+			State next = createState(input);
+			addTransition("COMPLETED", next);
+			addTransition("*", failedState);
+			currentState = next;
 		}
-		State next = createState(input);
-		addTransition("COMPLETED", next);
-		addTransition("*", failedState);
-		this.currentState = next;
```
injae-kim, line 256:
```java
		if (this.currentState == null) {
			doStart(input); // 1. currentState: input
		}

		State next = createState(input);
        // 2. input -> COMPLETED -> input -> COMPLETED -> input ..
		addTransition("COMPLETED", next); 
```

when `currentState == null`, I think we should not `addTransition("COMPLETED", next);` cause it makes infinite loop 🤔

spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/FlowBuilder.java
```diff
@@ -247,29 +248,32 @@ protected Flow flow() {
...
-		this.currentState = createState(input);
 	}
 
 	private void doFrom(Object input) {
 		if (currentState == null) {
 			doStart(input);
+		} else {
+			State state = createState(input);
+			tos.put(currentState.getName(), currentState);
+			currentState = state;
```
injae-kim, line 275: When `currrent == null`, previous `doFrom` logic seems `createState(input)` twice (inside of `doStart()` and `L273`) so I fix it~!

fmbenhassine:
LGTM 👍  Rebased and merged as 4a67b22e9040e5e4c69da3eb1b95dcd27f6cdc0e. Thank you for your contribution!

injae-kim:
Thanks a lot for review&merge! I'll make another contribution soon 🙇

PR #4477, opened 2023-10-28
When adding steps to the “Flow” via “next”, the task no longer goes into loop execution

bio-engineer:
When adding steps to the “Flow” via “next”, the task no longer goes into loop execution
 
Fixes spring-projects/spring-batch#4432

The project is completed build. All tests were completed successful


fmbenhassine:
Thank you for this PR! We have already an open PR #4475 with a similar fix. Can you please check if this is a duplicate? Thank you upfront.

bio-engineer:
I looked at the specified merge request and made sure that it partially duplicates my logic.

The problem is that the number of changes made to it is too large to close [Issue](https://github.com/spring-projects/spring-batch/issues/4432).

So I can't guarantee that the changes made ([there](https://github.com/spring-projects/spring-batch/pull/4475)) don't break other logic

Issue #4326, opened 2023-03-15
Deprecate jobRepository method in JobBuilderHelper & StepBuilderHelper


acktsap:
As of 5.0.0, it's recommanded to use JobBuilder, StepBuilder with `JobRepository` like this.

```java
@Bean
public Job testJob(JobRepository jobRepository) {
    return new JobBuilder("testJob", jobRepository)
        .start(
            new StepBuilder("testStep", jobRepository)
                .tasklet((a, b) -> RepeatStatus.FINISHED, new ResourcelessTransactionManager())
                .build()
        )
        .build();
}
```

But i still can set `JobRepository` when building `Job` or `Step` like this.

```java
@Bean
public Job testJob(JobRepository jobRepository) {
    return new JobBuilder("testJob", jobRepository)
        .start(
            new StepBuilder("testStep", jobRepository)
                .tasklet((a, b) -> RepeatStatus.FINISHED, new ResourcelessTransactionManager())
                .repository(jobRepository) // this
                .build()
        )
        .repository(jobRepository) // this
        .build();
}
```

It looks redundant since it's already set in constructor. So, how about remove repository method in `JobBuilderHelper` and `StepBuilderHelper`? Maybe it's better to deprecate first and remove in 6.0.0

https://github.com/spring-projects/spring-batch/blob/4513d78ba5241cd5273bfdb42b381ad7ce47140b/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilderHelper.java#L94-L104

https://github.com/spring-projects/spring-batch/blob/4513d78ba5241cd5273bfdb42b381ad7ce47140b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/StepBuilderHelper.java#L69-L72

fmbenhassine:
Thank you for reporting this. It makes sense, and indeed, it could be confusing. I will plan the deprecation for v5.1.

PR #4495, opened 2023-11-21
Deprecate overrides of `StepBuilderHelper#repository` for removal

hpoettker:
This is a follow-up on #4326. It deprecates also the overrides of `StepBuilderHelper#repository`.

The deprecation does not seem to be inherited by the overrides. At least my IDE does not highlight their use without explicit deprecations.


fmbenhassine:
LGTM. Rebased and merged as 89e0a0cd86d9072473338813d7d63d5c5a5f26b3. Thank you!

PR #4503, opened 2023-11-30
Check dirty flag of step execution context before update in inner loop

hpoettker:
The class `TaskletStep.ChunkTransactionCallback` persists updates of the `StepExecution` after each chunk. The update of the step execution context happens in a dedicated method, which is currently called independent of whether the step execution context has actually changed. As `ExecutionContext` provides a dirty flag, it can be checked before the update method is called.

The proposed change is a performance optimization for simple chunk-oriented steps that e.g. do not save state in between chunks. For such steps, the change results in one database round trip less per chunk.


fmbenhassine:
LGTM 👍  Rebased and merged as 74520a597abf4ee7faf7b83354f468e2ea6b002a. Thank you!

PR #4504, opened 2023-11-30
Updated documentation and added unit tests for flow builder priority

robertmcnees:
Fixes #4456

Issue #4456, opened 2023-09-25
Flow order priority does not match documented behavior


robertmcnees:
**Bug description**

When using a [JobExecutionDecider](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/job/flow/JobExecutionDecider.html) the behavior is not accurate according to the [documentation of DefaultStateTransitionComparator](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.html). Specifically 

> Hence * > foo* > ??? > fo? > foo

Additionally, the actual behavior when running a Job is not consistent with the test results found in [DefaultStateTransitionComparatorTests](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/test/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparatorTests.java)

In almost all test scenarios the documented behavior is not followed.

**Environment**

Boot 3.1.4
Java 17

**Steps to reproduce**

See sample GitHub project to be linked in comments below.  The JUnit tests have multiple failing scenarios that should pass.

The tests come in 3 different groups, all of which contain failing tests:

1. Tests based solely on the example found in the [documentation](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/job/flow/JobExecutionDecider.html)
2. Tests based on the scenarios in [DefaultStateTransitionComparator](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.html)
3. Tests created from interpretation of the documentation

**Expected behavior**

The stated documentation should accurate in regards to the priority Flow.  All tests in the sample project should pass.

hpoettker:
Hi @robertmcnees, in your tests you seem to assume that Spring Batch will jump to the most generic pattern if there is more than one match. This is **not** the case. Spring Batch will choose the most specific one.



#### Documented behavior of Spring Batch

The precedence of transitions is described in the reference documentation: https://docs.spring.io/spring-batch/docs/current/reference/html/step.html#conditionalFlow

It gives the following example
```
@Bean
public Job job(JobRepository jobRepository) {
    return new JobBuilder("job", jobRepository)
     .start(stepA())
     .on("*").to(stepB())
     .from(stepA()).on("FAILED").to(stepC())
     .end()
     .build();
}
```
and states

[...]

Issue #3996, opened 2021-09-10
Order of Transition Elements Still Does Not completely adhere to documentation


chpasha:
**Bug description**
The original issue was fixed here https://github.com/spring-projects/spring-batch/issues/3638 but the the way the DefaultStateTransitionComparator works is stll different from what documentation says:
it states that 
>The framework automatically orders transitions from most specific to least specific

but the javadoc in comparator says
> If wildcard counts are equal then falls back to alphabetic comparison. Hence * > foo* > ??? > fo? > foo.

this looks wrong to me since foo* is more specific pattern then  * so it should fire before. Consider situation where Step execution can return COMPLETED or COMPLETED_PARTIALLY (alongside other possible statuses) and those two completed states should lead to different flow then FAILED (or any other) but conditional flow 

```
.start(mainWorkflowStep).on("COMPLETED*").to(step1)
.from(mainWorkflowStep).on("*").to(step2)
```
doesn't work as expected because "*" fires and leads to step2 even on COMPLETED or COMPLETED_PARTIALLY
Current workaround is to use ** instead of single * so that COMPLETED* is considered "more important"

**Environment**
Spring Batch version : 4.3.3, 
Java version 11

**Expected behavior**
At least pattern "XXX*" should be considered more specific than just "*" since the latter is catchall pattern

fmbenhassine:
Thank you for opening this issue. We do not exclude that this could be a bug in Spring Batch. However, we would like to validate that with a minimal complete verifiable example.

Could you please take some time to create a minimal example that reproduces the problem? To help you in reporting your issue, we have prepared a project template that you can use as a starting point. Please check the [Issue Reporting Guidelines](https://github.com/spring-projects/spring-batch/blob/main/ISSUE_REPORTING.md) for more details about this.

Thank you for your collaboration.

chpasha:
Hi fmbenhassine,
here you go
[spring-batch-mcve.zip](https://github.com/spring-projects/spring-batch/files/11753669/spring-batch-mcve.zip)

robertmcnees:
Thanks @chpasha for the sample.  I easily got it running on my machine and can see the issue.  After entering my own issue #4456 that @hpoettker helped me sort out I came back to this and suggested a fix with a PR.

I was able to point your sample at my local Spring Batch build using this code and it should now have the expected behavior.  There is also a new test `testBuildWithDeciderPriority` in `FlowJobBuilderTests` that should be a direct test as well.

Looking for feedback...

PR #4509, opened 2023-12-04
Modified calculating order for DefaultStateTransitionComparator

robertmcnees:
Expanded the functionality of the `DefaultStateTransitionComparator` to match the documentation and catch edge cases.

The original documentation still holds:

`* > foo* > ??? > fo? > foo`

And I more complex scenarios are now also covered:

`* > foo* > *f* > *foo* > ??? > ?o? > foo?? > bar?? > fo? > foo > bar`

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.java
@@ -20,12 +20,31 @@ import org.springframework.util.StringUtils;
 import java.util.Comparator;
 
 /**
- * Sorts by decreasing specificity of pattern, based on just counting wildcards (with *
- * taking precedence over ?). If wildcard counts are equal then falls back to alphabetic
- * comparison. Hence * &gt; foo* &gt; ??? &gt; fo? &gt; foo.
+ * Sorts by ascending specificity of pattern, based on counting wildcards (with * taking
+ * precedence over ?). Hence * &gt; foo* &gt; ??? &gt; fo? &gt; foo.
+ *
+ * For more complex comparisons, any string containing at least one * token will be
+ * considered more generic than any string that has no * token. If both strings have at
+ * least one * token, then the string with fewer * tokens will be considered the most
+ * generic. If both strings have the same number of * tokens, then the comparison will
+ * fall back to length of the overall string with the shortest value being the most
+ * generic. Finally, if the * token count is equal and the string length is equal then the
+ * final comparison will be alphabetic.
+ *
+ * When two strings have ? tokens, then the string with the most ? tokens will be
+ * considered the most generic. If both strings have the same number of ? tokens, then the
+ * comparison will fall back to length of the overall string with the shortest value being
+ * the most generic. Finally, if the ? token count is equal and the string length is equal
+ * then the final comparison will be alphabetic
+ *
+ * If the strings contain neither * nor ? tokens then alphabetic comparison will be used.
+ *
+ * Hence * &gt; foo* &gt; *f* &gt; *foo* &gt; ??? &gt; ?o? &gt; foo?? &gt; bar?? &gt; fo?
+ * &gt; foo &gt; bar
  *
  * @see Comparator
  * @author Michael Minella
+ * @author Robert McNees
  * @since 3.0
  */
 public class DefaultStateTransitionComparator implements Comparator<StateTransition> {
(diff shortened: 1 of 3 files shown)


robertmcnees:
Fix for #3996

spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.java
```diff
@@ -20,12 +20,31 @@
 import java.util.Comparator;
 
 /**
- * Sorts by decreasing specificity of pattern, based on just counting wildcards (with *
- * taking precedence over ?). If wildcard counts are equal then falls back to alphabetic
- * comparison. Hence * &gt; foo* &gt; ??? &gt; fo? &gt; foo.
+ * Sorts by ascending specificity of pattern, based on counting wildcards (with * taking
+ * precedence over ?). Hence * &gt; foo* &gt; ??? &gt; fo? &gt; foo.
```
fmbenhassine, line 24: It is true that this example sorts the patterns from most generic to most specific (ie in ascending order of specificity), but the comparator itself returns the most specific pattern first and the least specific pattern last. This means the specificity is decreasing, not increasing. I know this is subtle, but I think the initial javadoc `Sorts by decreasing specificity of pattern` is correct (which I seem to have missed when I reviewed #4504). Do you agree?

spring-batch-core/src/test/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparatorTests.java
```diff
@@ -37,42 +37,98 @@ void testSimpleOrderingEqual() {
 
 	@Test
 	void testSimpleOrderingMoreGeneral() {
-		StateTransition transition = StateTransition.createStateTransition(state, "CONTIN???LE", "start");
-		StateTransition other = StateTransition.createStateTransition(state, "CONTINUABLE", "start");
-		assertEquals(1, comparator.compare(transition, other));
-		assertEquals(-1, comparator.compare(other, transition));
+		StateTransition generic = StateTransition.createStateTransition(state, "CONTIN???LE", "start");
+		StateTransition specific = StateTransition.createStateTransition(state, "CONTINUABLE", "start");
```
fmbenhassine, line 41: I like this! A simple naming change makes the tests immediately more readable.

fmbenhassine:
I tested this PR and the reported edge case where the pattern `*` is returned in priority instead of `foo*` (due to alphabetical sorting when both patterns have the same number of `*`) is now fixed without any regression 👍 

@robertmcnees I just wanted to check with you the subtle detail about the `decreasing` / `increasing` wording in the javadoc before merging the PR. Other than that, LGTM.

spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.java
robertmcnees: As you said this is subtle so I'm OK with leaving it as is.  I've been confused by this many times and I think the more important part is the example in the documentation than the word ascending/descending.  

However to detail my thinking for anyone who stumbles on this later:

> the comparator itself returns the most specific pattern first

The way I read it the comparator [considers the more generic patterns to be greater than the more specific](https://github.com/spring-projects/spring-batch/blob/21555c093e4c9b97a066878c9e392afcf9a58531/spring-batch-core/src/test/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparatorTests.java#L47).  So the comparator would return * > ?foo > foo, which I would call ascending order of specificity (most specific last).

This was the confusing part for me (quoting myself from a [separate issue](https://github.com/spring-projects/spring-batch/issues/4456#issuecomment-1834320231)):

> the order of specificity switches between the DefaultStateTransitionComparator and what is used in the FlowJobBuilder. That is while DefaultStateTransitionComparator will return * > foo* > ??? > fo? > foo, the FlowJobBuilder will use the most specific case first. This turns the example backwards to foo > fo? > ??? > foo* > *.

In short I think the `DefaultStateTransitionComparator` sorts in _ascending_ order of specificity, but `FlowJobBuilder` sorts in _descending_ order of specificity.  `FlowJobBuilder` and _descending_ order of specificity is what the end user interacts with, so maybe it's best to keep the documentation as 'descending' throughout.

robertmcnees:
Thanks for the review, @fmbenhassine!  I left a comment on the `decreasing` / `increasing` conversation, but I think it's very subtle and I'm fine with either way.

spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/support/DefaultStateTransitionComparator.java
fmbenhassine: Thank you for the clarification, @robertmcnees ! I see, my interpretation of the natural ordering of elements (ie the meaning/definition of the `>` symbol)  in the `DefaultStateTransitionComparator` was inaccurate.

> In short I think the DefaultStateTransitionComparator sorts in ascending order of specificity, but FlowJobBuilder sorts in descending order of specificity.

I believe this is the root cause of the confusion. There is probably a good reason for that (probably the use of `TreeSet` in `SimpleFlow`?), but my thinking is that since we want transitions to be sorted by the most specific first (like in the issue's example `foo*` should be fired before `*`), then the comparator should result in `foo*` being lower than `*`. Similar to the natural ordering of numbers, the analogy I like to compare to is that if I assign weights to transitions (like `foo* -> 1` and `* -> 2`), the "natural" ordering coming out of the comparator itself would rank `foo*` lower than `*` (that's the "natural" order I was expecting, not the opposite).

It would be great if we try to avoid the ordering to be opposite between the comparator and the flow implementation. This is important because as a Batch user, if I want to provide a custom comparator, I would not have to do this mental effort to think backward in the comparator's implementation. Otherwise, we can keep the changes in this PR and add a clear definition of what ">" means in this context.

robertmcnees: That's a great point about someone wanting to create their own comparator.  I was only thinking about it from my narrow perspective of debugging the code.  I think making the `DefaultStateTransitionComparator` match `FlowJobBuilder` makes even more sense with that consideration.

I created a new issue #4527 so that we can track the solution there.  It will likely involve changes that are out of scope to this PR and issue #3996.

robertmcnees:
I created a new issue #4527 to better track how `FlowJobBuilder` interprets the results from `DefaultStateTransitionComparator`.  That issue could be handled independently of this PR and issue #3996.  However given that this is a slight breaking change, I see the issue targeted for the 5.2.0 release which means we have time.

What is your preference @fmbenhassine?  Would you rather handle the new issue #4527 in a different PR or would you rather adapt this one to have a single PR that covers both issues (#4527 and #3996)?

fmbenhassine:
Rebased and merged as 9d139b565d4dc21ca3fd1034f47ec5ef23c5bfa4. Thank you for your contribution!

PR #4512, opened 2023-12-09
Use annotation `@Override` consistently

hpoettker:
The codebase currently contains some comments with "non-Javadoc", which, I think, are predeccesors in spirit to the `@Override` annotation and have been auto-generated by IDEs in the past.

This PR removes these remnants and adds the annotation `@Override` where it is currently missing.

In the test code, I've also removed some methods in support classes that actually don't override anything.


fmbenhassine:
Merged as 0d1d89c82fc07194e8e36f11be9a0e164caf6333. Thank you for your PR!

Issue #4519, opened 2023-12-17
Multiple trationDelegate$BeanPostProcessorChecker Warnings Arise When Using Spring Boot 3.2.0


woosungkim0123:
**Bug description**
When attempting to run a Batch process in Spring Boot 3.2.0, multiple BeanPostProcessorChecker warnings are being logged.

**Environment**
- Spring Batch version: 5.1
- Java version: 17, 21
- Database used (if applicable): H2
- Additional environment details: Spring Boot 3.2.0

**Steps to reproduce**
1. Configure an application using Spring Boot 3.2.0, Spring Batch 5.1, and H2 database.
2. Run the application.
3. BeanPostProcessorChecker warnings appear in the log.

**Expected behavior**
The application should initialize and run without any BeanPostProcessorChecker warnings.

**Minimal Complete Reproducible example**
The warnings that are being displayed have been documented in the README of the demo project.
[minimal example](https://github.com/woosungkim0123/batch-issue-demo)

hpoettker:
The issue is caused by the automatic bean exposure of `JobRegistryBeanPostProcessor` that has been introduced with #4245.

If you don't need the job beans to be registered with the `JobRegistry`, or use your own custom mechanism, you can disable it with code like this in one of your configuration classes:
```
@Bean
public static BeanDefinitionRegistryPostProcessor jobRegistryBeanPostProcessorRemover() {
  return registry -> registry.removeBeanDefinition("jobRegistryBeanPostProcessor");
}
```
The warnings should then disappear.

woosungkim0123:
thank you The issue has been resolved.

pmsinclair:
For applications with: @EnableBatchProcessing(modular = true)
This can actually cause: **DuplicateJobException**
This is due to both of these trying to register the same jobs:
1. JobRegistryBeanPostProcessor
2. AutomaticJobRegistrarBeanPostProcessor

woosungkim0123:
If this is what you intended or has been implemented in a new version, please feel free to close the issue.

nightswimmings:
The issue indicated by the warning actually affects the app. I have a  BeanPostProcessor that does not have any dependency and expects a certain bean autoconfigured and exposed by me in another section of the code, and as long as I don't add the removal mentioned by [hpoettker](https://github.com/hpoettker), my custompostprocessor cannot find my custom bean

fmbenhassine:
[...]

Issue #4489, opened 2023-11-17
PostProcessorRegistrationDelegate$BeanPostProcessorChecker.postProcessAfterInitialization WARNING after upgrade spring-framework-6.1.0


leonchen83:
Please do a quick search on Github issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
```
23-11-17 16:19:16.578 WARN [main PostProcessorRegistrationDelegate$BeanPostProcessorChecker.postProcessAfterInitialization:437]Bean 'batch.job.registry' of type [org.springframework.batch.core.configuration.support.MapJobRegistry] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying). Is this bean getting eagerly injected into a currently created BeanPostProcessor [org.springframework.batch.core.configuration.support.JobRegistryBeanPostProcessor#0]? Check the corresponding BeanPostProcessor declaration and its dependencies.
```

```
```

**Environment**
```
spring-batch : 5.0.3
spring-framework : 6.1.0
```

**Steps to reproduce**
```
    <bean class="org.springframework.batch.core.configuration.support.JobRegistryBeanPostProcessor">
        <property name="jobRegistry" ref="batch.job.registry" />
    </bean>

    <bean id="batch.job.registry" class = "org.springframework.batch.core.configuration.support.MapJobRegistry">
    </bean>
```

**Expected behavior**

no warning

**Minimal Complete Reproducible example**
Please provide a failing test or a [minimal complete verifiable example](https://github.com/spring-projects/spring-batch/blob/main/ISSUE_REPORTING.md) that reproduces the issue.
Bug reports that are reproducible will take priority in resolution over reports that are not reproducible.

mp911de:
Duplicate of https://github.com/spring-projects/spring-data-jpa/issues/3244 and https://github.com/spring-projects/spring-boot/issues/38558.

We're not quite sure where this comes from.

hpoettker:
The sudden attention comes from the changed log level from `INFO` to `WARN`: https://github.com/spring-projects/spring-framework/issues/24092

[...]

PR #4520, opened 2023-12-17
Fix `BeanPostProcessorChecker` warnings

hpoettker:
This PR is a suggestion how #4519 could be addressed. It might also offer a solution to #4489, but I haven't really checked the case of XML configuration.

The severety of the logged warnings can be debated. On the one hand, they can probably be ignored by many users, so silencing them in the logging configuration of their applications is often viable. On the other hand, there may well be users who are negatively affected by the early bean initializations, and the more than 10 lines of warnings that are currently produced by a plain Spring Boot app fresh from the Spring Initializr are definitely detrimental to the Getting Started experience.

I think it would be great to have a fix with release 5.1.1 but I don't see a fix that is completely backwards compatible. The PR contains the most backwards compatible version that I could come up with. Unless a better solution is found, I think it should be considered for release 5.1.1 for the sake of the Getting Started experience.

The change is backwards compatible for users that use the respective methods on `DefaultBatchConfiguration` to only access the `JobRegistry` or the `JobRegistryBeanPostProcessor`, or don't use these methods at all. It is a breaking change for users that have overridden these methods to change the beans. These users would need to adapt and expose their replacements as beans with name `jobRegistry` or `jobRegistryBeanPostProcessor` respectively.

Another point to note is that declaring the `JobRegistry` as an infrastructure bean is a bit of a gray area. Many end-users may not be aware of it but there are most likely some who directly use it.

The result of the change is that only the `MapJobRegistry` is initialized early as it is required by the `JobRegistryBeanPostProcessor`, and no more warnings are logged by a plain Spring Boot app (or any other similar app that uses `DefaultBatchConfiguration`).

Please let me know if you are interested in the PR. I'll mark it as a draft until then. It's mostly fine, but I think some additional tests, a polish of the Javadoc, and an edit of the reference documentation would be nice.

Please feel free to close the PR, in case you want to pursue a different direction.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/DefaultBatchConfiguration.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/DefaultBatchConfiguration.java
@@ -112,13 +112,18 @@ import org.springframework.transaction.annotation.Isolation;
  * @since 5.0
  */
 @Configuration(proxyBeanMethods = false)
-@Import(ScopeConfiguration.class)
+@Import({ ScopeConfiguration.class, BatchBeanDefinitionRegistryPostProcessor.class })
 public class DefaultBatchConfiguration implements ApplicationContextAware {
 
 	@Autowired
 	protected ApplicationContext applicationContext;
 
-	private final JobRegistry jobRegistry = new MapJobRegistry();
+	@Autowired
+	protected JobRegistry jobRegistry;
+
+	@Autowired
+	@Deprecated(forRemoval = true)
+	private JobRegistryBeanPostProcessor jobRegistryBeanPostProcessor;
 
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
@@ -187,9 +192,17 @@ public class DefaultBatchConfiguration implements ApplicationContextAware {
 		}
 	}
 
-	@Bean
+	/**
+	 * Defines a {@link JobRegistry} bean.
+	 * @return a {@link JobRegistry} bean
+	 * @throws BatchConfigurationException if unable to register the bean
+	 * @since 5.0
+	 * @deprecated in favor of bean injection, and declaring a custom bean with name
+	 * {@code jobRegistry}
+	 */
+	@Deprecated(forRemoval = true)
 	public JobRegistry jobRegistry() throws BatchConfigurationException {
-		return this.jobRegistry; // FIXME returning a new instance here does not work
+		return this.jobRegistry;
 	}
 
 	@Bean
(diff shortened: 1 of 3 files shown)


hpoettker:
An alternative would be to add a `JobRegistrySmartInitializingSingleton` that does the same thing as `JobRegistryBeanPostProcessor` but implements `SmartInitializingSingleton` instead of `BeanPostProcessor`.

The benefit of the `JobRegistrySmartInitializingSingleton` is that it is initialized like any other bean, but that `SmartInitializingSingleton::afterSingletonsInstantiated` is only called after all job beans have been initialized. This way, the `JobRegistry` can be injected without declaring it as an infrastructure bean and no early initialization of any beans that the `DefaultBatchConfiguration` depends on are triggered.

The `JobRegistrySmartInitializingSingleton` could then be exposed as a bean much more safely than the `JobRegistryBeanPostProcessor`. Such a change would also only break things added in 5.1.0., i.e. the automatic exposure of `JobRegistryBeanPostProcessor`.

This seems to work nicely. Let me know if you're interested. I'll then polish the code a bit and push it.

fmbenhassine:
Thank you for the PRs, @hpoettker ! It is unfortunate that the log level change in SF was silently unnoticed and that this issue was only caught after the release of Batch 5.1.. Anyway, those warnings definitely impact the getting started experience and I agree we should try to address them as early as 5.1.1 if possible.

Looking at your suggestions to solve the problem with a `BeanDefinitionRegistryPostProcessor` or  `JobRegistrySmartInitializingSingleton`, those are fine for a minor release but due to the breaking changes they introduce, we typically try to avoid such changes in patch releases. Have you tried to move the job registry and the BPP to a separate configuration class (similar to `ScopeConfiguration`)?. This should also fix the problem but I am not sure it will be backward compatible for a patch release. I will try on my side as well and share my feedback here.

hpoettker:
As far as I can tell, there is no backwards compatible way to address the issue, which is unfortunate with 5.2.0 only scheduled for the end of the year. But it should definitely be fixed, not only because of the warnings but also because the early bean initializations can cause difficult to track issues for users.

Moving the `MapJobRegistry` and the post processor to a different configuration class wouldn't really help, I think. You'd still get the warning for the `JobRegistry` or the new configuration class (depending on the implementation). Declaring the bean factory methods for the `JobRegistry` and the post processor as `static` would be the same thing.

The best way forward in my opinion would be to revert the unconditional exposure of the `BeanDefinitionRegistryPostProcessor` and offer the `JobRegistrySmartInitializingSingleton` (or a similar class) as an alternative. This should affect the smallest number of users and provide a solution for those that have been using the `BeanDefinitionRegistryPostProcessor` before the 5.1.0.

spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/DefaultBatchConfiguration.java
```diff
@@ -187,9 +192,17 @@ public JobExplorer jobExplorer() throws BatchConfigurationException {
...
 	}
 
-	@Bean
+	/**
+	 * Defines a {@link JobRegistry} bean.
+	 * @return a {@link JobRegistry} bean
+	 * @throws BatchConfigurationException if unable to register the bean
+	 * @since 5.0
+	 * @deprecated in favor of bean injection, and declaring a custom bean with name
+	 * {@code jobRegistry}
```
fmbenhassine, line 201: I am trying to test this from a user's point of view (which is also what is mentioned in the [docs](https://docs.spring.io/spring-batch/reference/job/advanced-meta-data.html#jobregistry)) against this PR but it fails (while it passes on the current `main`). Here is a patch that slightly modifies the canonical example I used for tests: 

[test_pr4520.patch](https://github.com/spring-projects/spring-batch/files/13991872/test_pr4520.patch)

fmbenhassine:
After digging into this, it seems that having dependencies in BPP is not recommended and might cause issues in SF. Here is the excerpt from the [SF docs](https://docs.spring.io/spring-framework/reference/core/beans/factory-extension.html#beans-factory-extension-bpp) I am referring to:

```
If you have beans wired into your BeanPostProcessor by using autowiring or @Resource (which may
fall back to autowiring), Spring might access unexpected beans when searching for type-matching
dependency candidates and, therefore, make them ineligible for auto-proxying or other kinds of
bean post-processing
```

So I think the mechanism of using a BPP to populate the registry has always been problematic in Spring Batch, and only made visible now after the recent log level change in SF. BTW, the same issue happens with XML config, as well as `@EnableBatchProcessing`. Therefore, I believe it is a good time to review this mechanism and replace it with a smart initializing singleton like in #4521. I am closing this PR in favor of #4521. Thank you for the suggestion!

PR #4521, opened 2023-12-18
Add `JobRegistrySmartInitializingSingleton`

hpoettker:
This PR adds `JobRegistrySmartInitializingSingleton` as an alternative to `JobRegistryBeanPostProcessor`. Using `JobRegistrySmartInitializingSingleton` instead of `JobRegistryBeanPostProcessor` allows users to solve problems like #4489, where  they configure their own application context and want a mechanism to automatically register job beans with the job registry.

As also discussed in #4520, the advantage of `JobRegistrySmartInitializingSingleton` over `JobRegistryBeanPostProcessor` is that it is not prone to trigger early initializations of beans.

One other difference is that `JobRegistrySmartInitializingSingleton` will instantiate all job beans that are declared as lazy (unless the property `allowEagerClassLoading` of `DefaultListableBeanFactory` is set to false). This shouldn't be a problem however as users who do use the `JobRegistry` are very unlikely to have lazily initialized job beans as that would imply that the `JobRegistry` is incomplete.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JobRegistrySmartInitializingSingleton.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JobRegistrySmartInitializingSingleton.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


fmbenhassine:
LGTM. Rebased and merged as fe26d007440f2582f7a4e5e9672f29dc68cf6026, I just refined some tests and updated the reference docs (see d6b6361e597b51c9262c8c1c9a3b5bbdb894ed57).

I couldn't find a backward compatible fix, but this will be available in 5.1.1 as a replacement of the BPP. I planned the switch to the SIS for 5.2 (#4547).

Again, thank you very much for your continuous contributions, really appreciated!

PR #4522, opened 2023-12-22
Fix StepExecution.getEndTime() returns null

### LINKED ISSUE #3846: StepExecution.getEndTime() returns null in StepExecutionListener.afterStep()

baezzys:
This PR resolves #3846

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java
@@ -63,6 +63,7 @@ import org.springframework.util.ClassUtils;
  * @author Michael Minella
  * @author Chris Schaefer
  * @author Mahmoud Ben Hassine
+ * @author Jinwoo Bae
  */
 public abstract class AbstractStep implements Step, InitializingBean, BeanNameAware {
 
@@ -261,7 +262,13 @@ public abstract class AbstractStep implements Step, InitializingBean, BeanNameAw
 			}
 		}
 		finally {
-
+			stepExecution.setEndTime(LocalDateTime.now());
+			Duration stepExecutionDuration = BatchMetrics.calculateDuration(stepExecution.getStartTime(),
+					stepExecution.getEndTime());
+			if (logger.isInfoEnabled()) {
+				logger.info("Step: [" + stepExecution.getStepName() + "] executed in "
+						+ BatchMetrics.formatDuration(stepExecutionDuration));
+			}
 			try {
 				// Update the step execution to the latest known value so the
 				// listeners can act on it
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java
```diff
@@ -287,14 +294,8 @@ public final void execute(StepExecution stepExecution)
 						name, stepExecution.getJobExecution().getJobInstance().getJobName()), e);
 			}
 			stopObservation(stepExecution, observation);
```
fmbenhassine, line 296: I think this should also be moved before the listener. Otherwise there will be a difference between the calculated stepExecutionDuration and the micrometer metric. I will take care of that change on merge.

fmbenhassine: I tried this but it breaks the step status in the metric. I will leave it as is.

fmbenhassine:
Rebased and merged as df5e8e1686ac27e0d2ffa6131bb2e515d0e40ed5. I just updated the test to use a custom listener (see 6b79e2c1f756bf87416baaf424813c84db8755b7).

PR #4525, opened 2023-12-28
Fix `SimpleBinaryBufferedReaderFactory` for longer endings

### LINKED ISSUE #811: SimpleBinaryBufferedReaderFactory misses line endings [BATCH-2792]
**[cosban](https://jira.spring.io/secure/ViewProfile.jspa?name=cosban)** opened **[BATCH-2792](https://jira.spring.io/browse/BATCH-2792?redirect=false)** and commented

If a line contains the first character of the line delimiter within N characters of the actual beginning of the line ending delimiter, where N is the length of the line ending delimiter, then the SimpleBinaryBufferedReaderFactory will incorrectly miss the actual line ending and continue reading.

The '#' character is within 5 characters of the actual line ending for this example. This causes the candidateEnding to contain `#3#@#` before determining that it is not the correct line ending, appending all of these characters to the line buffer, and continuing.
I suspect it would be better for the method to fail fast the moment that current candidate doesn't match the provided template. i.e. once `#3` is determined to not be a possible match of the template going forward.

I have provided a full example of a situation which causes this to fail below:

```
EXAMPLE OF BUG: lineEnding = `#@#@#`, lines = `Value_1,Value_2,Value#3#@#@#Value_4,Value_5,Value_6`

BAD OUTPUT: `Value_1` `Value_2` `Value#3#@#@#Value_4` `Value_5` `Value_6`

CORRECT OUTPUT: `Value_1` `Value_2` `Value#3`
                `Value_4` `Value_5` `Value_6`
```

---
No further details from [BATCH-2792](https://jira.spring.io/browse/BATCH-2792?redirect=false)

hpoettker:
Resolves #811

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/SimpleBinaryBufferedReaderFactory.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/SimpleBinaryBufferedReaderFactory.java
@@ -123,25 +123,24 @@ public class SimpleBinaryBufferedReaderFactory implements BufferedReaderFactory 
 			}
 
 			char c = (char) next;
-			if (ending.charAt(0) == c || candidate.length() > 0) {
+			if (ending.charAt(0) == c || !candidate.isEmpty()) {
 				candidate.append(c);
 			}
-
-			if (candidate.length() == 0) {
+			else {
 				buffer.append(c);
 				return false;
 			}
 
-			boolean end = ending.equals(candidate.toString());
-			if (end) {
+			if (ending.contentEquals(candidate)) {
 				candidate.delete(0, candidate.length());
+				return true;
 			}
-			else if (candidate.length() >= ending.length()) {
-				buffer.append(candidate);
-				candidate.delete(0, candidate.length());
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/SimpleBinaryBufferedReaderFactory.java
```diff
@@ -123,25 +123,24 @@ private boolean isEndOfLine(StringBuilder buffer, StringBuilder candidate, int n
...
 			}
 
-			boolean end = ending.equals(candidate.toString());
-			if (end) {
+			if (ending.contentEquals(candidate)) {
 				candidate.delete(0, candidate.length());
+				return true;
 			}
-			else if (candidate.length() >= ending.length()) {
+			if (!ending.startsWith(candidate.toString())) {
```
hpoettker, line 138: The change in this line is the only "actual" change. The other changes are only intended to clarify the logic.

fmbenhassine:
LGTM. Rebased and merged as 929a5514421eef7c4a388dfde3dd13284e8c82c7. Thank you for your contribution!

PR #4529, opened 2024-01-05
Fixed java-doc referencing Long when it should be Double

### LINKED ISSUE #4526: Incorrect Type of getter in Documentation for JobParameters GetDouble
Hi I was looking through the Spring Batch API for the Class JobParameters. In the documentation I noticed there was a slight error in the description for getDouble(String key) in which the Description states. 

Typesafe getter for the [Long](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Long.html) represented by the provided key.

I looked into it a bit and saw that it has to do with a comment in the class JobParameters.

https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java

In which the following comment is made.

```
	/**
	 * Typesafe getter for the {@link Long} represented by the provided key.
	 * @param key The key for which to get a value.
	 * @return The {@link Double} value or {@code null} if the key is absent.
	 */
```

The getDouble(String key, @Nullable Double defaultValue) seems to have it set to double which looking at the code I think is because of a spelling error. 

```
	/**
	 * Typesafe getter for the {@link Double} represented by the provided key. If the key
	 * does not exist, the default value is returned.
	 * @param key The key for which to return the value.
	 * @param defaultValue The default value to return if the value does not exist.
	 * @return the parameter represented by the provided key or, if that is missing, the
	 * default value.
	 */
```

I'm not sure if this is a big priority for you guys but I thought this could help.

NotBjoggisAtAll:
This will resolve #4526

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/JobParameters.java
@@ -141,7 +141,7 @@ public class JobParameters implements Serializable {
 	}
 
 	/**
-	 * Typesafe getter for the {@link Long} represented by the provided key.
+	 * Typesafe getter for the {@link Double} represented by the provided key.
 	 * @param key The key for which to get a value.
 	 * @return The {@link Double} value or {@code null} if the key is absent.
 	 */


NotBjoggisAtAll:
@pivotal-cla This is an Obvious Fix

fmbenhassine:
Merged as aa608246c14b9cde0320eea97e62b5e431cf66c3. Thanks!

PR #4541, opened 2024-02-02
Fix dependency injection style in DefaultBatchConfiguration

fmbenhassine:
This commit fixes the dependency injection style to be compatible with the currently disabled bean method proxying.


fmbenhassine:
Updated and merged as c61670bb81a6c91ba6c87db5b603f057421a152c.

PR #4548, opened 2024-02-07
Reduce formatter initializations in `DefaultFieldSet`

### LINKED ISSUE #1694: DefaultFieldSet creates a new SimpleDateFormat as a class member; newer JVMs cause SDF to get stuck in TimeZone.getDefaultInAppContext which is slow
**[Brian Tarbox](https://jira.spring.io/secure/ViewProfile.jspa?name=briantarbox)** opened **[BATCH-1902](https://jira.spring.io/browse/BATCH-1902?redirect=false)** and commented

DefaultFieldSet creates a SimpleDateFormat (SDF) object as a member variable.  This means it can not be overridden.  SDF eventually calls TimeZone which calls getDefaultInAppContext.  This is a static, synchronized and very slow method.  This results in extremely slow reads.

To fix the problem in my own code means I have to basically make a copy of the entire DefaultFieldSet class.  If the SDF were injected then I could change the slow behavior without having to copy the whole class.

I spoke with Gary Gregory (Spring Batch In Action) and he liked the idea of this change.

---

**Affects:** 2.1.8

**Reference URL:** http://stackoverflow.com/questions/12984345/java-7-calendar-getinstance-timezone-gettimezone-got-synchronized-and-slow-any

0 votes, 5 watchers

hpoettker:
Resolves #1694.

The PR addresses the problem that the default value for the field `numberFormat` of `DefaultFieldSet` is currently initialized twice: Once by the assignment in the declaration and once in the constructor. As the construction of the default number format is expensive, this should be avoided.

It also addresses the problem that the expensive construction of the default `numberFormat` and `dateFormat` currently cannot be avoided in the case where the default values are overwritten after calling the constructor.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DefaultFieldSet.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/transform/DefaultFieldSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM. Rebased and merged as 75abf1587ff1bf1043aa1b743ac2ad26609219fc. Thank you for your contribution!

PR #4550, opened 2024-02-15
Incorrect code example in documentation

chldppwls12:
- Message in `Assert.state(dir.isDirectory())` was missing.
- Message has been changed to start with a capital letter, consistent with other parts.

### DIFF
--- a/spring-batch-docs/modules/ROOT/pages/step/tasklet.adoc
+++ b/spring-batch-docs/modules/ROOT/pages/step/tasklet.adoc
@@ -122,7 +122,7 @@ public class FileDeletingTasklet implements Tasklet, InitializingBean {
     public RepeatStatus execute(StepContribution contribution,
                                 ChunkContext chunkContext) throws Exception {
         File dir = directory.getFile();
-        Assert.state(dir.isDirectory());
+        Assert.state(dir.isDirectory(), "Directory does not exist");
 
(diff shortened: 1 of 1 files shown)


spring-batch-docs/modules/ROOT/pages/step/tasklet.adoc
```diff
@@ -122,7 +122,7 @@ public class FileDeletingTasklet implements Tasklet, InitializingBean {
     public RepeatStatus execute(StepContribution contribution,
                                 ChunkContext chunkContext) throws Exception {
         File dir = directory.getFile();
-        Assert.state(dir.isDirectory());
+        Assert.state(dir.isDirectory(), "Directory does not exist");
```
fmbenhassine, line 125: Good catch! The message is missing and the code example does not compile. However, the message here is inaccurate, it would be appropriate if the assertion was `dir.exists()`. It should rather be something like "The resource must be a directory". Do you agree?

chldppwls12: Yes. I updated message. Thanks:)

fmbenhassine:
Thank you for the update! LGTM now 👍 

Rebased and merged as 6334fbf870a4a3b9b18bf792268861bbe87f471c. Thank you for your contribution!

PR #4561, opened 2024-03-04
Fix SimpleChunkProcessor transform output chunk end property

### LINKED ISSUE #4560: Incorrect Chunk property value in implementation of ItemWriter write method
**Bug description**
I was implementing an ItemWriter class by overriding write method. I wanted to add a condition to do if the chunk is the last one. I saw that there is a `isEnd()` method in itself. But it never comes true even if the chunk is the last chunk.

I debugged the spring batch classes. `end` property of the actual chunk object is true when it is the last chunk. However, the chunk object are reproduced by the actual chunk. The reproduced chunk passing to ItemWrite write method. It does not preserve variables such as `end`.

I also opened a pull request for this problem to fix. A test case is included in PR.

**Environment**
Java Version: 21 (openjdk-21.0.2)
Spring Batch Core: 5.0.0 (also tried 5.2.0)
Spring Batch Infrastructure: 5.1.0

**Steps to reproduce**

1. Create a Step with custom ItemWriter implementation.
2. Give the chunk size of the step as 10.
3. Launch the job with 10-20 items (for example 16).
4. Print the `chunk.isEnd()` in the write method of the implementation.

**Expected behavior**
In the last call of write method, `chunk.isEnd()` should be true.

```
public class ConditionalItemWriter implements ItemWriter<String> {

  @Override
  public void write(Chunk<? extends String> chunk) {
    if (chunk.isEnd()) {
      System.out.println("This was the last chunk!");
    }
   }
  }
```

myanar7:
This PR closes #4560.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/FaultTolerantChunkProcessor.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/FaultTolerantChunkProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -306,7 +306,9 @@ public class FaultTolerantChunkProcessor<I, O> extends SimpleChunkProcessor<I, O
 				break;
 			}
 		}
-
+		if (inputs.isEnd()) {
+			outputs.setEnd();
+		}
 		return outputs;
 
 	}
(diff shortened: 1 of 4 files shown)


fmbenhassine:
Thank you for the PR. The change should be applied to `FaultTolerantChunkProcessor` as well, as it overrides `SimpleChunkProcessor#transform`.

My concern regarding the edge case mentioned in https://github.com/spring-projects/spring-batch/issues/4560#issuecomment-1983295279 is that it makes the method unreliable. I am not sure there is an obvious way to fix this with the current implementation, but that's probably for another issue.

myanar7:
Thank you. The change has been implemented to FaultTolerantChunkProcessor as well. Additionally, the edge case in Chunk#isEnd has been briefly mentioned in the javadoc of the method.

fmbenhassine:
Thank you for the update. LGTM 👍  Rebased and merged as a4a8ab7aaa8e3ead006157c08d845acfe9993e85. I just amended a test in `FaultTolerantChunkProcessorTests`.

Issue #3939, opened 2021-06-13
Different behavior when interrupting a job, depending on the parallel flow order


kkmikhaylova:
**Bug description**
Different behavior when interrupting a job, depending on the parallel flow order.
I am creating a parallel flow and trying to interrupt a job from one of them. Further behavior depends on the order of the flow inside the add() function. In one case, the job is finished immediately, in the other, all steps from another parallel flow are executed.

`Flow parallelFlow = new FlowBuilder<SimpleFlow>("parallelFlow")`
`	.split(new SimpleAsyncTaskExecutor())`
`	.add(nonInterruptingFlow, interruptingFlow)`
`	.build();`

**Environment**
Spring Batch: 4.3.1

**Steps to reproduce**
1. Create a job with two parallel flows, where the last flow in the list interrupts the job.
2. Make sure that the job has finished after the interruption.
3. Change flow order (now the first flow in the list interrupts the job).
4. Make sure that all the steps of the job have been executed after the interruption.

**Expected behavior**
Same behavior after interruption.

**Minimal Complete Reproducible example**
[springbatchdemo.zip](https://github.com/spring-projects/spring-batch/files/6644341/springbatchdemo.zip)

fmbenhassine:
Thank you for opening this issue and for providing a minimal sample. Nicely done!

Both jobs complete as expected with a `STOPPED` status, but indeed the `finishedJob` proceeds to the `nonExecutableStep` even though the flow was interrupted in the previous step. That should not be the case. Since those flows are running in parallel, the interruption order should not matter.

doontagi:
Hi @fmbenhassine, I checked this issue. And I've made a fix PR https://github.com/spring-projects/spring-batch/pull/4567.
PTAL 🙏

PR #4567, opened 2024-03-24
Fix unfinished step in parallel flow

doontagi:
related issue #3939 

## Motivation
- On https://github.com/spring-projects/spring-batch/issues/3939, when two flows are executed in parallel, I found that even though one flow throw an interruption and terminated the job, the next step of the other flow was executed.
- And this abnormal behavior occurs only when the other flow was added secondly to the job
- Problem is that even if the job containing both flows were terminated and went into STOPPED status, the other flow couldn't recognize the interruption because interruption had been checked by STOPPING status.
https://github.com/spring-projects/spring-batch/blob/9a2b1bbc96f338017aaa5df51214896f5a32d1e5/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java#L290C16-L290C17
Because the job already had been terminated and in STOPPED status, a step in the other flow cannot recognize it and proceeded as like there had been no interruption.
- The reason why this behavior didn't occur when the interruptedFlow added secondly to the job is that secondly added flow was always executed secondly in SplitState. So, if the two step flow would be added firstly, first step in this flow will be processed before the interrupted flow made the job status STOPPED, so it can recognize the interruption when one step has been executed.
- If there are remaining tasks (steps) but the job was terminated due to an exception as like now, there is also a problem that the status of the job may be changed by the remaining tasks.

## Modification
- Changed to wait for all tasks to finish before throwing an exception.
- Changed a job to handle exceptions after waiting for tasks currently being processed to complete. Prevent the job from terminating even though there are remaining tasks.

## Result
- Job, Flow and Steps work same when an exception does not occur.
- Regardless of the order in which flows are added to the Job for parallel processing, if an interruption occurs, only the currently executing step is processed. That is, no further steps are executed after that.
- Close https://github.com/spring-projects/spring-batch/issues/3939

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/job/builder/FlowJobBuilderTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/job/builder/FlowJobBuilderTests.java
@@ -16,11 +16,14 @@
 package org.springframework.batch.core.job.builder;
 
 import java.util.Arrays;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 
 import javax.sql.DataSource;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.batch.core.BatchStatus;
@@ -45,6 +48,8 @@ import org.springframework.batch.core.repository.support.JobRepositoryFactoryBea
 import org.springframework.batch.core.step.StepSupport;
 import org.springframework.batch.core.step.builder.StepBuilder;
 import org.springframework.batch.item.support.ListItemReader;
+import org.springframework.batch.repeat.RepeatStatus;
+import org.springframework.batch.support.transaction.ResourcelessTransactionManager;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/test/java/org/springframework/batch/core/job/builder/FlowJobBuilderTests.java
```diff
@@ -419,4 +424,38 @@ public JdbcTransactionManager transactionManager(DataSource dataSource) {
...
+		Flow interruptedFlow = new FlowBuilder<SimpleFlow>("interruptedFlow").start(interruptedStep).build();
+
+		Flow splitFlow = new FlowBuilder<Flow>("splitFlow").split(new SimpleAsyncTaskExecutor())
+			.add(interruptedFlow, twoStepFlow)
+			.build();
+		FlowJobBuilder jobBuilder = new JobBuilder("job", jobRepository).start(splitFlow).build();
+		jobBuilder.preventRestart().build().execute(execution);
+
+		boolean isExecutedNonExecutableStep = countDownLatch.await(1, TimeUnit.SECONDS);
+		assertEquals(BatchStatus.STOPPED, execution.getStatus());
```
doontagi, line 457: Finished with `BatchStatus.STOPPED` without executing `nonExecutableStep`.
This test would fail with executing `nonExecutableStep` if it runs on the current branch.

doontagi:
Gentle ping to @fmbenhassine. Could you review this PR? I've fixed https://github.com/spring-projects/spring-batch/issues/3939.
Thanks!

fmbenhassine:
Thank you for your PR!

> Changed to wait for all asynchronous tasks to finish before throwing an exception.

Isn't that not desired when one of the tasks is interrupted? Meaning if one wants to interrupt a step, the idea is not to wait for other steps to finish.

The change in this PR fixes the reported issue, I just want to make sure there is no side effect on the behaviour of step interruption.

doontagi:
Thank you for the comment! And sorry for the late reply. @fmbenhassine 

I've also thought about your concern. And I don't think this change of exception processing location affects the current process of handling interruptions in steps, so has no side effect. 
Because even if an interruption occurs in one of the steps being processed at the same time, the other step may recognize the interruption only after its execution is finished.
https://github.com/spring-projects/spring-batch/blob/425134c74e6b69dde40a81278e4fddbec4778f80/spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java#L230
Interruption only can be recognized in catch or final blocks after `doExecute` has been done.

Therefore, even if interrupt occurs in a step, other steps are processed until the end regardless of my modification.

The difference is that the current code immediately propagates the exception generated by interrupt and terminates the job (`STOPPED`) without caring of other steps. Therefore, the remaining flows and steps cannot recognize the interrupt because it recognize it by `STOPPING` status of the job https://github.com/spring-projects/spring-batch/blob/425134c74e6b69dde40a81278e4fddbec4778f80/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java#L287-L294
and continue to be executed. It is a cause of the bug.

In conclusion, there is no change from the view of steps because in the current code, the execution is also not stopped if an exception caused by interrupt occurs in another step. And I think `Job`  shouldn't be stopped if there are running flows and steps.

So I think my changes will prevent other unknown bugs that can occur as flow and step continue to run with job stopped.

Sorry for the lengthy explanation, please comment if you have any other opinions!

doontagi:
Gentle remind @fmbenhassine. Could you check this PR?
Thanks!

fmbenhassine:
Thank you for your feedback! I saw your comment back then and added a thumbs-up emoji and the `status: feedback-provided` tag to acknowledge your update. I should have mentioned the release in which i was planning to include this to avoid you waiting for an update from my side. Anyway, this looks good to me and should go into 5.2.1 planned for tomorrow.

Thank you for going into details with your last update! I love how you explained the step interruption process and how this change should not have a side effect 👍

fmbenhassine:
Rebased and merged as 76789494d7be2f6d658763039046c0cea80a1fe1. Thank you for your contribution!

PR #4582, opened 2024-04-24
Fix code sample that uses deprecated StepBuilderFactory

Seungpang:
I found a deprecated StepBuilderFactory while looking at the documentation and tried to modify it


fmbenhassine:
Good catch! Merged as b020da4766d306950cc6141a5eeb8fffcd60dd6b. Thank you for the PR.

Issue #4555, opened 2024-02-25
DELETE CASCADE on Foreign Keys


hannosgit:
### Current Behaviour
The current implementation to delete job instances or job executions issues a separate SQL statement for each metadata table. (see https://github.com/spring-projects/spring-batch/pull/4497)

### Suggested Change
If we define the  `DELETE CASCADE` option on each Foreign Key in the metadata table, we could remove all metadata by simply deleting the desired table row and the database will take care of the rest.

hannosgit:
@mayleaf Are you sure that all databases support the `DELETE CASCADE` option? For example, I am not sure if Sybase does.

mayleaf:
As far as I had found, there was ON DELETE option. But I confused ASE and IQ. As you said, It seems doesn't support ON DELETE on sybase.  Then how about add delete trigger for those DBMSs that doesn't support delete cascade? @hannosgit @fmbenhassine

hannosgit:
@mayleaf One option would be to create a new separate method in `JobRepository` that only deletes the job instance and lets the DB do the rest.  Developers that use DBs that do not support `ON DELETE CASCADE` and developers that do not define the `DELETE CASCADE` can use the existing methods that delete the data on the Java side. All other developers can use the new method.

mayleaf:
@hannosgit There is already a JobRepository.deleteJobInstance() method. I thought the positive effect of this PR was that it simplifies the SimpleJobRepository implementation, not changes the JobRepository interface. Because there is no need to create a duplicated signature. If I misunderstand your opinion, please let me know 🙇

hannosgit:
Yeah that was the idea but if we cannot come up with a solution for these DBs, we have to live with some kind of workaround such as the one I suggested. I have no experience/access with e.g. Sybase, so I cannot help with this. I found this [workaround](https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc32300.1570/html/sqlug/X15877.htm) of using triggers to simulate `ON DELETE CASCADE` but I cannot verify it.

mayleaf:
![Screenshot 2024-10-20 at 8 22 42 PM](https://github.com/user-attachments/assets/b9b720a4-ffe1-4753-a4a9-61cd652d325c)
https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.dc32300_1251/html/sqlug/X15877.htm
@hannosgit 
[...]

PR #4584, opened 2024-05-05
Add delete cascade on metadata tables' foreign keys

mayleaf:
Solved #4555

* Motivation
On https://github.com/spring-projects/spring-batch/issues/4555, We found delete methods in SimpleJobRepository are removing with separated SQL queries for each metadata tables
* Modification
Add ON DELETE CASCADE to each metadata tables' foreign keys.


spring-batch-test/src/main/java/org/springframework/batch/test/JobRepositoryTestUtils.java
```diff
@@ -135,8 +135,17 @@ public void removeJobExecutions(Collection<JobExecution> jobExecutions) {
 		for (JobExecution jobExecution : jobExecutions) {
```
mayleaf, line 135: Can I delete this method since it's no longer used after this pull request?

Issue #4586, opened 2024-05-05
Add schema locations to DatabaseType to prevent typos in the metadata tables string path


spring0913:
From the [#4252](https://github.com/spring-projects/spring-batch/issues/4252), I realized that I had to use addScript to create the metadata table when using an In-memory database in Spring Boot 3.0 / Spring Batch 5.0.

I think it is better to use static variables to prevent typos in the metadata table creation script path string.

After I create PR about this(schema-h2), if it is approved, I would like to add static variables to the other databases as well.

**AS-IS**
``` java
@Bean
public DataSource dataSource() {
    return new EmbeddedDatabaseBuilder()
		.addScript("/org/springframework/batch/core/schema-h2.sql")
		.build();
}
```

**TO-BE**
``` java
@Bean
public DataSource dataSource() {
    return new EmbeddedDatabaseBuilder()
		.addScript(DatabaseResource.SCHEMA_H2)
		.build();
}
```

injae-kim:
Fix PR: https://github.com/spring-projects/spring-batch/pull/4587

fmbenhassine:
> **TO-BE**

```java
@Bean
public DataSource dataSource() {
    return new EmbeddedDatabaseBuilder()
		.addScript(DatabaseResource.SCHEMA_H2)
		.build();
}
```

Would it make sense to add this information in the existing `DatabaseType` enum instead of creating a new one `DatabaseResource`? This would make the information about supported databases and their corresponding schemas centralized in one place (which is easier to use and maintain). I am thinking about a new field named `productSchema` next to `productName` in `DatabaseType`.  What do you think?

PR #4587, opened 2024-05-05
Add static variable for metadata table string path

spring0913:
Issue #4586

### Motivation
```java
@Bean
		.addScript("/org/springframework/batch/core/schema-h2.sql")
		.build();
}
```

- On https://github.com/spring-projects/spring-batch/issues/4586, I think it is better to use static variables to prevent typos in the metadata table creation script path string.

### Modification
```java
@Bean
		.addScript(DatabaseResource.SCHEMA_H2)
		.build();
}
```
- Add static variable for metadata table string path.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/aot/CoreRuntimeHints.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/aot/CoreRuntimeHints.java
@@ -57,6 +57,7 @@ import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.StepContribution;
 import org.springframework.batch.core.StepExecution;
+import org.springframework.batch.core.resource.DatabaseResource;
 import org.springframework.batch.core.scope.context.JobContext;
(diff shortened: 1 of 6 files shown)


haiderRizvii (approved):

fmbenhassine:
Thanks for picking this up! As explained in https://github.com/spring-projects/spring-batch/issues/4586#issuecomment-3532348451, I'd like to go with a slightly different approach: adding `productSchema`/`productSchemaDrop` fields to the existing `DatabaseType` enum (in `spring-batch-infrastructure`) rather than a new `DatabaseResource` class in `spring-batch-core`. This keeps everything about a supported database (name + schema locations) in one place, usable from both `spring-batch-core` and `spring-batch-integration`.

Also, `DatabaseResource.SCHEMA_H2` has a leading /, which breaks when used in `CoreRuntimeHints.registerPattern(...)` (AOT resource hints need classpath-relative paths, no leading /) — that would have silently broken the H2 native-image hint.

Given the scope of consolidating this across all supported databases and call sites, I'll close this PR and take care of the changes myself.  Thanks again for raising it!

PR #4595, opened 2024-05-15
Fix detail link

LichKing-lee:
old link was broken
fix to new link


fmbenhassine:
Good catch! Rebased and merged as 67b7f380b450def0704064d0ff4cf423564ae1ea. Thank you for your contribution!

PR #4530, opened 2024-01-07
Poll the count of running step executions (updated)

pstetsuk:
Fix for #3790 (updated version of #3791)

PR #4599, opened 2024-05-19
Keep heap lean during remote polling

### LINKED ISSUE #4598: Memory leak in MessageChannelPartitionHandler when polling the database
**Bug description**
In remote partitioning jobs which use the `MessageChannelPartitionHandler` with database polling, on each poll of the database where there is 1 or more new completed worker `StepExecutions`, the partition handler loads and keeps in memory an additional copy of the corresponding `JobInstance`, `JobExecution`, and all `StepExecutions` and their `ExecutionContexts` until the partition step is completed.

This leads to high memory consumption during the partition step and can lead to out of memory errors if the poll interval is short enough and the number of partitions is high enough, especially since the `ExecutionContexts` are held in memory as well.

**Environment**
Any environment using `spring-batch-integration` `5.0.1` and above (93800c6) which also uses the `MessageChannelPartitionHandler` with database polling.

**Steps to reproduce**
Run a remote partitioning batch job with database polling, a short poll interval, high number of partitions, and limited available memory.

**Expected behavior**
Polling of the database in remote partitioning jobs does not lead to constant gradual increase of consumed memory until the partition step completes.

Minimal Complete Reproducible example is here: [spring-batch-mcve-memory-leak.zip](https://github.com/spring-projects/spring-batch/files/15354433/spring-batch-mcve-memory-leak.zip)

The example runs a remote partitioning batch job with 1000 partitions, each having an `ExecutionContext` containing a single `UUID`. In order to exacerbate the memory consumption, the poll interval is set to very low (`2ms`), and the worker step sleeps for `50ms` before completing. This allows a new copy of the `JobInstance`, `JobExecution`, `StepExecutions`, and `ExecutionContexts` to be loaded and held in memory each time a worker step completes.

_**Please run the example with the `-Xmx64m -XX:+HeapDumpOnOutOfMemoryError` jvm options:**_
```
MAVEN_OPTS='-Xmx64m -XX:+HeapDumpOnOutOfMemoryError' mvn package exec:java -Dexec.mainClass=org.springframework.batch.MyBatchJobConfiguration
```
This should cause an `OutOfMemoryError` to be thrown rather quickly and the resulting heap dump should contain the following:
- 60-65 instances of `JobInstance`, `JobParameters`, `JobExecition`
- 60k-65k instances of `StepExecution`, `ExecutionContext`

**Analysis**
In `MessageChannelPartitionHandler#pollReplies`, the callback calls `JobExplorer#getJobExecution`. The `SimpleJobExplorer` implementation loads the `JobExecution` and all of the `StepExecutions` as well as their `ExecutionContexts`. Each of these `StepExecutions` also contains a reference to the `JobExecution` and thus to all other `StepExecutions` indirectly. If any of the loaded `StepExecutions` is completed and not present in the `result` Set, they are added to it, and this causes the currently loaded `JobExecution` instance and all of the other `StepExecution` instances to be held in memory.

hpoettker:
Resolves #4598.

The polling in `MessageChannelPartitionHandler` repeatedly fetches the running job execution with all associated step executions from the job repository. At the moment, it also retains references to step executions that have completed since the last iteration. This can accrue a lot of heap memory as each step execution holds references to a job execution, and that job execution holds references to all step executions.

This PR proposes not to retain any references to step or job executions while there are still running step executions. Instead, the step executions from the last iteration are used to build the result. This works as completed step executions are not expected to change after completion. And it reduces the memory foot print as the objects created in intermediate iterations can be handled by the garbage collector.

No test is added or changed as the changed code is already covered by appropriate tests.

### DIFF
--- a/spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/MessageChannelPartitionHandler.java
+++ b/spring-batch-integration/src/main/java/org/springframework/batch/integration/partition/MessageChannelPartitionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009-2023 the original author or authors.
+ * Copyright 2009-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -251,24 +251,22 @@ public class MessageChannelPartitionHandler extends AbstractPartitionHandler imp
 
 	private Set<StepExecution> pollReplies(final StepExecution managerStepExecution, final Set<StepExecution> split)
 			throws Exception {
-		final Set<StepExecution> result = new HashSet<>(split.size());
+		Set<Long> partitionStepExecutionIds = split.stream().map(StepExecution::getId).collect(Collectors.toSet());
 
 		Callable<Set<StepExecution>> callback = () -> {
-			Set<Long> currentStepExecutionIds = split.stream().map(StepExecution::getId).collect(Collectors.toSet());
 			JobExecution jobExecution = jobExplorer.getJobExecution(managerStepExecution.getJobExecutionId());
-			jobExecution.getStepExecutions()
+			Set<StepExecution> finishedStopExecutions = jobExecution.getStepExecutions()
 				.stream()
-				.filter(stepExecution -> currentStepExecutionIds.contains(stepExecution.getId()))
-				.filter(stepExecution -> !result.contains(stepExecution))
+				.filter(stepExecution -> partitionStepExecutionIds.contains(stepExecution.getId()))
 				.filter(stepExecution -> !stepExecution.getStatus().isRunning())
-				.forEach(result::add);
+				.collect(Collectors.toSet());
 
 			if (logger.isDebugEnabled()) {
 				logger.debug(String.format("Currently waiting on %s partitions to finish", split.size()));
 			}
 
-			if (result.size() == split.size()) {
-				return result;
+			if (finishedStopExecutions.size() == split.size()) {
+				return finishedStopExecutions;
 			}
 			else {
 				return null;


hpoettker:
@fmbenhassine Do you think this can be included in the 5.0/5.1 releases tomorrow?

It's backward-compatible and seems to solve the problem considering the reproducing example from #4598. I think it even fixes the problem in a way that requires no further refactoring of the polling. But effective feedback on that will probably only come when it's released in minor release.

fmbenhassine:
Thank you for the follow up on #4598 and for opening this PR!

Yes, the change LGTM and will be included in 5.0.6 and 5.1.2.

fmbenhassine:
Rebased and merged as 03a2b4d04938f06dcf6da10a4236b4671da4b389. Thank you!

hpoettker:
@adamsaghy I hope it's okay to tag you here.

You commented on #4530, and I assume you were interested due to your involvement in Apache Fineract. You recently upgraded Spring Batch there to 5.1.2 (https://github.com/apache/fineract/pull/3945), mainly for the change from this PR, if I understand correctly.

I would be very interested in feedback. Did this resolve the issue you were seeing? Was there no improvement at all? Or only some degree of improvement?

adamsaghy:
> @adamsaghy I hope it's okay to tag you here.
> 
> 
> 
> You commented on #4530, and I assume you were interested due to your involvement in Apache Fineract. You recently upgraded Spring Batch there to 5.1.2 (https://github.com/apache/fineract/pull/3945), mainly for the change from this PR, if I understand correctly.
> 
> 
> 
> I would be very interested in feedback. Did this resolve the issue you were seeing? Was there no improvement at all? Or only some degree of improvement?

Hi

We havent had the chance to battle test it yet but its on the todo list. i will let you know once it's done :)

PR #4601, opened 2024-05-22
Use Threadlocal.remove() instead of  Threadlocal.set(null)

Seungpang:
Hi Team,

This PR improves memory management by replacing instances of `ThreadLocal.set(null)` with `ThreadLocal.remove()` and removing the unnecessary `ThreadLocal.set(null)` in the constructor.

In the current implementation, `ThreadLocal.set(null)` is used to clear the `ThreadLocal` variable. This approach can leave the `ThreadLocal` variable in the thread's local variable map with a null value, potentially causing memory leaks. Using `ThreadLocal.remove()` instead properly removes the `ThreadLocal` variable from the thread's local variable map, preventing memory leaks.

Additionally, the constructor for `JobFlowExecutor` contains an unnecessary `ThreadLocal.set(null)`, which is redundant because the `ThreadLocal` variable is already initialized to `null` by default.


fmbenhassine:
LGTM. Rebased and merged as 3b9ceea0b443fd1c17fbd8250d95ae48257de0d6. Thank you for your contribution!

Issue #4605, opened 2024-05-29
Use '@EnableBatchProcessing' carefully on Spring Boot3.


kang-seongbeom:
When using '@EnableBatchProcessing' in Spring Boot3, default settings are initialized. Therefore, batch operation may not work.

fmbenhassine:
With Spring Boot 3, it is no longer required to use `@EnableBatchProcessing` to enable Spring Boot’s auto-configuration of Spring Batch. If you use that annotation, the auto-configuration of batch (including the automatic job execution at startup) will back-off. 

There is a section about this change in the migration guide of Spring Boot 3: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#spring-batch-changes . There is also a note as well in the getting started guide of Spring Batch: https://spring.io/guides/gs/batch-processing (See `Note how the application does not use the @EnableBatchProcessing annotation`..). Is this what you are looking for?

kang-seongbeom:
That's right. I was able to remove @EnableBatchProcessing by looking at "https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#spring-batch-changes". However, there is no warning about back-off in "https://docs.spring.io/spring-batch/reference/job/java-config.html". Therefore, I added a cautionary note to “https://github.com/spring-projects/spring-batch/pull/4606”!

fmbenhassine:
> However, there is no warning about back-off in "https://docs.spring.io/spring-batch/reference/job/java-config.html".

There is no such a note because this is a behaviour of Spring Boot. We do not document Spring Boot behaviour in Spring Batch docs. If you think docs can be improved, please open an issue or a PR on the boot side. I will close this issue and the corresponding PR, but As mentioned previously, there is a note about this in the migration guide of boot as well as the official docs: https://docs.spring.io/spring-boot/how-to/batch.html#howto.batch.specifying-a-data-source.

The only sample that uses Spring Boot on the batch side is the getting started guide, in which there is a note about that behaviour as well: https://spring.io/guides/gs/batch-processing (See `Note how the application does not use the @EnableBatchProcessing annotation`..).

PR #4606, opened 2024-05-29
Fix docs java configuration(#4605)

kang-seongbeom:
Solved spring-projects#4605

- Motivation
  There is nothing in the official document that says that using `@EnableBatchProcessing` in Spring Boot3 will initialize the configuration. I used `@EnableBatchProcessing` as explained in the official document, but I had a lot of trouble because the batch operation did not work.
- Modification
  Add phrase : When using @EnableBatchProcessing in Spring Boot3, default settings are initialized. Therefore, batch operation may not work.


fmbenhassine:
Closing this, as explained in https://github.com/spring-projects/spring-batch/issues/4605#issuecomment-2552052244

PR #4620, opened 2024-06-17
Fix the way of creating String for paging clause

pongdangx2:
Hi!!
I have been looking into MySqlQueryProvider these days.
And I found that StringBuilder is used for building Limit(or Top, RowNum) clause.

However, "+" operator is few times faster than StringBuilder, when concatenate just two or three Strings.

I wrote a simple code to compare execution of "+"operator and StringBuilder.
| try count | "+" Operation average execution time(ns) |StringBuilder average execution time(ns)|
|---|---|---|
|100|150.04|799.56|
|1000|99.588|369.835|
|10000|87.3189|269.9823|


pongdangx2:
I find that my sample code was wrong :(

PR #4625, opened 2024-06-23
fix: remove unnecessary null check

### LINKED ISSUE #4617: Is it correct for isNonDefaultExitStatus to return true when status is null?
I have a question regarding the behavior of the isNonDefaultExitStatus method in the [ExitStatus class](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/ExitStatus.java#L278). According to the current implementation, the method returns true when the provided status is null. Here is the relevant code:

```java
/**
 * @param status The {@code ExitStatus} object containing the exit code to be evaluated.
 * @return {@code true} if the value matches a known exit code.
 */
public static boolean isNonDefaultExitStatus(ExitStatus status) {
    return status == null || status.getExitCode() == null
            || status.getExitCode().equals(ExitStatus.COMPLETED.getExitCode())
            || status.getExitCode().equals(ExitStatus.EXECUTING.getExitCode())
            || status.getExitCode().equals(ExitStatus.FAILED.getExitCode())
            || status.getExitCode().equals(ExitStatus.NOOP.getExitCode())
            || status.getExitCode().equals(ExitStatus.STOPPED.getExitCode())
            || status.getExitCode().equals(ExitStatus.UNKNOWN.getExitCode());
}
```

Based on the method's purpose, which is to determine if the given ExitStatus matches a known exit code, I am wondering if returning true when status is null is the intended behavior. Intuitively, it seems that null should not be considered as a known exit code, and thus, the method might be expected to return false in such cases.

I referred to the [Spring Batch API documentation](https://docs.spring.io/spring-batch/docs/current/api/org/springframework/batch/core/ExitStatus.html#isNonDefaultExitStatus(org.springframework.batch.core.ExitStatus))

Can you tell me if the intended behavior is for the method to return true when the input is null? If null is also a known exit code, I wonder if we could add something to the documentation about it.

kariskan:
resolves #4617

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/ExitStatus.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/ExitStatus.java
@@ -275,8 +275,7 @@ public class ExitStatus implements Serializable, Comparable<ExitStatus> {
 	 * @return {@code true} if the value matches a known exit code.
 	 */
 	public static boolean isNonDefaultExitStatus(ExitStatus status) {
-		return status == null || status.getExitCode() == null
-				|| status.getExitCode().equals(ExitStatus.COMPLETED.getExitCode())
+		return status.getExitCode().equals(ExitStatus.COMPLETED.getExitCode())
 				|| status.getExitCode().equals(ExitStatus.EXECUTING.getExitCode())
 				|| status.getExitCode().equals(ExitStatus.FAILED.getExitCode())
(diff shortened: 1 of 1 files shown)


haiderRizvii (approved):

fmbenhassine:
Closing this as the corresponding issue https://github.com/spring-projects/spring-batch/issues/4617 was closed. Thank you for your time anyway.

Issue #4627, opened 2024-06-27
Add DatabaseType POSTGRES for the product name EnterpriseDB


sheelprabhakar:
In version 5.1.1, getting an exception for PostGres Enterprise Database version (EDB JDBC Connector), when using com.edb.Driver for connection
java.lang.IllegalArgumentException: DatabaseType not found for product name: [EnterpriseDB]

A possible solution can be by adding one condition in the method fromMetaData of class DatabaseType for EnterpriseDB
```java
    public static DatabaseType fromMetaData(DataSource dataSource) throws MetaDataAccessException {
        String databaseProductName = (String) JdbcUtils.extractDatabaseMetaData(dataSource, DatabaseMetaData::getDatabaseProductName);
        if (StringUtils.hasText(databaseProductName) && databaseProductName.startsWith("DB2")) {
            String databaseProductVersion = (String)JdbcUtils.extractDatabaseMetaData(dataSource, DatabaseMetaData::getDatabaseProductVersion);
            if (databaseProductVersion.startsWith("ARI")) {
                databaseProductName = "DB2VSE";
            } else if (databaseProductVersion.startsWith("DSN")) {
                databaseProductName = "DB2ZOS";
            } else if (!databaseProductName.contains("AS") || !databaseProductVersion.startsWith("QSQ") && !databaseProductVersion.substring(databaseProductVersion.indexOf(86)).matches("V\\dR\\d[mM]\\d")) {
                databaseProductName = JdbcUtils.commonDatabaseName(databaseProductName);
            } else {
                databaseProductName = "DB2AS400";
            }
        }else if(StringUtils.hasText(databaseProductName) && databaseProductName.equals("EnterpriseDB")) {
            databaseProductName = "PostgreSQL";
        }else {
            databaseProductName = JdbcUtils.commonDatabaseName(databaseProductName);
        }

        return fromProductName(databaseProductName);
    }
```

PR #4628, opened 2024-06-29
Add DatabaseType POSTGRES for the product name EnterpriseDB

sheelprabhakar:
Issue: [#4627]
EnterpriseDB is an enterprise version of PostGres database, which identifies itself as EnterpriseDB.
Added support for EnterpriseDB by adding the following condition in DatabaseType enum.

```java
else if (StringUtils.hasText(databaseProductName) && databaseProductName.startsWith("EnterpriseDB")) {
			databaseProductName = "PostgreSQL";
		}
```
Since Enterprise version of PostGre requires license not able to add integration tests


fmbenhassine:
That's a valid issue. Thank you for raising it and for opening a PR!

LGTM 👍 Rebased and merged as 5fb7b5a3f183f34b685dfd3277d1037518c60d32. Thank you for your contribution!

PR #4658, opened 2024-09-06
Move ORDER BY in getLastStepExecution from DB to java

### LINKED ISSUE #4657: Bad performance of GET_LAST_STEP_EXECUTION query on DB2 with large STEP_EXECUTION table
**Bug description**
We are migrating from Spring Batch 3.x to 5.x and are experiencing performance issues (long delays between each step).
This is on DB2, with a quite large STEP_EXECUTION table (50 million records).
After some digging, I found this to be caused by the [GET_LAST_STEP_EXECUTION](https://github.com/spring-projects/spring-batch/blob/45b3c5ef93642f5fbd46302a38e24cd303927bb8/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcStepExecutionDao.java#L95-L101) query, introduced in https://github.com/spring-projects/spring-batch/commit/62a8f441ab0ab0447cf9e5eabcfd72a106558ef2.

**Environment**
Spring Batch 5.1.2, Java 21, DB2 v10.5.

Our DBA has not been able to find a solution to improve the performance with an additional index.
What we did find out is that, by removing the ORDER BY, we get the result (typically just a single row in the normal case) in a matter of milliseconds. But with the ORDER BY, it takes ~ 60 seconds.

So a possible workaround would be to remove the ORDER BY and perform the sorting on the java side.
That is something I am currently trying out with the approach of overriding `JdbcStepExecutionDao#getLastStepExecution` with a custom implementation as described [here](https://github.com/spring-projects/spring-batch/pull/3635#issuecomment-611571008). Note that overriding `JdbcStepExecutionDao#getLastStepExecution` is not trivial though. Neither `JdbcStepExecutionDao` nor `JobRepositoryFactoryBean` are very inheritance-friendly.

jpraet:
Fixes #4657.

This addresses performance issues with large STEP_EXECUTION table on DB2.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcStepExecutionDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JdbcStepExecutionDao.java
@@ -24,6 +24,7 @@ import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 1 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 5a62de9031923aac0e76fc52b972ee2aafdd00c3. Thank you for your contribution!

PR #4659, opened 2024-09-09
Remove Duplicate If Condition in SimpleJobTests

chldppwls12:
There is already a section above that handles `JobInterruptedException`.
https://github.com/spring-projects/spring-batch/blob/425134c74e6b69dde40a81278e4fddbec4778f80/spring-batch-core/src/test/java/org/springframework/batch/core/job/SimpleJobTests.java#L573-L578
Therefore, this part is unreachable.
https://github.com/spring-projects/spring-batch/blob/425134c74e6b69dde40a81278e4fddbec4778f80/spring-batch-core/src/test/java/org/springframework/batch/core/job/SimpleJobTests.java#L591-L596


fmbenhassine:
LGTM 👍 Rebased and merged. Thank you for your contribution!

Issue #2020, opened 2010-05-12
ExecutionContext.isDirty() is not very accurate [BATCH-1566]


spring-projects-issues:
**[Dave Syer](https://jira.spring.io/secure/ViewProfile.jspa?name=david_syer)** opened **[BATCH-1566](https://jira.spring.io/browse/BATCH-1566?redirect=false)** and commented

A context is only dirty if a **new** value is put or an old one is removed, but the implementation sets dirty on all puts.

---

**Referenced from:** commits https://github.com/spring-projects/spring-batch/commit/963142cfa837e5d766013510ea4063bce8167dd2

PR #4691, opened 2024-10-27
Fix `ExecutionContext` dirty flag set to false on same value put or null put

### LINKED ISSUE #4685: ExecutionContext dirty flag is reset by a new put
**Bug description**
ExecutionContext.isDirty is supposed to tell if the context has been changed with a "put" operation since the dirty flag was last cleared (or the context was created). Only the clearDirtyFlag method is supposed to clear the flag once it has been set.

Instead, a "put" that sets a value to the same as its original value will also set the dirty flag to false, even if another put has happened before that did make a meaningfull change

**Environment**
Any code since  commit https://github.com/spring-projects/spring-batch/commit/963142cfa837e5d766013510ea4063bce8167dd2 will have this issue. The environment is not relevant

**Steps to reproduce**
- Do a meaningful put (change a value to something different)
- Do another put that does not actually change a value (for example, repeating the previous put)

**Expected behavior**
The dirty flag will still be true

    @Test
    void dirtyContextIsDirty() {
        ExecutionContext context = new ExecutionContext();
        context.put("hello", "world");
        assert context.isDirty(); // succeeds
        context.put("hello", "world");
        assert context.isDirty(); // fails
    }

**Further context**
I did not have a problem with this yet. I just stumbled across it when looking at ExecutionContext
The issue was mentioned, but not addressed in https://github.com/spring-projects/spring-batch/issues/2020#issuecomment-566289477

**Proposed fix**

    this.dirty = this.dirty || result == null || !result.equals(value);

GGHDMS:
#### Description
Resolved an issue where the `dirty` flag was incorrectly set to `false` when a `put` operation used the same value as the existing one or when null was pushed. Now, the flag remains true if a prior change has occurred.

#### Solution
Adjusted the `put` method logic to correctly handle value comparisons:
```java
if (value != null) {
	Object result = this.map.put(key, value);
	boolean newDirty = result == null || !result.equals(value);

	if (!this.dirty) {
		this.dirty = newDirty;
	}
}
else {
	Object result = this.map.remove(key);

	if (!this.dirty) {
		this.dirty = result != null;
	}
}
```

#### Fix #4685 #4692

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
@@ -133,7 +133,10 @@ public class ExecutionContext implements Serializable {
 		}
 		else {
 			Object result = this.map.remove(key);
-			this.dirty = result != null;
+
+			if (!this.dirty) {
+				this.dirty = result != null;
+			}
 		}
 	}
 
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,6 +37,7 @@ import org.springframework.lang.Nullable;
  * @author Lucas Ward
  * @author Douglas Kaminsky
  * @author Mahmoud Ben Hassine
+ * @author Seokmun Heo
  */
 public class ExecutionContext implements Serializable {
 
@@ -124,7 +125,11 @@ public class ExecutionContext implements Serializable {
 	public void put(String key, @Nullable Object value) {
 		if (value != null) {
 			Object result = this.map.put(key, value);
-			this.dirty = result == null || !result.equals(value);
+			boolean newDirty = result == null || !result.equals(value);
+
+			if (!this.dirty) {
+				this.dirty = newDirty;
+			}
 		}
 		else {
 			Object result = this.map.remove(key);
--- a/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/ExecutionContextTests.java
+++ b/spring-batch-infrastructure/src/test/java/org/springframework/batch/item/ExecutionContextTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,7 @@ import org.springframework.util.SerializationUtils;
 /**
  * @author Lucas Ward
  * @author Mahmoud Ben Hassine
- *
+ * @author Seokmun Heo
  */
 class ExecutionContextTests {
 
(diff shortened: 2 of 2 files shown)


mjwiq:
This should also be in the `else` branch. If the inserted value is null we don't want to remove the dirty flag either.

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/ExecutionContextTests.java
```diff
@@ -88,13 +88,13 @@ void testNotDirtyWithDuplicate() {
...
 	@Test
-	void testNotDirtyWithRemoveMissing() {
+	void testDirtyWithRemoveMissing() {
 		context.putString("1", "test");
 		assertTrue(context.isDirty());
 		context.putString("1", null); // remove an item that was present
 		assertTrue(context.isDirty());
 		context.putString("1", null); // remove a non-existent item
-		assertFalse(context.isDirty());
+		assertTrue(context.isDirty());
```
mjwiq, line 97: By changing this line you change the meaning of the test. I think it would be better to clear the dirty flag before setting it to null for the second time

mjwiq: actually testing that it stays dirty when inserting null twice could then be done in its own test or testDirtyWithDuplicate

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
```diff
@@ -124,11 +125,18 @@ public void putDouble(String key, double value) {
 	public void put(String key, @Nullable Object value) {
 		if (value != null) {
 			Object result = this.map.put(key, value);
-			this.dirty = result == null || !result.equals(value);
+			boolean newDirty = result == null || !result.equals(value);
```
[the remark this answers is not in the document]
GGHDMS: I didn't think one line of code was intuitive, so I made it through newDirty

spring-batch-infrastructure/src/test/java/org/springframework/batch/item/ExecutionContextTests.java
GGHDMS: I made some adjustments to the test based on the feedback. I added a step to clear the dirty flag before trying to remove an already non-existent item. This way, the test clearly checks that the dirty state doesn’t change in that situation. I think it makes the intent of the test more explicit and separates the different scenarios, making it easier to understand.

spring-batch-infrastructure/src/main/java/org/springframework/batch/item/ExecutionContext.java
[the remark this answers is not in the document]
mjwiq: I guess that's a personal thing then. I find the or statements easier to read. Let's see if we get a 3rd opinion from the maintainers

Solodye:
Just curious, are there any UT that proved that the `dirty` flag affects the partitioner behaviour? 

I am curious about this because the previous issue  https://github.com/spring-projects/spring-batch/issues/2020#issuecomment-566289477 mentioned this.

fmbenhassine:
Thank you for the PR, @GGHDMS ! And thanks to @mjwiq for the follow up and review, appreciated.

The changes LGTM 👍 Rebased and merged as 42b1464a0994460112c854c4f93903f59c2a9bc9. I just amended a small Javadoc update and inlined the `newDirty` variable (I have no preference regarding the coding style, but I just wanted to keep the condition in the same style as before).

> Just curious, are there any UT that proved that the dirty flag affects the partitioner behaviour?
I am curious about this because the previous issue https://github.com/spring-projects/spring-batch/issues/2020#issuecomment-566289477 mentioned this.

@Solodye  Not to my knowledge, but if this is still a problem, please open a separate issue and I will debug the case. Thank you upfront.

PR #4696, opened 2024-10-29
Fix link section

ngocnhan-tran1996:
Reference link
https://docs.spring.io/spring-batch/reference/spring-batch-integration.html


fmbenhassine:
LGTM. Thank you for the fix!

PR #595, opened 2018-04-12
BATCH-2515: Integrate SAP Hana as supported Spring Batch database

breglerj:
This commit adds SAP HANA as a supported Spring Batch database, enabling
developers to seamlessly move their existing Spring Batch projects to SAP HANA
or easily starting new Spring Batch projects on SAP HANA.

This commit contains the following changes:

- Add SAP HANA to the DatabaseType enum
- Add HanaPagingQueryProvider and tests
- Add properties files for SAP HANA
- Increase Spring version to 5.0.5 which is the first version to include
  HanaSequenceMaxValueIncrementer (see [https://jira.spring.io/browse/SPR-16558](https://jira.spring.io/browse/SPR-16558))

PR #4711, opened 2024-11-19
Add OceanBase database type

xzxiaoshan:
The use of OceanBase database is increasing, and we hope to increase support for OceanBase database types.


fmbenhassine:
Thank you for opening this PR. However, there is more than updating the `DatabaseType` enum to add support for a database in Spring Batch. Please take a look at #595 as an example.

Moreover, there is no plan for the moment to add support for OceanBase. Are you willing to contribute that to the https://github.com/spring-projects/spring-batch-extensions repository?

fmbenhassine:
As mentioned previously, we have no plan to add support for OceanBase in the official repository of Spring Batch for the moment. However, you can request to contribute your effort to the https://github.com/spring-projects/spring-batch-extensions repository and we will guide you through the process to lead an extension there if you are interested. Thank you.

Issue #1430, opened 2014-01-17
SqlWindowingPagingQueryProvider does not generate correct SQL for Oracle [BATCH-2166]


spring-projects-issues:
**[Philippe BAYE](https://jira.spring.io/secure/ViewProfile.jspa?name=pbaye)** opened **[BATCH-2166](https://jira.spring.io/browse/BATCH-2166?redirect=false)** and commented

The generated SQL requests look like this (for both first page & remaining pages) :
SELECT * from (....) AS TMP_SUB WHERE ...

But for Oracle, the correct SQL requests look like this :
SELECT * from (....) TMP_SUB WHERE ...

The problem is located in the getSubQueryAlias() method.

In previous version (2.0.4) the SQL requests are corrected.

---

**Affects:** 2.1.9

cppwfs:
Thank you for opening the issue.   Can you retry with the latest release of Spring Batch(5.0.2) and report back the results?   
If the issue is reproducible, can you provide a sample project that uses the latest release of Spring Batch and that exhibits the behavior?  To help you in reporting your issue, we have prepared a project template that you can use as a starting point. Please check the [Issue Reporting Guidelines](https://github.com/spring-projects/spring-batch/blob/main/ISSUE_REPORTING.md) for more details about this.

Issue #1253, opened 2015-02-11
DerbyPagingQueryProvider remaining pages query is not correct


spring-projects-issues:
**[Jimmy Praet](https://jira.spring.io/secure/ViewProfile.jspa?name=jpraet)** opened **[BATCH-2348](https://jira.spring.io/browse/BATCH-2348?redirect=false)** and commented

```
DerbyPagingQueryProvider queryProvider = new DerbyPagingQueryProvider();
queryProvider.setFromClause("TEST.TABLE");
queryProvider.setSelectClause("X, Y, Z");
queryProvider.setWhereClause("X = ?");
Map<String, Order> sortKeys = new HashMap<String, Order>();
sortKeys.put("Y", Order.ASCENDING);
sortKeys.put("Z", Order.ASCENDING);
queryProvider.setSortKeys(sortKeys);
System.out.println(queryProvider.generateRemainingPagesQuery(10));
```

results in the following query:

```
SELECT * FROM 
 ( SELECT TMP_ORDERED.*, ROW_NUMBER() OVER () AS ROW_NUMBER FROM 
   (SELECT X, Y, Z FROM TEST.TABLE WHERE X = ? ) 
    AS TMP_ORDERED
 ) 
 AS TMP_SUB WHERE TMP_SUB.ROW_NUMBER <= 10 
 AND ((Y > ?) OR (Y = ? AND Z > ?)) ORDER BY Y ASC, Z ASC
```

I believe the sort key selection and order by should be applied to the inner subquery instead of the outer query:

```
SELECT * FROM 
 ( SELECT TMP_ORDERED.*, ROW_NUMBER() OVER () AS ROW_NUMBER FROM 
   (SELECT X, Y, Z FROM TEST.TABLE WHERE X = ? 
    AND ((Y > ?) OR (Y = ? AND Z > ?)) ORDER BY Y ASC, Z ASC) 
	AS TMP_ORDERED
 ) 
 AS TMP_SUB WHERE TMP_SUB.ROW_NUMBER <= 10 
```

---

**Affects:** 2.2.7, 3.0.3

fmbenhassine:
Resolved with #4713

PR #4713, opened 2024-11-23
Update implementations of `PagingQueryProvider`

### LINKED ISSUE #1673: Update DerbyPagingQueryProvider for Derby 10.7 and above
**[Mikalai Dushkou](https://jira.spring.io/secure/ViewProfile.jspa?name=elgad)** opened **[BATCH-1921](https://jira.spring.io/browse/BATCH-1921?redirect=false)** and commented

Since Derby DB from version 10.7 support "OFFSET" and "FETCH FIRST"/"FETCH NEXT" clauses it nice to have this option. Please see implemented solution in attached file.

---

**Attachments:**
- [DerbySqlPagingQueryProvider.java](https://jira.spring.io/secure/attachment/20801/DerbySqlPagingQueryProvider.java) (_2.42 kB_)

hpoettker:
Resolves #1673 and #1253.

As reported in #1253, the current implementation of `DerbyPagingQueryProvider` doesn't really work, which can be confirmed by running the added `DerbyPagingQueryProviderIntegrationTests` against the current implementation.

This PR replaces the implementation of `DerbyPagingQueryProvider` effectively with that of `Db2PagingQueryProvider` as suggested by #1253 and is tested against Derby in the added `DerbyPagingQueryProviderIntegrationTests`. The build uses Derby version 10.16.1.1, which is the first and only release with a baseline of Java 17 as the next Derby release already requires Java 21.

I took the liberty of deprecating `SqlWindowingPagingQueryProvider` for removal. The only `PagingQueryProvider` that currently uses its implementation is `DerbyPagingQueryProvider`. The other 3 subclasses (and with this PR also `DerbyPagingQueryProvider`) don't actually depend on its functionality as they override any relevant methods.


hpoettker:
I've pushed a second commit that includes not only an integration test for Derby, but also for Db2, HSQL, MariaDB, MySQL, Oracle, PostgreSQL, Sql Server, and SQLite.

The setup is mostly from the job repository integration tests. For Oracle, I've switched the tests to the publicly available image from here: https://github.com/gvenzl/oci-oracle-xe/

This should allow to also close #1430, as the PR proposes to remove `SqlWindowingPagingQueryProvider` anyway, and the tests show that `OraclePagingQueryProvider` should work as expected.

The two tests in `AbstractPagingQueryProviderIntegrationTests` are currently quite minimal. In the future, it would be interesting to also investigate queries with joins on columns with the same name. There are some open bug reports about this situation.

hpoettker:
done

fmbenhassine:
Perfect 👍  Rebased and merged. Thank you for this awesome contribution!!

PR #4714, opened 2024-11-24
Fix parsing of the `stop` element in step XML

### LINKED ISSUE #1287: Job restart after stop fails [BATCH-2315]
**[Chris Haas](https://jira.spring.io/secure/ViewProfile.jspa?name=chrishaas)** opened **[BATCH-2315](https://jira.spring.io/browse/BATCH-2315?redirect=false)** and commented

In our jobs we use the following construct to conditionally stop a job based on a property passed into the job:

```xml
<decision id="decisionId" decider="decider">
    <stop on="NEEDS_FINALIZATION" restart="placeAccounts"/>
    <next on="*" to="placeAccounts"/>
</decision>
```

In Spring Batch 2.1, "placeAccounts" is executed when the job is restarted.  In Spring Batch 3, the decision is re-executed, which of course means the job immediately stops again.

In 3.0.1, the job's flow has stateTransitions for the decision look like

```
StateTransition: [state=placementImport.placementFinalizationDecision, pattern=*, next=placementImport.placeAccounts],
StateTransition: [state=placementImport.placementFinalizationDecision, pattern=NEEDS_FINALIZATION, next=placementImport.stop60],
StateTransition: [state=placementImport.stop60, pattern=*, next=null],
```

but in 2.1, they look like:

```
StateTransition: [state=placementImport.placementFinalizationDecision, pattern=*, next=placementImport.placeAccounts],
StateTransition: [state=placementImport.placementFinalizationDecision, pattern=NEEDS_FINALIZATION, next=placementImport.stop60],
StateTransition: [state=placementImport.stop60, pattern=*, next=placementImport.placeAccounts],
```

Notice that the next element on stop60 is blank, which is probably the cause of the problem.

---

**Affects:** 3.0.1

2 votes, 3 watchers

hpoettker:
Resolves #1287.

Previously, users were forced to set the attribute `exit-code` of the `stop` element to `""` as a work-around in some cases as the default value `STOPPING` or any non-null value of `exit-code` would lead to failing job restarts.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/xml/AbstractFlowParser.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/xml/AbstractFlowParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


fmbenhassine:
Thank you for fixing this! I remember I worked on a reproducer (as you might have seen in the original issue), but I've never got the chance to fix the issue (or more precisely, to prioritize the issue).

Rebased and merged as 39c593e150d0e86094d14561fd57471358d58499. Thanks again for all your contributions! Really appreciated 👍

Issue #4703, opened 2024-11-10
Add StepExecution parameter to StoppableTasklet.stop()


andrianov17:
Currently, StoppableTasklet.stop() doesn't take any parameters. So, from within tasklet it is impossible to distinguish which exactly step execution is requested to stop out of multiple job executions running. This is important when some external request needs to be executed from StoppableTasklet.stop().

Looking into SimpleJobOperator.stop(long executionId) implementation, it looks very straightforward to implement this request - just pass stepExecution to the tasklet stop() method:

```
@Override
	public boolean stop(long executionId) throws NoSuchJobExecutionException, JobExecutionNotRunningException {

		JobExecution jobExecution = findExecutionById(executionId);
		// Indicate the execution should be stopped by setting it's status to
		// 'STOPPING'. It is assumed that
		// the step implementation will check this status at chunk boundaries.
		BatchStatus status = jobExecution.getStatus();
		if (!(status == BatchStatus.STARTED || status == BatchStatus.STARTING)) {
			throw new JobExecutionNotRunningException(
					"JobExecution must be running so that it can be stopped: " + jobExecution);
		}
		jobExecution.setStatus(BatchStatus.STOPPING);
		jobRepository.update(jobExecution);

		try {
			Job job = jobRegistry.getJob(jobExecution.getJobInstance().getJobName());
			if (job instanceof StepLocator) {// can only process as StepLocator is the
												// only way to get the step object
				// get the current stepExecution
				for (StepExecution stepExecution : jobExecution.getStepExecutions()) {
					if (stepExecution.getStatus().isRunning()) {
						try {
							// have the step execution that's running -> need to 'stop' it
							Step step = ((StepLocator) job).getStep(stepExecution.getStepName());
							if (step instanceof TaskletStep) {
								Tasklet tasklet = ((TaskletStep) step).getTasklet();
								if (tasklet instanceof StoppableTasklet) {
									StepSynchronizationManager.register(stepExecution);
									((StoppableTasklet) tasklet).stop(**stepExecution**);
									StepSynchronizationManager.release();
								}
							}
						}
```

Thank you in advance!

HyunSangHan:
Can I work on this issue?

HyunSangHan:
I have submitted a PR!: https://github.com/spring-projects/spring-batch/pull/4715

fmbenhassine:
Thank you for the feedback. The step execution is already available to any tasklet through the chunk context (or even the step contribution):

```java
[...]

PR #4715, opened 2024-11-24
GH-4703: Enhance StoppableTasklet interface to support step-specific stop requests

HyunSangHan:
Closes: https://github.com/spring-projects/spring-batch/issues/4703

Resolves a limitation in the `StoppableTasklet` interface where distinguishing which specific StepExecution to stop was not possible during concurrent job executions.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/StoppableTasklet.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/StoppableTasklet.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.batch.core.step.tasklet;
 
+import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.core.launch.JobOperator;
 
 /**
@@ -28,6 +29,7 @@ import org.springframework.batch.core.launch.JobOperator;
  * so the appropriate thread safety and visibility controls should be put in place.
  *
  * @author Will Schipp
+ * @author Hyunsang Han
  * @since 3.0
  */
 public interface StoppableTasklet extends Tasklet {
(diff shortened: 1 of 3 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/StoppableTasklet.java
```diff
@@ -38,4 +39,14 @@ public interface StoppableTasklet extends Tasklet {
...
+	/**
+	 * Used to signal that the job should stop, providing access to the current
+	 * {@link StepExecution} context.
+	 *
+	 * @param stepExecution the current {@link StepExecution} context in which the job
+	 * is being executed
+	 */
+	default void stop(StepExecution stepExecution) {
+		stop();
+	}
```
HyunSangHan, line 51: The new method provides a **default implementation** that delegates the call to the existing stop() method, ensuring backward compatibility for implementations that do not need step-specific stopping logic.

spring-batch-core/src/main/java/org/springframework/batch/core/step/tasklet/StoppableTasklet.java
```diff
@@ -38,4 +40,14 @@ public interface StoppableTasklet extends Tasklet {
...
```
HyunSangHan, line 52: The new method provides a **default implementation** that delegates the call to the existing stop() method, ensuring backward compatibility for implementations that do not need step-specific stopping logic.

If you have any suggestions for improvement, please feel free to provide feedback!

fmbenhassine:
Rebased, squashed and merged as bdb2bfc8fe2124b5c53121ef4e2d4f7597a65c74. Refined in 0056722fffa3bbf9cf9cd114b2ebb42293fb7db3. Thank you for your contribution!

Issue #4670, opened 2024-09-25
Pass the StepExecution to the Partitioner API


djechelon:
In order to take an if-else decision based on job arguments (specifically, my job can be executed against a "reference date" so I can pass it from command like), I would like to use that information in the Partitioner to select files/item/rows according to that information.

Currently, only information available is the grid size. According to SimpleStepExecutionSplitter, the StepExecution is available before calling the Partitioner.

I would kindly ask if it is possible, for the next version, to add the parameter to the Partitioner interface

Thanks for your attention

injae-kim:
Fix PR: https://github.com/spring-projects/spring-batch/pull/4716

fmbenhassine:
Thank you for opening this feature request. We need a concrete example that shows the limitation of the current API to introduce such a big change. Can you please provide a code example for your use case?

From what I understand, the implementation of your partitioner depends on a job parameter. In this case, you can declare your partitioner as a step-scoped bean and use late binding to inject job parameters. If you elaborate with more details, I can try to give an example.

fmbenhassine:
@djechelon as mentioned in my previous comment, I believe a step-scoped partitioner is what you are looking for to use late-binding of job parameters. Here is a quick example:

```java
// Step 1: define the partitioner
public static class MyPartitioner implements Partitioner {

    private Date referenceDate;

    public MyPartitioner(Date referenceDate) {
        this.referenceDate = referenceDate;
    }

    @Override
    public Map<String, ExecutionContext> partition(int gridSize) {
        Map<String, ExecutionContext> partitions = new HashMap<>(gridSize);
        // use reference date as needed to determine partitions
        return partitions;
    }
}

// Step 2: declare it as a step-scoped bean and inject the job parameter in it
@Bean
@StepScope
public MyPartitioner partitioner(@Value("#{jobParameters['referenceDate']}") Date referenceDate) {
    return new MyPartitioner(referenceDate);
}
```

[Late-binding of job/step context attributes](https://docs.spring.io/spring-batch/reference/step/late-binding.html) is a really powerful feature in Spring Batch!

Another option is to make the partitioner a job execution listener and inject the job execution in it:

```java
public static class MyPartitioner implements Partitioner {

    private JobExecution jobExecution;

    @BeforeJob
    public void beforeJob(JobExecution jobExecution) {
        this.jobExecution = jobExecution;
    }

    @Override
    public Map<String, ExecutionContext> partition(int gridSize) {
[...]

PR #4716, opened 2024-11-24
Add function that accepts StepExecution parameter to Partitioner API

wesley-94:
Add function that accepts StepExecution parameter to Partitioner API

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/Partitioner.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/Partitioner.java
@@ -18,6 +18,7 @@ package org.springframework.batch.core.partition.support;
 
 import java.util.Map;
 
+import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.item.ExecutionContext;
 
 /**
@@ -30,7 +31,7 @@ import org.springframework.batch.item.ExecutionContext;
  * @author Taeik Lim
  * @since 2.0
  */
-@FunctionalInterface
+// @FunctionalInterface
 public interface Partitioner {
 
 	/**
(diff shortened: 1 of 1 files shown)


injae-kim:
Closes: https://github.com/spring-projects/spring-batch/issues/4670

spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/Partitioner.java
```diff
@@ -30,7 +31,7 @@
  * @author Taeik Lim
  * @since 2.0
  */
-@FunctionalInterface
+// @FunctionalInterface
```
injae-kim, line 34: can you remove `//`? (maybe we should remove `@FunctionalInterface` cause we have 2 method

wesley-94: I modified it as you reviewed. Thank you!

djechelon:
Thanks so much for this PR.
But this looks incomplete.

1. You have to amend the caller, which I don't see in the PR. The caller must point at the new method passing the StepExecution
2. I don't like the idea of having non-default `partition(int)` because what we want is both to not create regressions, but to encourage users transition to the new API. There is no other possibility than making both methods default IMO

wesley-94:
Thanks for your answer.
But ‘No target method found’ message was found at @FunctionalInterface when using ‘default’ keyword in both of methods.

quaff:
You should update `SimpleStepExecutionSplitter` to call the new method.

https://github.com/spring-projects/spring-batch/blob/961734b260ff20c5b936fb17d37338bf90436efb/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/SimpleStepExecutionSplitter.java#L191

https://github.com/spring-projects/spring-batch/blob/961734b260ff20c5b936fb17d37338bf90436efb/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/SimpleStepExecutionSplitter.java#L207

fmbenhassine:
Closing this as explained in https://github.com/spring-projects/spring-batch/issues/4670#issuecomment-3354015674.

Thank you for your PR anyway 🙇‍♂️

PR #4720, opened 2024-11-25
fixed class name and grammatical error in docs whatsnew

uli-f:
two small fixes on whatsnew 5.2 page


fmbenhassine:
Good catch! Thank you for these fixes!

PR #4731, opened 2024-12-06
Add state reset on close of `MongoPagingItemReader`

### LINKED ISSUE #1086: MongoPagingItemReader never reset page number
**[Oleksandr Loushkin](https://jira.spring.io/secure/ViewProfile.jspa?name=alisternoorm)** opened **[BATCH-2517](https://jira.spring.io/browse/BATCH-2517?redirect=false)** and commented

MongoItemReader should override **doClose()** method with reset page number.
Otherwise page will never be returned to 0.

As we can see **AbstractPaginatedDataItemReader** increases page by **page ++**

For example **RepositoryItemReader** has method.

I can create pull request with quick fix.

---

**Affects:** 3.0.7

hpoettker:
Resolves #1086

The `MongoPagingItemReader` does not have any stateful fields beyond those of its superclasses.

The `AbstractPaginatedDataItemReader` has two more deprecated subclasses: `MongoItemReader` and `Neo4jItemReader`. The fix works the same for all three classes.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/AbstractPaginatedDataItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/AbstractPaginatedDataItemReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2023 the original author or authors.
+ * Copyright 2013-2024 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 I like the idea of putting the change in the base class.

Rebased and merged as 2bd3c15587c1dab5a31ffc1f235eede248b61244. Thank you for your contribution!

PR #4737, opened 2024-12-22
Allow CI on any branch, skip CD on forks

scordio:
Prior to this commit, running the workflow on a fork's `main` branch would fail due to the deployment steps, and no CI was executed on branches and pull requests.

With this change, the deployment steps are executed only when the current repository is `spring-projects/spring-batch` and the branch is `main`. Otherwise, only `mvn verify` is executed. Plus, CI execution is now available for branches and pull requests.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,9 +1,6 @@
 name: CI/CD build
 
-on:
-  workflow_dispatch:
-  push:
-    branches: [ "main" ]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


fmbenhassine:
LGTM 👍 Thank you for this PR!

scordio:
I just noticed that, after yesterday's rebase, the execution took 6 hours and eventually failed. Maybe further optimizations are possible like excluding long running tests from PR executions. I'll try to dig a bit more.

hpoettker:
The problem with the stalling build has been there for a while: #4741

The PR might solve it or might not. So far, I couldn't reproduce it locally.

fmbenhassine:
@hpoettker I was working on #4741 just after this PR, which is now merged.

Thank you both for your contributions!

PR #4741, opened 2025-01-07
Polish integration tests

hpoettker:
Due to `FaultTolerantStepFactoryBeanRollbackIntegrationTests`, the build pipeline has been intermittently stalling recently. The similar test `FaultTolerantStepFactoryBeanIntegrationTests` was disabled in https://github.com/spring-projects/spring-batch/commit/3758f86b for the same behavior.

So far, I couldn't reproduce the problem locally. But the PR offers some minor improvements.

In the Mongo integration tests, the application context and in particular the Mongo client are now closed after each test, respectively. This prevents the Mongo clients from closing on a failed health check during later tests, which is harmless but produces noisy logging during those tests.

I don't have a good hypothesis why the fault-tolerance tests are stalling intermittently. So I don't know whether this PR will resolve the issue. But it implements some concurrency best practices:

- the readers have been replaced by the battle tested `ListItemReader` wrapped in a `SynchronizedItemReader`.
- unused fields in the processors and writers have been removed
- fields in the processors and writers are now either constant or thread-safe collections
- fields in the processors and writers are now `final` and set in the constructor

Furthermore, the tests now have a time-out of 30 seconds such that they are cleanly aborted if they should still stall.

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/repository/support/MongoDBIntegrationTestConfiguration.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/repository/support/MongoDBIntegrationTestConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 the original author or authors.
+ * Copyright 2024-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


fmbenhassine:
That's great! Thank you for all these improvements 👍 

Rebased and merged as bbd46fb29d69026a60314ad44c8834462f2dad9c.

hpoettker:
The test still fails. 😞 
But at least it doesn't take 6 hours to do so. 😄 

It's probably easiest to disable the two tests until we have an idea what tht root cause is.

fmbenhassine:
Yes I just saw the build failure and tried to re-run it.

I remember I was never able to reproduce it locally and was failing on Github Actions, that's why I disabled it at the time. 

OK no problem I will disable them again.

PR #4750, opened 2025-01-23
Attempt to close all delegate writers even when some fail

elimelec:
Motivation:

When a delegate writer fails during close(), the original implementation would stop immediately, preventing other delegates from completing their cleanup.

For example: no footers written to all files because of an Exception from the first delegate, and the main content is already written at this point.


fmbenhassine:
LGTM 👍  Thank you for reporting and fixing the issue!

I planned this for the next patch release.

fmbenhassine:
Rebased and merged bd8f9a8d7603fba5fcf5fc03252c13d095cec175. Thank you for your contribution!

I think the same issue can happen for delegates in the composite reader, so if you want to contribute the same fix then you are welcome. Otherwise please let me know and I will take care of it.

PR #4756, opened 2025-02-03
Step names must be unique #3757

FBibonne:
the names of different steps in a job must be different

Sloves #3757

A few comments :
  - pom is updated to have a have a version of slf4j-api corresponding to simple-slf4j
  - unicity of names is checked when starting the job , just before calling listeners because step names may be known only in JobScope
  - in FlowJob.findStepsThrowingIfNameNotUnique, I suggest to remove the else block when state is instance of StepHolder : it is never executed


FBibonne:
this PR  makes the same changes as #4544 which was closed dur to trouble with git

PR #4764, opened 2025-02-20
Attempt to close all delegate readers even when some fail

elimelec:
Motivation:

When a delegate reader fails during close(), the original implementation would stop immediately, preventing other delegates from completing their cleanup.

Similar to https://github.com/spring-projects/spring-batch/pull/4750

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/step/item/TaskletStepExceptionTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/step/item/TaskletStepExceptionTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2008-2022 the original author or authors.
+ * Copyright 2008-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


elimelec:
@fmbenhassine This is split into 2 commits: 
- One for `CompositeItemReader`
- One for `CompositeItemStream`

I think this PR (and also the previous one: https://github.com/spring-projects/spring-batch/pull/4750) may change the public API. I realized only now because some tests were failing because of the changes from `CompositeItemStream`. Also, the documentation was explicit that the close method is not fully cleaning all delegates.

I still left both changes (since the interfaces is defined with `throws ItemStreamException` and I am not changing that part.

Let me know if I should change something.

fmbenhassine:
Thank you for this second PR! Appreciated 👍 

I see no public API changes, but I agree there is a small behavioral change in the sense that the exception is now suppressed instead of being the main one. Since we are fixing the partial cleanup bug, I believe this is acceptable for a patch release as long as we document the change about the way the exception is thrown (which is missing by now, but I will take care of that).

The PR LGTM, I will plan it for 5.2.2.

fmbenhassine:
Rebased and merged, thank you for your contributions!

As mentioned previously, I updated the javadocs about the change in suppressed exceptions: 75edb29d2.

PR #4766, opened 2025-02-24
Fix typo in intercepting-execution.adoc

charlie881007:
fix doc typo


fmbenhassine:
Good catch! Thank you for the fix 👍

PR #4772, opened 2025-02-26
Fix wrong statement in Javadoc of SimplePartitioner

quaff:


fmbenhassine:
Good catch 👍  Thank you your PR!

PR #4775, opened 2025-02-27
Use Files.delete() for better error reporting

elimelec:
Motivation:

`File.delete` provides only a simple boolean result. In case of failure, there's no information available. `Files.delete` will throw an exception with more details.

Example: Writer configured with `shouldDeleteIfEmpty` set to `true` fails to delete the empty file. No information is provided  other than a simple message: `"Failed to delete empty file on close"`.


fmbenhassine:
This is a great enhancement 👍 Rebased, squashed and merged as 4671b62b1. Thank you for your contribution !

PR #4779, opened 2025-03-05
docs: replaced outdated AssertFile reference with AssertJ

### LINKED ISSUE #4754: Outdated AssertFile usage in section "Validating Output Files" of the reference documentation
Since the removal of `AssertFile` in 5.2 (#4181), the docs about "Validating Output Files"  are outdated ([source](https://github.com/spring-projects/spring-batch/blob/8e91adb9ab95d66a1dc127f0b52134603517b589/spring-batch-docs/modules/ROOT/pages/testing.adoc?plain=1#L277-L295)).

I suggest either removing it (since it's no longer part of `spring-batch-test` or rewrite it, for example to something like:

```adoc
[[validatingOutputFiles]]
== Validating Output Files

When a batch job writes to the database, it is easy to query the database to verify that
the output is as expected. However, if the batch job writes to a file, it is equally
important that the output be verified.  Assertion libraries like AssertJ can help you with that by 
asserting that two files have the same content, line by line. Therefore, it is possible to create
a file with the expected output and to compare it to the actual result, as the following example 
shows:

[source, java]
----
private static final File EXPECTED_FILE = new File("src/main/resources/data/input.txt");
private static final File OUTPUT_FILE = new File("target/test-outputs/output.txt");

Assertions.assertThat(EXPECTED_FILE).hasSameTextualContentAs(OUTPUT_FILE);
----
```

If this altered text is okay, I could turn it into a PR.

g00glen00b:
closes #4754

### DIFF
--- a/spring-batch-docs/modules/ROOT/pages/testing.adoc
+++ b/spring-batch-docs/modules/ROOT/pages/testing.adoc
@@ -279,19 +279,17 @@ int count = StepScopeTestUtils.doInStepScope(stepExecution,
 
 When a batch job writes to the database, it is easy to query the database to verify that
 the output is as expected. However, if the batch job writes to a file, it is equally
-important that the output be verified. Spring Batch provides a class called `AssertFile`
-to facilitate the verification of output files. The method called `assertFileEquals` takes
-two `File` objects (or two `Resource` objects) and asserts, line by line, that the two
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Closing this, as discussed in #4754. Thank you for your PR anyway.

PR #4781, opened 2025-03-14
Fix typo in Javadocs of FlatFileItemWriterBuilder#name(String)

quaff:


fmbenhassine:
Good catch again 👍 Thank you for the fix.

PR #4783, opened 2025-03-15
Refactor `ExitStatus#isRunning` for type-safety

sjiwon:
## Overview

This PR improves the type-safety and readability of the `ExitStatus#isRunning` method in the Spring Batch core module.

## Changes

- Refactored the existing string-based comparisons in `ExitStatus#isRunning` by utilizing type-safe comparisons with predefined ExitStatus constants
- Added unit tests covering the possible scenarios for the isRunning method, ensuring correct behavior for each relevant status:
  - `return true`: EXECUTING, UNKNOWN
  - `return false`: COMPLETED, FAILED, STOPPED, NOOP


sjiwon:
@fmbenhassine 

I have a few questions regarding the ExitStatus#isRunning method:
- ExitStatus is designed as an immutable object, and it provides the replaceExitCode method, which returns a new instance with a different exit code. In this context, is the original intention of ExitStatus#isRunning that any ExitStatus instance having the exit code "EXECUTING" (regardless of object identity) should return true?
- If the above assumption is correct, my PR might actually be introducing a breaking change and unintentionally altering the original intent. Could you confirm whether my understanding is correct?
- I’m new to Spring Batch and initially thought that comparing enum-like singleton instances would suffice for checking running status. However, after considering the immutability and possibility of creating new instances via replaceExitCode, it appears that my proposed change may not align with the intended design. I’d appreciate your clarification.

Thanks for your help!

fmbenhassine:
LGTM 👍 Rebased, squashed and merged as 30b5120c931649c178b1f995097041045a73b599.

PR #4784, opened 2025-03-17
Stop using deprecated StepExecutionListenerSupport in document

quaff:


fmbenhassine:
Good catch! Thank you for updating this.

PR #4803, opened 2025-03-28
Use typesafe variant of `ctx::getBean` instead

quaff:


fmbenhassine:

PR #4811, opened 2025-04-21
Fix NullPointerException in MongoSequenceIncrementer

blake-bauman:
When using a MongoDB repository for the first time, the collections used by MongoSequenceIncrementer don't yet exist, which results in the call to findOneAndUpdate().getLong() to reutrn a null, which then results in an NPE unboxing the Long into a long.

Adding upsert(true) creates the initial document with a count of 0 and immediately increments it to 1.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/MongoSequenceIncrementer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/MongoSequenceIncrementer.java
@@ -47,7 +47,7 @@ public class MongoSequenceIncrementer implements DataFieldMaxValueIncrementer {
 		return mongoTemplate.execute("BATCH_SEQUENCES",
 				collection -> collection
 					.findOneAndUpdate(new Document("_id", sequenceName), new Document("$inc", new Document("count", 1)),
-							new FindOneAndUpdateOptions().returnDocument(ReturnDocument.AFTER))
+							new FindOneAndUpdateOptions().returnDocument(ReturnDocument.AFTER).upsert(true))
 					.getLong("count"));
 	}
 


blake-bauman:
Is this a valid PR?  We aren't able to use MongoDB as a repository without manually creating these documents and it took some reverse engineering to figure out what to create.

fmbenhassine:
> Is this a valid PR? We aren't able to use MongoDB as a repository without manually creating these documents and it took some reverse engineering to figure out what to create.

Similar to the JDBC job repository configuration, it is expected to execute [this script](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/resources/org/springframework/batch/core/schema-mongodb.js) in MongoDB before running any job. There is a note about that in the what's new section, but I believe this is not clear enough and we need to update the reference docs with this prerequisite in the job repository configuration section as well. I created #4859 for that.

With that script, there is no need for the `upsert` flag added in this PR (this [test](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/test/java/org/springframework/batch/core/repository/support/MongoDBJobRepositoryIntegrationTests.java), which initializes the database with that script, passes without the `upset` flag), so I guess this PR is indeed not valid. Do you agree?

blake-bauman:
I can see that you'd want to ensure that the indexes get created, so perhaps upsert isn't the right option after all.  I appreciate the documentation change.  I didn't think of looking in "what's new" since I just went straight to the reference guide.

The big problem, I think is that if the collection is not available, there is an NPE being thrown with no explanation, which could be an issue for those that didn't realize they missed a step.  A couple of suggestions I would have is:

* Detecting on startup if a required collection isn't available and failing with an informative message.
* Or.....One of the benefits of MongoDB over JDBC is the ability to create these collections with the right attributes on the fly.  So, a nice enhancement would be to detect that the collection is not created and automatically create it with the expected attributes before starting the job.

fmbenhassine:
I fully agree. We tend to prefer the fail-fast approach, so I would argue that option 1 is better (explicit is better than implicit). This applies to a JDBC setup as well. In addition to the documentation's clarification issue, I created #4865 to improve the code with a configuration check.

I am closing this PR in favor of #4865. Thank you for reporting this issue and for your time on this PR!

PR #4812, opened 2025-04-21
Allow keys other than _id for primary key in MongoItemWriter

blake-bauman:
When moving data between different data store types, such as between MongoDB and a SQL database, not all systems have a "_id" as their pimary key.  Most often it has a different name or can be multiple keys/columns used as a composite key.  This change allows the app to specify an alternate key or a set of keys which, together, can be used as the primary key for a MongoDB insert/update/upsert.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/data/MongoItemWriter.java
@@ -16,6 +16,10 @@
 
 package org.springframework.batch.item.data;
 
+import static java.util.stream.Collectors.*;
+
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


blake-bauman:
@fmbenhassine Does this seem reasonable?

fmbenhassine:
Yes, and the change is backward compatible. Thank you for the PR!

I will plan this for v6.0.0 / v5.2.3.

blake-bauman:
Thanks for this!

fmbenhassine:
LGTM 👍  Rebased and merged as 08085d3c84978c153725a3b31e36eb0be5a2f286 and will be included in the upcoming v5.2.3.

blake-bauman:
Awesome, thanks!

Issue #4673, opened 2024-10-02
Migrate to JSpecify annotations for nullability constraints


fmbenhassine:
Spring Framework is planning to migrate to [jSpecify](https://github.com/jspecify/jspecify) for the next major release (https://github.com/spring-projects/spring-framework/issues/28797). For consistency with other projects, Spring Batch will also migrate to jSpecify for the next major release.

scordio:
Hi @fmbenhassine, is this one up for grabs?

fmbenhassine:
@scordio Thank you for your offer to help! Based on my experience with #920, this is a _substantial_ amount of work, that's why I assigned it to myself. But if you are really up to it, then you are welcome! IIRC, in #920, we worked one module at a time (one PR per module to divide/simplify the work as well as the review).

scordio:
Happy to pick it up and I'll let you know if I don't see it feasible in a reasonable time 🙂

lobaorn:
Hey @scordio before tackling this, check [this comment](https://github.com/pgjdbc/pgjdbc/issues/3634#issuecomment-2941361286) for the OpenRewrite recipe.

scordio:
Thanks for the heads up, @lobaorn! Checking how OpenRewrite can simplify the task is indeed on my list 🙂

scordio:
Many thanks for pushing it forward, @fmbenhassine!

PR #4823, opened 2025-05-04
Mark argument to ExitStatus#equals as nullable

jbotuck:
I think this commit will fix this question of mine from 5 years ago https://stackoverflow.com/questions/56653315/why-is-kotlin-double-equals-not-finding-a-equals-method


fmbenhassine:
LGTM 👍 And the usage of the JSpecify annotation is in line with #4673, so this PR is timely welcome. Thank you for contribution!

Issue #4565, opened 2024-03-11
Access Job Description


VascoFreire:
**Issue:**
In a project that I'm working on, we are listing the instantiated jobs.

One of the requests that we have is to have a description for each job.

**Suggestion:**
Since it is possible to define a description for the job (in xml), it would be useful to have access to this field in the java object.

fmbenhassine:
I agree, thank you for raising this discrepancy between the XML and Java config styles. I will plan this for v6.

fmbenhassine:
I just wanted to write down some findings here after a quick look to estimate the cost of the required change to support this feature. Adding the description to `BATCH_JOB_INSTANCE` would be duplicated (the name is already duplicated in all records..), so this will be problematic for a normalized DB schema as we already have. Therefore, this might require a new table for jobs.

Here is the change I think we might need to support this feature properly:

```patch
Index: spring-batch-core/src/main/resources/org/springframework/batch/core/schema-mysql.sql
 -- Autogenerated: do not edit this file
+CREATE TABLE BATCH_JOB (
+    JOB_ID BIGINT  NOT NULL PRIMARY KEY,
+    JOB_NAME VARCHAR(100) NOT NULL,
+    JOB_DESCRIPTION VARCHAR(2500) NOT NULL,
+    constraint JOB_NAME_UN unique (JOB_NAME)
+)
 
 CREATE TABLE BATCH_JOB_INSTANCE (
 	JOB_INSTANCE_ID BIGINT  NOT NULL PRIMARY KEY,
+    JOB_ID BIGINT NOT NULL,
 	VERSION BIGINT,
-	JOB_NAME VARCHAR(100) NOT NULL,
-	JOB_KEY VARCHAR(32) NOT NULL,
-	constraint JOB_INST_UN unique (JOB_NAME, JOB_KEY)
-) ENGINE=InnoDB;
+	JOB_INSTANCE_KEY VARCHAR(32) NOT NULL,
+    constraint JOB_FK foreign key (JOB_ID) references BATCH_JOB(JOB_ID),
+	constraint JOB_INSTANCE_KEY_UN unique (JOB_INSTANCE_KEY)
+)
 
```

Job names and descriptions would be saved only once in `BATCH_JOB`, and `BATCH_JOB_INSTANCE` would link to that new table via the new `JOB_ID` column.

EDIT: The `Job` and `Step` interfaces could be turned into domain entities.

benelog:
@fmbenhassine 
[...]

PR #4830, opened 2025-05-06
 Add a description for a step both with the XML and Java configuration styles.

minkukjo:
Related Issue: https://github.com/spring-projects/spring-batch/issues/4565

## What is the problem
- The Description of a Step of a Job declared in XML cannot be accessed through Java code.

## What I Did
- Provides an interface that allows access to the description field of a Step declared in XML from Java code.
- Add a test to see if it can get the description field of a Step declared in XML.
- Fix documentation for example


spring-batch-core/src/test/java/org/springframework/batch/core/job/builder/TestTasklet.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author Minkuk Jo
+ */
+public class TestTasklet implements Tasklet {
+
+	@Override
+	public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) {
+		return RepeatStatus.FINISHED;
+	}
+
+}
```
minkukjo, line 33: I created this class expecting it to be used globally in my tests, but would it have been better to create it as an inner class in JobXmlDescriptionTests?

minkukjo:
Hi, @fmbenhassine 

I hope you're doing well. I just wanted to kindly follow up on my PR 
it's been open for about a month, and I’d be happy to make any changes if needed.

Thank you for your time and for maintaining this great project!

Best regards,  
Minkuk Jo

fmbenhassine:
Thank you for the PR, @minkukjo ! The change set in this PR seems to only add the description to the job , not the step. So I updated the title accordingly.

PR #4838, opened 2025-05-08
Fix javadoc in JdbcCursorItemReaderBuilder

itis821:
- JdbcCursorItemReaderBuilder.queryTimeout(int) -> milliseconds
  - but, AbstractCursorItemReader.setQueryTimeout(int) -> seconds


itis821:
JdbcCursorItemReaderBuilder's build() method calls JdbcCursorItemReader.setQueryTimeout(int). This setQueryTimeout(int) method is actually defined in AbstractCursorItemReader.setQueryTimeout(int). Therefore, the queryTimeout parameter from JdbcCursorItemReaderBuilder is passed as an argument to AbstractCursorItemReader.setQueryTimeout(). However, there's an issue where the javadoc specifies different units for this timeout value between JdbcCursorItemReaderBuilder and AbstractCursorItemReader.

fmbenhassine:
That's incorrect, good catch! I will rebase and merge your PR. Thank you for your contribution 👍

PR #4839, opened 2025-05-09
Remove unnecessary reflection in `RemoteChunkHandlerFactoryBean`

scordio:
Since commit 81e0c38d, `TaskletStep` exposes `getTasklet()`, making reflection in `RemoteChunkHandlerFactoryBean` unnecessary.

In addition, the `instanceof` check can be simplified to a `null` check as the `step` field is already of type `TaskletStep`.


fmbenhassine:
This is great! Thank you for your contribution 👍

Issue #4818, opened 2025-04-28
Use contextual lambdas to configure batch artefacts


fmbenhassine:
This request is about improving the builders of item readers and writers to use Lambdas for configuration options:

Current API:

```java
var reader = new FlatFileItemReaderBuilder()
 .resource(...)
 .delimited()
 .delimiter(",")
 .quoteCharacter('"')
 ...
 .build();
```

Proposal:

```java
var reader = new FlatFileItemReaderBuilder()
 .resource(...)
 .delimited ( config -> config.delimiter(',').quoteCharcter( '"' ))
 ...
 .build();
```

cc @joshlong

kwondh5217:
Hi @fmbenhassine, thank you for your guidance earlier!

I’d like to clarify the intended direction of the enhancement.

From what I understand, the idea is not just to support lambda-based configuration in FlatFileItemReaderBuilder, but to establish a general DSL-style configuration approach across all ItemReader and ItemWriter builders.

Would you recommend introducing a shared abstraction (e.g. a ConfigurerAwareBuilder base class similar to AbstractConfiguredSecurityBuilder in Spring Security) to support this pattern?

Also, in terms of behavior:

Should we throw an exception when both chaining and lambda styles are used?

Or should we allow overriding?

Or should we allow both and apply in order?

I want to align with the broader design direction before proceeding. Thank you again for your support.

fmbenhassine:
I don't think we need new builders. My initial thinking was about adding new methods to existing builders that accept `Consumer<Spec>`, similar to the one in SF here: https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/factory/BeanRegistry.java#L96

Here is also the original issue in Spring Security: https://github.com/spring-projects/spring-security/issues/5557

So we can imagine new configuration specifications like `DelimitedSpec`, `FixedLengthSpec`, etc and use them in current builders.

kwondh5217:
Thanks for the detailed guidance @fmbenhassine !
The direction is clear now. I’ll proceed with adding Consumer<Spec>-based configuration methods to the existing builders using DelimitedSpec and FixedLengthSpec style objects as discussed.

I’ll share a draft PR soon for feedback. Appreciate your support!

kwondh5217:
Hi @fmbenhassine,
I’ve submitted a pull request that addresses this issue.
Could you take a look? 🙇‍♂️
Thank you !

[...]

PR #4840, opened 2025-05-10
Support Lambda configuration in FlatFileItem builders

kwondh5217:
### Description
This PR addresses #4818

- Renamed internal builder field prefixes from `builder` to `spec` for clarity
- Added overloaded delimited(Consumer<DelimitedSpec<T>>) and formatted(Consumer<FormattedSpec<T>>) methods to FlatFileItemWriterBuilder.
- Added overloaded delimited(Consumer<DelimitedSpec<T>>) and fixedLength(Consumer<FixedLengthSpec<T>>) methods to FlatFileItemReaderBuilder.

Closes gh-4818


fmbenhassine:
Hi @kwondh5217 

Thank you for your PR! It's great to see we are on the right path to support lambda-style configuration in Spring Batch!

The overloaded methods are what I was thinking of. However:

> Renamed internal builder field prefixes from builder to spec for clarity

Those builders are actually not really internal as they are public APIs, so this change is a breaking change. Even though we are targeting this in a major release v6 in which we can break things, I am not planning to introduce any breaking changes to support this feature. 

Moreover, the fact the specification is actually a builder (since it is just a renaming), users can do things like:

```java
FlatFileItemReader<Foo> reader = new FlatFileItemReaderBuilder<Foo>().name("fooReader")
    .resource(getResource("1,2,3"))
    .delimited(
            config -> config.delimiter(",")
            .names("first", "second", "third")
            .build() // This is not correct
    ).targetType(Foo.class)
    .build();
```

The call to `build` on a specification seems incorrect to me. 

The initial idea I shared in https://github.com/spring-projects/spring-batch/issues/4818#issuecomment-2862801515 is to add new specification interfaces `DelimitedSpec` and `FixedLengthSpec` (similar to [this example](https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/factory/BeanRegistry.java#L148) from Spring Framework) and use that to configure the current builder when calling `delimited(Consumer<DelimitedSpec<T>> consumer)` or `fixedLength(Consumer<FixedLengthSpec<T>> consumer)`.

Can you please update the PR accordingly? Many thanks upfront.

kwondh5217:
@fmbenhassine Thanks for the detailed feedback! I’ll update the PR accordingly.

Issue #4844, opened 2025-05-22
Update GraalVM runtime hints


fmbenhassine:
Related to https://github.com/spring-projects/spring-framework/issues/33847, runtime hints for Spring Batch should be updated accordingly.

See https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-7.0-Release-Notes#graalvm-native-applications

goafabric:
@fmbenhassine 
As an fyi, while trying out boot 4.0.0-M1 in combination with GraalVM 24 (both local and paketo),
the native compile now breaks with the exception below.
With GraalVM 21 the compile works, but already with simple Spring Boot Apps, the Application wont start, due to another issue.

While the build will work, when overriding the CoreRuntimeHints and escaping the curly braces "\\{", the compile works.
Though i am not sure if this is the correct way to do it.


--- cut ---
Error: Error: invalid glob patterns found:
Pattern ALL_UNNAMED/org/springframework/batch/core/schema-{h2,derby,hsqldb,sqlite,db2,hana,mysql,mariadb,oracle,postgresql,sqlserver,sybase}.sql : Pattern contains unescaped character {. Pattern contains unescaped character }.

fmbenhassine:
@goafabric Thank you for reporting this! I created #4937 to track this failure and planned to fix it in the upcoming 6.0.0-M2. FYI, I have planned to add a CI build against GraalVM to avoid this kind of issues (#3871).

goafabric:
> [@goafabric](https://github.com/goafabric) Thank you for reporting this! I created [#4937](https://github.com/spring-projects/spring-batch/issues/4937) to track this failure and planned to fix it in the upcoming 6.0.0-M2. FYI, I have planned to add a CI build against GraalVM to avoid this kind of issues ([#3871](https://github.com/spring-projects/spring-batch/issues/3871)).

thank you for the very fast reply and creating the issue !
i was unsure at first if to report this, due to the milestone and how .. but its awesome that it was so easy

PR #4850, opened 2025-05-24
Update GraalVM runtime hints

therepanic:
Why? GraalVM and Spring AOT 7.0 adopts a unified glob pattern syntax and a type-only reflection model, where explicit `MemberCategory` declarations are no longer required.

Resolves: #4844


fmbenhassine:
This is great! Merged as 369652cc73d227690de065d0fe3c16a14631774f. Thank you for your contribution 👍

Issue #4854, opened 2025-06-02
Remove usage of JobFactory in JobRegistry


fmbenhassine:
Along the lines of #4847 , this issue is to remove the usage of `JobFactory` from `JobRegistry` (and `MapJobRegistry`).

The API change is as follows:

```diff
-- void register(JobFactory jobFactory) throws DuplicateJobException;
++ void register(Job job) throws DuplicateJobException;
```

`JobFactory` is not a public (but not "user facing") API, but is still used in a user facing API. To simplify things, registering a job in a registry should not be done through a factory.

PR #4857, opened 2025-06-02
Remove usage of `JobFactory` in `JobRegistry`

therepanic:
This change is to make it easier to register job not through `JobFactory` but directly.

In the tests I used `SimpleJob` for throwing jobs.

By the way, `SimpleJobRepositoryTests` and `SimpleStepFactoryBeanTests` are crashing now, but this has nothing to do with the current change.

Resolves: #4854


fmbenhassine:
Hi @therepanic , thank you for the quick PR!

It seems like we were both working on the same issue at the same time and I pushed a bigger change that includes this one (see a7f090a45d1aa055e3bdfbc4fdfd06d02ed6d0ac).

So apologies but I have to close this PR. Thank you very much for your time anyway!

Issue #3797, opened 2020-11-02
Cannot deserialize TopicPartition from JobRepository


MinJunKweon:
Hi.

I use MySQL for JobRepository. It serialize ExecutionContext as String by JacksonObjectMapper.
It seems to forcing to `Map`'s key type must be `String`. (`Map<String, Object>`)
You can see [this](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/Jackson2ExecutionContextStringSerializer.java#L130).

For Example, `SHORT_CONTEXT` in `BATCH_STEP_EXECUTION_CONTEXT`:
```
{"batch.taskletType":"org.springframework.batch.core.step.item.ChunkOrientedTasklet","topic.partition.offsets":["java.util.HashMap",{"test-topic":["java.lang.Long",42]}],"batch.stepType":"org.springframework.batch.core.step.tasklet.TaskletStep"}
```

However, `KafkaItemReader` uses `TopicPartition` as key. So It has problem in deserializing `ExecutionContext`. You can see [this](https://github.com/spring-projects/spring-batch/blob/master/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/kafka/KafkaItemReader.java#L168).

```java
        @Override
	public void open(ExecutionContext executionContext) {
		...
		if (this.saveState && executionContext.containsKey(TOPIC_PARTITION_OFFSETS)) {
			Map<TopicPartition, Long> offsets = (Map<TopicPartition, Long>) executionContext.get(TOPIC_PARTITION_OFFSETS);
			for (Map.Entry<TopicPartition, Long> entry : offsets.entrySet()) {
				this.partitionOffsets.put(entry.getKey(), entry.getValue() == 0 ? 0 : entry.getValue() + 1);
			}
		}
                ...
	}
```

```
2020-11-02 14:30:50 [main] ERROR o.s.batch.core.step.AbstractStep - Encountered an error executing step testStep in job testJob
java.lang.ClassCastException: java.lang.String incompatible with org.apache.kafka.common.TopicPartition
	at org.springframework.batch.item.kafka.KafkaItemReader$$Lambda$911/00000000EF270020.accept(Unknown Source)
	at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:684)
	at org.springframework.batch.item.kafka.KafkaItemReader.open(KafkaItemReader.java:174)
[...]

Issue #4842, opened 2025-05-22
Add support for Jackson 3


fmbenhassine:
Following Spring Framework 7 (see https://github.com/spring-projects/spring-framework/issues/33798), Spring Batch should support Jackson 3 by default.

This means `Jackson2ExecutionContextStringSerializer` should be renamed (by removing the version number from the name) and updated to Jackson 3. The Jackson 2 based serializer will be supported for v5.2.

HyunsangHan:
I’m interested in contributing to this feature. Since [Jackson 3 is currently only available as an RC version](https://github.com/FasterXML/jackson/wiki/Jackson-Releases#under-development), would it be better to wait until the official release before starting this work?

Here’s my understanding of the tasks involved:

1. Add a Jackson 3 dependency to `spring-batch-core/pom.xml`.
2. Implement a new `ExecutionContextSerializer` to JacksonExecutionContextStringSerializer based on Jackson 3.
3. Create `JacksonExecutionContextStringSerializerTests` to verify the implementation.
4. Mark the existing `Jackson2ExecutionContextStringSerializer` with the `@Deprecated` annotation. For backward compatibility, we can keep it until v5.2 and consider removing it afterwards.
5. Update the existing usages of `Jackson2ExecutionContextStringSerializer` to depend on `JacksonExecutionContextStringSerializer`.

If this plan sounds good to you, I’d be happy to contribute.

fmbenhassine:
@HyunSangHan I worked on this and forgot about your message.. Apologies for that. Thank you anyway for your offer to help!

> This means Jackson2ExecutionContextStringSerializer should be renamed (by removing the version number from the name) and updated to Jackson 3. The Jackson 2 based serializer will be supported for v5.2.

After discussion with the broader team, we decided to support both Jackson 2 (in deprecated form) and Jackson 3 in the v6 generation to give our users time to migrate. Therefore, a new serialiser named `JacksonExecutionContextStringSerializer` (based on Jackson 3) will be added next to the `Jackson2ExecutionContextStringSerializer`.

PR #4863, opened 2025-06-05
Fix KafkaItemReader ClassCastException during ExecutionContext deserialization

noojung:
Using Jackson2ExecutionContextStringSerializer causes KafkaItemReader to throw a ClassCastException.

Because KafkaItemReader stores a Map<TopicPartition, Long> in the ExecutionContext, the serializer converts all map keys to strings.

To fix this, I changed it to store only the partition number (as a string) instead of the full TopicPartition object.

Resolves: #3797

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/kafka/KafkaItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/kafka/KafkaItemReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2023 the original author or authors.
+ * Copyright 2019-2025 the original author or authors.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


fmbenhassine:
> Using Jackson2ExecutionContextStringSerializer causes KafkaItemReader to throw a ClassCastException.

`Jackson2ExecutionContextStringSerializer` is deprecated in favor of `JacksonExecutionContextStringSerializer` (based on Jackson 3, see #4842). Can you please try with the new serializer and see if the issue persists?

noojung:
@fmbenhassine I've tested it, and the same ClassCastException still occurs with the new serializer.

I've written a sample test case on my branch: https://github.com/noojung/spring-batch/commit/c91f775ccb8804d1967d2db3787e67d7970d338a

Please let me know if you have any feedback.

fmbenhassine:
LGTM 👍 Rebased and merged as 3080a6ec8fa58c7991af3170b9024bfb6cd8c5aa. Thank you for your contribution!

PR #4879, opened 2025-06-13
remove JobRegistrySmartInitializingSingleton usage

patrickwinti:
JobRegistrySmartInitializingSingleton is deprecated and not needed anymore.
Removed it from tests and samples.


fmbenhassine:
That's great! Thank you very much for your first contribution 👍

PR #4887, opened 2025-06-13
Refactor JobKeyGenerator to use JobParameters instead of generic type

### LINKED ISSUE #4886: Remove unnecessary generic from JobKeyGenerator interface
The `JobKeyGenerator<T>` interface currently uses a generic type parameter `<T>` to represent the source used to generate a job key. However, in practice, the implementations and usages of this interface rely on JobParameters as the source type.

Since the generic parameter is not providing meaningful flexibility and introduces unnecessary complexity (e.g., requiring casts or wildcard types in consumers), it would be cleaner to refactor the interface to:

```
public interface JobKeyGenerator {
    String generateKey(JobParameters source);
}
```

patrickwinti:
The JobKeyGenerator<T> interface currently uses a generic type parameter <T> to represent the source used to generate a job key. However, in practice, all known implementations and usages of this interface rely specifically on JobParameters as the source type.

resolves #4886

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/DefaultJobKeyGenerator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/DefaultJobKeyGenerator.java
@@ -36,7 +36,7 @@ import org.springframework.util.DigestUtils;
  * @author Mahmoud Ben Hassine
  * @since 2.2
  */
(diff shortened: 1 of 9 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as b1e4a94019c81d022d0823472e7f814bac572f90. Thank you for your contribution!

PR #4897, opened 2025-07-01
Refactor JDBC DAOs to use JdbcClient where feasible

### LINKED ISSUE #4804: Refactor JDBC DAOs to use named parameters instead of positional parameters
It will eliminate duplicate parameters from java code although JDBC prepared statement does not.
And migrate from `JdbcTemplate` to `NamedParameterJdbcTemplate` or fluent `JdbcClient`?

quaff:
Closes GH-4804

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcJobInstanceDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcJobInstanceDao.java
@@ -32,7 +32,6 @@ import org.springframework.batch.core.repository.dao.AbstractJdbcBatchMetadataDa
 import org.springframework.batch.core.repository.dao.JobInstanceDao;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.dao.DataAccessException;
-import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.dao.OptimisticLockingFailureException;
 import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.core.RowMapper;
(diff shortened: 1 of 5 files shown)


quaff:
`JdbcTemplate` is only used for batch update which is not covered by `JdbcClient`.

spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcJobInstanceDao.java
```diff
@@ -229,43 +233,32 @@ public List<JobInstance> extractData(ResultSet rs) throws SQLException, DataAcce
...
 	@Override
 	public long getJobInstanceCount(@Nullable String jobName) throws NoSuchJobException {
-
-		try {
-			return getJdbcTemplate().queryForObject(getQuery(COUNT_JOBS_WITH_NAME), Long.class, jobName);
-		}
-		catch (EmptyResultDataAccessException e) {
-			throw new NoSuchJobException("No job instances were found for job name " + jobName);
-		}
+		return getJdbcClient().sql(getQuery(COUNT_JOBS_WITH_NAME)).param("jobName", jobName).query(Long.class).single();
```
fmbenhassine, line 261: Will this throw a `NoSuchJobException` as before if the result set is empty?

quaff: Actually count statement never generate empty result set.

fmbenhassine:
> JdbcTemplate is only used for batch update which is not covered by JdbcClient.

It's unfortunate to have to use `JdbcClient` and `JdbcTemplate` at the same time. In hindsight, do you think there is an added value in using `JdbcClient` if we can't remove the usage of `JdbcTemplate` altogether (I mean a 100% replacement)?

quaff:
> > JdbcTemplate is only used for batch update which is not covered by JdbcClient.
> 
> It's unfortunate to have to use `JdbcClient` and `JdbcTemplate` at the same time. In hindsight, do you think there is an added value in using `JdbcClient` if we can't remove the usage of `JdbcTemplate` altogether (I mean a 100% replacement)?

Let's wait for https://github.com/spring-projects/spring-framework/issues/33843

Issue #4935, opened 2025-08-04
FlatFileItemWriterBuilder - RecordFieldExtractor does not honor names() for Java Record types


DevSeongmin:
Bug description
When using FlatFileItemWriterBuilder with Java Record types, specifying partial fields via .names() does not work as expected. Internally, the RecordFieldExtractor is created but the .setNames() method is not called, causing the extractor to output all record fields instead of only the specified subset.

Environment
- Spring Batch version: 6.x (e.g., 6.0.0-M1)
- Java version: 17 or higher
- Build tool: Maven or Gradle
- Database: In-memory database (H2) or none for this test
- Operating system: Any
  

Steps to reproduce
Define a Java Record class, e.g.:

```
public record UserInfo(
    String username,
    String email,
    int age,
    String phoneNumber
) {}
```

Create and configure the writer with .sourceType(UserInfo.class) and partial fields with .names("username", "phoneNumber").
Write items using this writer and verify that the output file contains all fields instead of only the specified ones.

Expected behavior
Only the fields specified in .names() should be extracted and written to the output file, excluding all others.

Minimal Complete Reproducible example

```
@Bean
@StepScope
public FlatFileItemWriter<UserInfo> userInfoWriter(
        @Value("#{jobParameters['outputDir']}") String outputDir) {
    return new FlatFileItemWriterBuilder<UserInfo>()
            .name("userInfoWriter")
            .resource(new FileSystemResource(outputDir + "/user_infos.csv"))
            .delimited()
            .delimiter(",")
            .sourceType(UserInfo.class)
            .names("username", "phoneNumber")
            .build();
}
```

// After writing, the output CSV still contains all fields instead of only username and phoneNumber.

Expected.
```
Alice Johnson,555-123-4567
Bob Smith,555-987-6543
Carol Williams,555-234-5678
David Brown,555-345-6789
Eve Davis,555-456-7890
```

Result.
```
Alice Johnson,alice.johnson@example.com,28,555-123-4567
Bob Smith,bob.smith@example.com,35,555-987-6543
Carol Williams,carol.williams@example.com,42,555-234-5678
David Brown,david.brown@example.com,30,555-345-6789
Eve Davis,eve.davis@example.com,27,555-456-7890

```

Additional Notes
Powerd by KILL-9 💀

fmbenhassine:
Thank you for opening this issue. Is this the same as #4916?

DevSeongmin:
[...]

PR #4936, opened 2025-08-04
Fix: FlatFileItemWriterBuilder - Ensure names() is respected for RecordFieldExtractor

DevSeongmin:
Description
When using FlatFileItemWriterBuilder with Java Record types, specifying .names() is currently required, but RecordFieldExtractor ignores those names due to a missing .setNames() call. This causes all record components to be extracted regardless of the .names() setting, leading to unexpected behavior.

This patch fixes the issue by invoking .setNames() on RecordFieldExtractor when names are specified, aligning its behavior with BeanWrapperFieldExtractor used for non-record types.

No breaking changes are introduced.

Changes
Added .setNames() call on newly created RecordFieldExtractor if names are provided.

Maintains consistency and backward compatibility in builder API.

Motivation
Enables partial field extraction from Java Records via .names() method, matching the behavior for Java Beans.

Resolves a long-standing inconsistency and confusion reported by users and educators.

Additional Notes
Powerd by KILL-9 💀

Related Issue
This PR addresses issue #4935


fmbenhassine:
That was fixed in 0eeacd583ffbb2d47dd6ed9bc76f914fd320b496. Thank you for your PR anyway 🙏

PR #4938, opened 2025-08-04
Reset CompositeItemReader iterator on open to support multiple executions

### LINKED ISSUE #4926: CompositeItemReader Does Not Reset Iterator for Subsequent Job Executions
**Bug description**

The [CompositeItemReader](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/CompositeItemReader.java) implementation  has an issue where the iterator `delegatesIterator` does not reset after the first execution of a job. This causes subsequent executions of the job (e.g., in a scheduled method) to skip calling the delegate readers because the iterator remains at the end.

**Environment**

Spring Batch version : 5.2.2

**Steps to reproduce**

1. Create a Spring Batch job that uses `CompositeItemReader` with multiple delegate readers.
2. Launch the job  in a scheduled method using : jobLauncher.run(job, jobParameters);
3. Allow the scheduled method to trigger the job a second time.
4. Observe that the `CompositeItemReader` does not call the delegate readers again because the iterator is at the end.

**Expected behavior**

The `CompositeItemReader`  should reset its iterator and call the delegate readers for each job execution, regardless of whether it is the first or subsequent execution.

**Proposed Solution**

Modify the `open`  method of `CompositeItemReader` to reset the iterator `delegatesIterator`  and reinitialize the `currentDelegate` to the first reader. This ensures that the iterator starts fresh for each job execution.

```java
	@Override
	public void open(ExecutionContext executionContext) throws ItemStreamException {
		for (ItemStreamReader<? extends T> delegate : delegates) {
			delegate.open(executionContext);
		}
		initIteratorAndCurrentDelegate();
	}

	private void initIteratorAndCurrentDelegate() {
		this.delegatesIterator = this.delegates.iterator();
		this.currentDelegate = this.delegatesIterator.hasNext() ? this.delegatesIterator.next() : null;
	}
```

therepanic:
Reset delegatesIterator and currentDelegate in open() to ensure that CompositeItemReader works correctly on repeated job executions. Without this reset, the reader skips all delegates after the first run.

Closes: gh-4926

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/CompositeItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/support/CompositeItemReader.java
@@ -29,6 +29,7 @@ import org.springframework.batch.item.ItemStreamReader;
  *
  * @author Mahmoud Ben Hassine
  * @author Elimelec Burghelea
+ * @author Andrey Litvitski
  * @param <T> type of objects to read
  * @since 5.2
  */
@@ -36,7 +37,7 @@ public class CompositeItemReader<T> implements ItemStreamReader<T> {
 
 	private final List<ItemStreamReader<? extends T>> delegates;
 
-	private final Iterator<ItemStreamReader<? extends T>> delegatesIterator;
+	private Iterator<ItemStreamReader<? extends T>> delegatesIterator;
 
 	private ItemStreamReader<? extends T> currentDelegate;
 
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Since the composite reader is an item stream, it is expected that the initialisation of the delegate is done in the `open` method and not in the constructor (in fact, clients are expected to call `open` before using the item stream). Therefore the initialisation of the delegate should rather be removed from the constructor and moved to the `open` method.

I am closing this PR in favor of #5169. Thank you for your contribution anyway 🙏

PR #4944, opened 2025-08-10
GH4876: Add recover method to JobOperator

### LINKED ISSUE #4876: Add API to recover failed job executions
This was initially discussed with @Nhoutain in https://github.com/spring-projects/spring-batch/discussions/4831

Abrupt shutdowns of Spring Batch jobs leave the job execution in a running state at the job repository level as Spring Batch doesn't have a chance to update the status correctly (see [Aborting a Job](https://docs.spring.io/spring-batch/reference/job/advanced-meta-data.html#aborting-a-job)). Restarting such failed jobs requires a manual database update as follows:

```
> update BATCH_JOB_EXECUTION set status = 'FAILED', END_TIME = 'non null value' where job_execution_id = X;
> update BATCH_STEP_EXECUTION set status = 'FAILED' where job_execution_id = X and step_name='failed step name';
```

There is already an API in the `JobOperator` to abandon a job execution, so it would be great to provide a new method to mark a job as failed to be able to restart it (which is conceptually the opposite of abandoning a job execution, I think `recover` would be a good name for such an operation). The goal is to provide an API to mark a job as failed as opposed to asking users to manually execute SQL statements against the database. Providing an API is also necessary to support that feature consistently with other job repositories as well (thinking about No SQL stores for instances).

thelightway24:
## Description

This PR introduces a new `recover` method to the `JobOperator` interface and its implementations. 
This method allows to recover a failed job execution based on its ID. This is useful for restarting a failed job from the point where it left off.

This PR also adds unit tests for the new `recover` method in `JobOperator`.

## Changes

- Add `recover` method to `JobOperator` interface.
- Implement the `recover` method in `SimpleJobOperator`.
- Add unit tests for `JobOperator.recover` in `TaskExecutorJobOperatorTests`.
- Add `recover` method to `CommandLineJobOperator`
- Add unit tests for `CommandLineJobOperator.recover` in `CommandLineJobOperatorTests`.

## Additional Notes
At SimpleJobOperator.java line 403, when a job execution is already recovered,
I currently throw an existing exception class.
I’m considering whether it would be better to:

Define a new exception class (e.g., AlreadyRecoveredException) — similar to how abandon throws an exception, or
Skip throwing an exception and instead log the event while returning the JobExecution.

For now, I’ve chosen to throw an exception for consistency with abandon

## Related Issue

Resolves #4876


spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
```diff
@@ -391,6 +392,45 @@ public JobExecution abandon(JobExecution jobExecution) throws JobExecutionAlread
...
+
+		if (jobExecution.getExecutionContext().containsKey("recovered")) {
+			if (logger.isInfoEnabled()) {
+				logger.info("already recovered job execution: " + jobExecution);
+			}
+			throw new UnexpectedJobExecutionException("JobExecution is already recovered");
+		}
+
+		BatchStatus jobStatus = jobExecution.getStatus();
+		if (jobStatus == BatchStatus.COMPLETED || jobStatus == BatchStatus.ABANDONED) {
```
fmbenhassine, line 407: `BatchStatus.UNKNOWN` is also non recoverable, see https://github.com/spring-projects/spring-batch/blob/d95397faf023ee3293ee10b41977231734a0f5d1/spring-batch-core/src/main/java/org/springframework/batch/core/step/AbstractStep.java#L291. I will add that on merge.

spring-batch-core/src/test/java/org/springframework/batch/core/launch/support/TaskExecutorJobOperatorTests.java
```diff
@@ -427,6 +428,52 @@ void testAbortNonStopping() {
...
+	void testRecover() {
+		JobInstance jobInstance = new JobInstance(123L, job.getName());
+		JobExecution jobExecution = new JobExecution(jobInstance, 111L, jobParameters);
+		jobExecution.setStatus(BatchStatus.STARTED);
+		jobExecution.createStepExecution("step1").setStatus(BatchStatus.STARTED);
+		when(jobRepository.getJobExecution(111L)).thenReturn(jobExecution);
+		when(jobRepository.getLastJobExecution(jobInstance)).thenReturn(jobExecution);
+		JobExecution recover = jobOperator.recover(jobExecution);
+		assertEquals(BatchStatus.FAILED, recover.getStatus());
+		assertTrue(recover.getExecutionContext().containsKey("recovered"));
```
fmbenhassine, line 441: We need to asset that the end time is not null as well. I will take care of adding it.

fmbenhassine:
This is awesome! Thank you for your contribution 🙏

I added a couple inline comments about some minor changes that I will take care of on merge.

> At SimpleJobOperator.java line 403, when a job execution is already recovered,
I currently throw an existing exception class.
I’m considering whether it would be better to:
Define a new exception class (e.g., AlreadyRecoveredException) — similar to how abandon throws an exception, or
Skip throwing an exception and instead log the event while returning the JobExecution.
For now, I’ve chosen to throw an exception for consistency with abandon

Unlike attempting to abandon a running execution which is an error, attempting to recover an already recovered execution does not really require an exception, so logging a warning is enough. I will update the code and tests accordingly.

injae-kim:
Nice work @thelightway24 !!
Thanks for your detailed review @fmbenhassine 👍👍👍

thelightway24:
Thanks for the review and fix, really appreciate it.
I’ll be more thorough next time.

Issue #4910, opened 2025-07-16
Incorrect handling of job parameters when using a JobParametersIncrementer


fmbenhassine:
The `JobParametersIncrementer` concept is a useful abstraction to use when there is a natural sequence of job instances (hourly, daily, etc). The goal of this abstraction is to bootstrap the initial set of job parameters and let the framework calculate the parameters of the next instance in the sequence. This was clear since the initial design of the `JobOperator#startNextInstance(String jobName)` method (notice the lack of job parameters in the signature). The concept is similar to database sequences that are initialized once by the user and incremented automatically by the database system (ie not altered by the user anymore).Therefore, it does not make sense to provide job parameters when using an incrementer.

Unfortunately, this feature has been incorrectly used through the command line runners of Spring Batch and Spring Boot for many years and I have seen users messing with job parameters between sequences, which goes against the initial intent of the incrementer concept, and which in turn caused several confusing issues like #882, https://github.com/spring-projects/spring-boot/issues/22602 and https://github.com/spring-projects/spring-boot/issues/14484. If someone starts to modify the sequence's logic and altering job parameters in between instances, (s)he should not use an incrementer in the first place.

v6 is a good opportunity to fix that. When a job parameters incrementer is attached to a job definition, the parameters of the next instance should be calculated by the framework, and any additional parameters supplied by the user should be ignored with a warning.

PR #4946, opened 2025-08-15
Gh 4755: Fix wrong behavior of JobRepository with empty identifying job parameters

### LINKED ISSUE #4755: Spring batch allows job without parameters to restart even if the job already completed successfully
### Discussed in https://github.com/spring-projects/spring-batch/discussions/4694

<sup>Originally posted by **ELMORABITYounes** October 28, 2024</sup>
Right now even if a job was completed successfuly, spring batch allow It to be restarted if It contains no identifying params as shown here:

```				
if (!identifyingJobParameters.isEmpty()                                                        
		&& (status == BatchStatus.COMPLETED || status == BatchStatus.ABANDONED)) {            
	throw new JobInstanceAlreadyCompleteException(                                             
			"A job instance already exists and is complete for identifying parameters="       
					+ identifyingJobParameters + ".  If you want to run this job again, "    
					+ "change the parameters.");                                             
}                                                                                              
```

I am wondering why is that done like this? I mean if the job already completed why It does not throw JobInstanceAlreadyCompleteException? Shouldn't second job instance without param considered the same and hence not allow It to be restarted if already succeeded?
</div>

isanghaessi:
## Summary
This PR fixes #4755 .
Problem is, job with having zero identifier job parameters be able to restart.
Job what having COMPLETED status should not be able to restart with same Job Parameters.

## Changes
- Removed exception condition for job parameters with no identifier job parameters.
- Test case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Latest news
 
+* July 23, 2025: [Spring Batch 6.0.0 M1 is out!](https://spring.io/blog/2025/07/23/spring-batch-6) 
 * March 19, 2025: [Spring Batch 5.2.2 available now](https://spring.io/blog/2025/03/19/spring-batch-5-2-2-available-now) 
 * December 18, 2024: [Spring Batch 5.1.3 and 5.2.1 available now](https://spring.io/blog/2024/12/18/spring-batch-5-1-3-and-5-2-1-available-now) 
 * November 24, 2024: [Bootiful Spring Boot 3.4: Spring Batch](https://spring.io/blog/2024/11/24/bootiful-34-batch) 
@@ -10,8 +11,6 @@
 # Spring Batch [![build status](https://github.com/spring-projects/spring-batch/actions/workflows/continuous-integration.yml/badge.svg)](https://github.com/spring-projects/spring-batch/actions/workflows/continuous-integration.yml)
 
 Spring Batch is a lightweight, comprehensive batch framework designed to enable the development of robust batch applications vital for the daily operations of enterprise systems.  Spring Batch builds upon the productivity, POJO-based development approach, and general ease of use capabilities people have come to know from the [Spring Framework](https://github.com/spring-projects/spring-framework), while making it easy for developers to access and leverage more advanced enterprise services when necessary.
-
-If you are looking for a runtime orchestration tool for your Batch applications, or need a management console to view current and historic executions, take a look at [Spring Cloud Data Flow](https://cloud.spring.io/spring-cloud-dataflow/).  It is an orchestration tool for deploying and executing data integration based microservices including Spring Batch applications.
 
 # Getting Started
 


isanghaessi:
Hi, @fmbenhassine.
Can you please review this PR when has chance?
I am very interesting spring batch project and using it.
I hope this ambigious behavior will be fixed.
Thanks.

fmbenhassine:
Hi @isanghaessi ,

Thank you for the PR. #4755 is planned for v6, so I will check your PR soon. I will keep you posted.

fmbenhassine:
I am looking into including this PR in v6.0.0-RC2 if possible. Can you please rebase it on the latest `main`? Thank you upfront.

isanghaessi:
Hi @fmbenhassine,
Since there were too many changes in the middle, I force pushced changes.

I moved my changes from `SimpleJobRepository` from `TaskExecutorJobLauncher`.
I saw restartability logic moved from JobRepository to JobOperator - 90d89595.
```java
					if (!identifyingJobParameters.isEmpty()
							&& (status == BatchStatus.COMPLETED || status == BatchStatus.ABANDONED)) {
						throw new JobInstanceAlreadyCompleteException(
								"A job instance already exists and is complete for identifying parameters="
										+ identifyingJobParameters + ".  If you want to run this job again, "
										+ "change the parameters.");
					}
```
It is hard to change test cases in core package because they are almost broken and disabled.
And so it is hard to add test case.
At least, I checked that test cases in other packages are not broken.

Please let me know if there is anything I am missing.

fmbenhassine:
Thank you for the updates! LGTM 👍 Rebased and merged as 5225249585fec7e479bf4b3194974d96a848c3c0. I added a test in 0564ce6293e5178b12aa95b7bce5a461a38e4eb0. Thank you for your contribution !

goafabric:
Unfortunately this was added last minute with RC2,
but at least with an H2 Database for local testing now things break.

With a simple List<job> and a for loop
jobs.forEach(job -> jobOperator.start(job, new JobParameters()));

The second job started will bail out with the JobInstanceAlreadyCompleteException.

- The reasons seams to be that internally a "ResourceLessJobrepository" is created
- And this will just always return the last instance created when asking for "jobRepository.getJobInstance".
- which for the first job from the list will return null 
- And for the second, it will return the previous job => Running into the exception.

I don't know if all if this the desired behaviour, but this never happened in any previous version of spring batch and breaks existing code.

fmbenhassine:
@goafabric This was a serious bug and I wonder how it survived all these generations (see https://github.com/spring-projects/spring-batch/issues/4755#issuecomment-2626989125).

In a snippet like `jobs.forEach(job -> jobOperator.start(job, new JobParameters()));`, how do you distinguish which job instance has failed when things go wrong? This is just impossible since empty parameter sets yield the same hash, see https://github.com/spring-projects/spring-batch/blob/b1e4a94019c81d022d0823472e7f814bac572f90/spring-batch-core/src/test/java/org/springframework/batch/core/DefaultJobKeyGeneratorTests.java#L74

I am really glad we fixed this in v6 because it was confusing to many people (myself included). 

In your case, you can attach a `JobParametersIncrementer` to the job (only if job instances have a natural sequence, beware of #4910), or add a discriminator parameter to distinguish job instances (I typically use a random UUID).

_EDIT_: This is not related to the job repository implementation. `ResourceLessJobRepository` is not the issue here, and is **not** designed for state management in the first place (it is designed for a single execution of a job: 1 job instance + 1 job execution + n step executions, we might need to make this more explicit in the docs).

PR #4952, opened 2025-08-17
Improve error message when using ResourcelessJobRepository with partitioned steps

### LINKED ISSUE #4732: Inaccurate error message when using ResourcelessJobRepository with a partitioned step
**Bug description**
The ResourcelessJobRepository cannot be used with a Partitioner:
```
[main] ERROR org.springframework.batch.core.step.AbstractStep - Encountered an error executing step step in job partitionJob
org.springframework.batch.core.JobExecutionException: Cannot restart step from STARTING status.  The old execution may still be executing, so you may need to verify manually that this is the case.
```

**Steps to reproduce**
Simply change `spring-batch-samples/src/main/resources/simple-job-launcher-context.xml` to use `ResourcelessJobRepository` and then run `spring-batch-samples/src/test/java/org/springframework/batch/samples/partition/file/PartitionFileJobFunctionalTests.java`

khj68:
## Summary
This PR improves the error message when attempting to use ResourcelessJobRepository with partitioned steps. Previously, users would see a confusing error about 'Cannot restart step from STARTING status' which didn't clearly indicate the actual problem.

## Issue
Fixes #4732

## Changes
- Added explicit check in SimpleStepExecutionSplitter.split() to detect ResourcelessJobRepository usage
- Throws a clear JobExecutionException with informative message explaining that ResourcelessJobRepository cannot be used with partitioned steps
- Added test case to verify the new behavior

## Before
Users would see:
'Cannot restart step from STARTING status. The old execution may still be executing, so you may need to verify manually that this is the case.'

## After  
Users now see:
'ResourcelessJobRepository cannot be used with partitioned steps as it does not support execution context. Please use a different JobRepository implementation that supports batch meta-data storage for partitioned steps.'

## Testing
- Added testResourcelessJobRepositoryThrowsException() test in SimpleStepExecutionSplitterTests
- All existing tests continue to pass
- The test verifies that the appropriate exception is thrown with the correct error message

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/SimpleStepExecutionSplitter.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/partition/support/SimpleStepExecutionSplitter.java
@@ -33,6 +33,7 @@ import org.springframework.batch.core.partition.PartitionNameProvider;
 import org.springframework.batch.core.partition.Partitioner;
 import org.springframework.batch.core.partition.StepExecutionSplitter;
 import org.springframework.batch.core.repository.JobRepository;
+import org.springframework.batch.core.repository.support.ResourcelessJobRepository;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


fmbenhassine:
This was fixed as mentioned in https://github.com/spring-projects/spring-batch/issues/4732#issuecomment-3532321449. Thank you for your PR anyway 🙏

PR #4967, opened 2025-09-04
Clarify Javadoc for FaultTolerantStepBuilder#skipLimit

### LINKED ISSUE #4963: FaultTolerantStepBuilder#skipLimit Javadoc is misleading with custom skipPolicy
The Javadoc for [FaultTolerantStepBuilder#skipLimit(int) in Spring Batch 5.1.0](https://docs.spring.io/spring-batch/docs/5.1.0/org/springframework/batch/core/step/builder/FaultTolerantStepBuilder.html) currently says:

Sets the maximum number of failed items to skip before the step fails. Ignored if an explicit skipPolicy(SkipPolicy) is provided.  ￼
(Identical wording appears in the attached API PDF we consulted.)

This wording can be misread as: “skipLimit is always irrelevant whenever a custom SkipPolicy is set.”
A clearer description is that skipLimit participates only when the default, limit-based skip policy is in effect; once a custom SkipPolicy is provided, that default policy is not used, so the configured skipLimit will no longer have any effect.

This aligns with how Spring Batch models skip behavior: there is a built-in limit-checking skip policy (LimitCheckingItemSkipPolicy) used to enforce the max number of skips; when users install their own SkipPolicy, that default policy is not applied. See the API docs for LimitCheckingItemSkipPolicy (describes the limit-based behavior).  ￼ ￼

Why this matters

Developers reading “Ignored if… provided” often wonder whether their configured skipLimit is silently dropped (vs. simply not applicable because the default policy is replaced). Clarifying the wording avoids confusion when migrating from skipLimit/skip(...) to a custom SkipPolicy.

Suggested doc tweak

Update the Javadoc of skipLimit(int) to something along the lines of:

Sets the maximum number of failed items to skip before the step fails. Effective only when the default limit-based skip policy is used. If a custom SkipPolicy is provided via skipPolicy(SkipPolicy), this limit does not apply.

Optionally, add a short note in skipPolicy(SkipPolicy) pointing out that installing a custom policy replaces the default limit-checking behavior.

References
	•	Spring Batch 5.1.0 API — FaultTolerantStepBuilder#skipLimit: wording that says “Ignored if an explicit skipPolicy(SkipPolicy) is provided.”  ￼
	•	Same page from the 5.1.0 API PDF (attached in this report): identical wording.
	•	LimitCheckingItemSkipPolicy — documents the default limit-checking skip behavior.

Tahnks!

rla124:
Fixes #4963

## Summary

The current Javadoc for the `FaultTolerantStepBuilder#skipLimit` method is misleading. It states that the skip limit is "ignored" when a custom `SkipPolicy` is provided, which can be misinterpreted by users to mean that the value is always irrelevant in such cases.

## Solution

This pull request improves the Javadocs for both `skipLimit()` and the related `skipPolicy()` method to make their relationship and behavior crystal clear. 

Based on valuable community feedback, the updated documentation now provides better context on default behavior and clarifies how setting a custom policy overrides the default settings.

## Key Changes

- **In `skipLimit()`:**
  - The description has been reworded to remove the ambiguous term "ignored".
  - Details about the default behavior have been added, explicitly mentioning the `LimitCheckingItemSkipPolicy` and the default limit value of 10.

- **In `skipPolicy()`:**
  - A new note has been added to its Javadoc, providing a clear cross-reference that it overrides the behavior configured via `skipLimit()`.

- The copyright year in the license header has been updated as per the contributor guidelines.

- Adding a `@see` tag that links to the `SkipPolicy` interface for better context and navigation.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/FaultTolerantStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/FaultTolerantStepBuilder.java
@@ -305,10 +305,18 @@ public class FaultTolerantStepBuilder<I, O> extends SimpleStepBuilder<I, O> {
 	}
 
 	/**
-	 * Sets the maximum number of failed items to skip before the step fails. Ignored if
-	 * an explicit {@link #skipPolicy(SkipPolicy)} is provided.
-	 * @param skipLimit the skip limit to set. Default is 10.
-	 * @return this for fluent chaining
+	 * Sets the maximum number of failed items to skip before the step fails.
+	 * The default value is 10.
+	 * <p>
+	 * This limit is enforced using the default
+	 * {@link org.springframework.batch.core.step.skip.LimitCheckingItemSkipPolicy}.
+	 * If a custom {@link SkipPolicy} is provided
+	 * via {@link #skipPolicy(SkipPolicy)},
+	 * this limit will not be enforced by the step directly,
+	 * but it can be implemented to be honored by the custom policy.
+	 * @param skipLimit the maximum number of failed items to skip.
+	 * @return this for fluent chaining.
+	 * @see SkipPolicy
 	 */
 	public FaultTolerantStepBuilder<I, O> skipLimit(int skipLimit) {
 		this.skipLimit = skipLimit;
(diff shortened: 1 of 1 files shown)


fmbenhassine:
LGTM 👍 Merged as 82bd3a2bad3d43771d0df5cdd190c1ebd2a8e5f7. Thank you for your contribution!

PR #4982, opened 2025-09-22
Fix/null pointer exception parallel remove map step registry

### LINKED ISSUE #4981: Race Condition MapStepRegistry unregisterStepsFromJob and getStep
**Bug description**
calling unregisterStepsFromJob and getStep from two Threads in parallel can lead to a NullPointerException

**Environment**
Spring Batch Version 5.2.3
All Jdk Versions, I tried jdk 17 and 21

**Steps to reproduce**
Here is an example which can lead to this race condition:
````
StepRegistry registry = new MapStepRegistry();
registry.register("foo", List.of(new TaskletStep("first")));
Thread first = new Thread() {
	@Override
	public void run() {
		registry.unregisterStepsFromJob("foo");
	}
};
first.start();
try {
	registry.getStep("foo", "first");
}
catch (NoSuchJobException e) {
}
````

**Expected behavior**
getStep should either return the Step or throw a NoSuchJobException.

[Failing Tests](https://github.com/spring-projects/spring-batch/commit/7b004fb50a8d7911cba3faaa2a4b506adfda7b82)

I used vmlens.com to make the execute all thread intelraving and make the test reproducable

ThomasKrieger:
Resolves #4981 

replaced map.containsKey with get and != null in MapStepRegistry
And test with [vmlens](https://vmlens.com) to test the race condition

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/MapStepRegistry.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/MapStepRegistry.java
@@ -65,11 +65,11 @@ public class MapStepRegistry implements StepRegistry {
 	public Step getStep(String jobName, String stepName) throws NoSuchJobException {
 		Assert.notNull(jobName, "The job name cannot be null.");
 		Assert.notNull(stepName, "The step name cannot be null.");
-		if (!map.containsKey(jobName)) {
+		final Map<String, Step> jobSteps = map.get(jobName);
+		if (jobSteps == null) {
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Hi @ThomasKrieger ,

Thank you for opening this PR! `MapStepRegistry` is now deprecated for removal and we are not planning to put any effort in maintaining it anymore. Therefore, I am going to close this PR.

However, if you manage to find other issues thanks to vmlens, you are welcome to contribute other PRs. Please note that we have no plan to add vmlens as a dependency and use its maven plugin, but we are open to merge any bug-fix that it may detect.

Keep the good work up on vmlens!

Issue #4908, opened 2025-07-15
Setting FormattedBuilder's sourceType to record forces unnecessary method calls


apptie:
When specifying record as the sourceType for FlatFileItemWriter, RecordFieldExtractor operates due to the following conditional statement.

```java
if (this.sourceType != null && this.sourceType.isRecord()) {
	this.fieldExtractor = new RecordFieldExtractor<>(this.sourceType);
}
```
[source code link](https://github.com/spring-projects/spring-batch/blob/584c3c6eaa250cca3d0068ea28fce073fc1bc2e3/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java#L393-L395)

RecordFieldExtractor internally calls targetType.getRecordComponents() to return the record's fields as RecordComponent.
In this process, the values specified in FormattedBuilder.names() are ignored.

It was confirmed that in the existing test FlatFileItemWriterBuilderTests.testSetupFormatterLineAggregatorWithRecordItemType(), even when arbitrary values are specified through names(), RecordFieldExtractor ignores the values specified in names() and properly retrieves the record's fields.

<img width="1036" height="568" alt="Image" src="https://github.com/user-attachments/assets/acf35829-0d05-4271-a089-83963bc211f8" />
<img width="833" height="427" alt="Image" src="https://github.com/user-attachments/assets/9b5432c9-5ca9-46af-84ee-7cad7af2b9cf" />

Therefore, if record is specified as the sourceType for FlatFileItemWriter, I think there is no need to call FormattedBuilder.names().

However, since the return type of FormattedBuilder.sourceType() is FormattedBuilder, in order to continue method chaining for FlatFileItemWriter, FormattedBuilder.fieldExtractor() or FormattedBuilder.names() must be called.

This appears to be an unnecessary process.

fmbenhassine:
> In this process, the values specified in FormattedBuilder.names() are ignored.

Yes, this is a duplicate of #4916

PR #5009, opened 2025-10-04
Fix RecordFieldExtractor to honor names() configuration in FlatFileItemWriterBuilder

### LINKED ISSUE #4916: RecordFieldExtractor in FlatFileItemWriterBuilder doesn't reflect names() setting
This is related to [spring-projects/spring-batch#4908](https://github.com/spring-projects/spring-batch/issues/4908), which points out that using names() with a record type and sourceType() might seem redundant. While that discussion raises a valid concern, this issue is coming from a slightly different angle: it would be helpful if the names() setting actually worked when a record is used with sourceType(). That way, developers wouldn't be surprised when their field selections are silently ignored.

---
### **Bug description**
When using FlatFileItemWriterBuilder’s build() method with sourceType() set to a record class and specifying field names via names(), the RecordFieldExtractor used internally ignores the names() configuration.
In contrast, when sourceType() is not specified, the BeanWrapperFieldExtractor honors the names() configuration as expected.

This inconsistency can cause confusion, as the same record type behaves differently depending on whether sourceType() is provided.

---
### **Steps to reproduce**
```
public record MyRecord(String name, int age, String address) {}

FlatFileItemWriter<MyRecord> writer = new FlatFileItemWriterBuilder<MyRecord>()
    .name("myRecordWriter")
    .resource(new FileSystemResource("output.csv"))
    .sourceType(MyRecord.class)  // triggers RecordFieldExtractor
    .names("name", "age")        // currently ignored
    .delimited()
    .build();
```
Expected output: name, age
Actual output: name, age, address

---
### **Suggested Fix**
Update the builder to pass names into RecordFieldExtractor:

```
if (this.sourceType != null && this.sourceType.isRecord()) {
    RecordFieldExtractor<T> extractor = new RecordFieldExtractor<>(this.sourceType);
    extractor.setNames(this.names.toArray(new String[this.names.size()]));
    this.fieldExtractor = extractor;
}

```
This would make the behavior consistent and avoid surprises.

---
Let me know if you'd like a pull request with this change - I'd be happy to help!

powerd by KILL-9 💀

LeeHyungGeol:
Related Issues
Fixes #4916
Related to #4908, #4935

## Description
This PR fixes an issue where `RecordFieldExtractor` ignores the `names()` configuration when using Java Records with `FlatFileItemWriterBuilder`.

### Problem

When using `FlatFileItemWriterBuilder` with a Java Record type and specifying field names via `names()`, the `RecordFieldExtractor` was created but the `setNames()` method was not called, causing all record fields to be extracted instead of only the specified subset.

```java
// This configuration was being ignored
FlatFileItemWriter<UserRecord> writer = new FlatFileItemWriterBuilder<UserRecord>()
    .sourceType(UserRecord.class)
    .names("name", "age")  // <-- These names were not being applied
    .delimited()
    .build();
 ```
 
### Solution

Updated both DelimitedBuilder and FormattedBuilder in FlatFileItemWriterBuilder to properly call setNames() on the RecordFieldExtractor when names are specified:

```
if (this.sourceType != null && this.sourceType.isRecord()) {
    RecordFieldExtractor<T> recordFieldExtractor = new RecordFieldExtractor<>(this.sourceType);
    if (!this.names.isEmpty()) {
        recordFieldExtractor.setNames(this.names.toArray(new String[0]));
    }
    this.fieldExtractor = recordFieldExtractor;
}
```

### Testing
Added comprehensive tests covering:

- Selecting specific fields from a record
- Reordering fields in the output
- Both DelimitedBuilder and FormattedBuilder configurations
- Backward compatibility when all fields are specified

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/file/builder/FlatFileItemWriterBuilder.java
@@ -47,6 +47,7 @@ import org.springframework.util.StringUtils;
  * @author Mahmoud Ben Hassine
  * @author Drummond Dawson
  * @author Stefano Cordio
+ * @author Hyunggeol Lee
  * @since 4.0
  * @see FlatFileItemWriter
  */
@@ -394,7 +395,11 @@ public class FlatFileItemWriterBuilder<T> {
 
 			if (this.fieldExtractor == null) {
 				if (this.sourceType != null && this.sourceType.isRecord()) {
-					this.fieldExtractor = new RecordFieldExtractor<>(this.sourceType);
+					RecordFieldExtractor<T> recordFieldExtractor = new RecordFieldExtractor<>(this.sourceType);
+					if (!this.names.isEmpty()) {
+						recordFieldExtractor.setNames(this.names.toArray(new String[0]));
+					}
+					this.fieldExtractor = recordFieldExtractor;
 				}
 				else {
 					BeanWrapperFieldExtractor<T> beanWrapperFieldExtractor = new BeanWrapperFieldExtractor<>();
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 0eeacd583ffbb2d47dd6ed9bc76f914fd320b496. I just updated the test by changing the resources with mocks (there is no need to create temporary files in these unit tests as we do not really write data to them, but only assert on fields through reflection).

PR #5043, opened 2025-10-21
Improve javadoc of JobInstanceDao#getJobNames method

HJC96:
Fixed javadoc for `getJobNames()` to clarify it returns job names for which job instances exist.

Removed FIXME comment.


fmbenhassine:
Thank you for the PR! After re-reading that javadoc, I think I misunderstood it at first, but it is actually correct: the content of the job_instance table is actually about all the jobs that have ever been executed (ie that have been executed at least once). 

But the improvement you suggest is more explicit 👍

PR #5044, opened 2025-10-22
Fix potential NPE in JdbcJobExecutionDao

quaff:
The result set may be empty then unbox `lastJobExecutionId` will cause NPE.

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/repository/dao/AbstractJobDaoTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/repository/dao/AbstractJobDaoTests.java
@@ -254,6 +254,8 @@ public abstract class AbstractJobDaoTests {
 	@Transactional
 	@Test
 	void testGetLastJobExecution() {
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/test/java/org/springframework/batch/core/repository/dao/AbstractJobDaoTests.java
```diff
@@ -254,6 +254,8 @@ private void validateJobExecution(JobExecution lhs, JobExecution rhs) {
 	@Transactional
 	@Test
 	void testGetLastJobExecution() {
+		assertEquals(null, jobExecutionDao.getLastJobExecution(jobInstance));
```
quaff: @fmbenhassine It's necessary  to enable `JdbcJobDaoTests`.

antshaykolo: image

fmbenhassine:
LGTM 👍 Merged as ed383e2a965afdbc75b8d5ceb52ba0650da1400b. Thanks!

Issue #5049, opened 2025-10-24
JobParameter constructor validates wrong parameter (value instead of name)


KMGeon:
### Summary
The `JobParameter` record's compact constructor has a bug in parameter validation. It validates the `value` parameter twice instead of validating both `name` and `value` parameters.

---

## Current Behavior

```java
public record JobParameter<T>(String name, T value, Class<T> type, boolean identifying) implements Serializable {
    public JobParameter {
        Assert.notNull(value, "name must not be null");  // ❌ Bug: validates 'value' but message says 'name'
        Assert.notNull(value, "value must not be null"); // ❌ Bug: validates 'value' twice
        Assert.notNull(type, "type must not be null");
    }
}
```

---

## Expected Behavior

```java
public record JobParameter<T>(String name, T value, Class<T> type, boolean identifying) implements Serializable {
    public JobParameter {
        Assert.notNull(name, "name must not be null");   // ✅ Correct: validates 'name'
        Assert.notNull(value, "value must not be null"); // ✅ Correct: validates 'value'
        Assert.notNull(type, "type must not be null");
    }
}
```

---

### 1. Create Test Case

```java
@Test
void testNameParameterIsNull() {
    JobParameter<String> jobParameter = new JobParameter<>(null, "test", String.class, true);
    assertEquals("param", jobParameter.name());
    assertEquals("test", jobParameter.value());
    assertEquals(String.class, jobParameter.type());
    assertTrue(jobParameter.identifying());
}
```

### 2. Test Result

The test demonstrates that a `JobParameter` can be created with a `null` name, which should not be allowed:

```
[ERROR] Failures: 
[ERROR]   JobParameterTests.testNameParameterIsNull:37 expected: <param> but was: <null>
[INFO] 
[ERROR] Tests run: 7, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  5.049 s
[INFO] Finished at: 2025-10-25T01:43:01+09:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.3:test (default-test) on project spring-batch-core: There are test failures.
```

### 3. Analysis

The test fails because:
1. **No exception is thrown** when `name` is `null` during object construction (due to the validation bug)
[...]

PR #5050, opened 2025-10-24
Fix JobParameter validation bug in compact constructor

KMGeon:
## Summary
Fixed validation bug in `JobParameter` compact constructor that validated `value` parameter twice instead of validating both `name` and `value` parameters.

Fixes gh-5049

## Problem Description
The compact constructor in `JobParameter.java` had a typo:
```java
// Before (Bug)
Assert.notNull(value, "name must not be null");  // ❌ validates 'value' but message says 'name'
Assert.notNull(value, "value must not be null"); // ❌ validates 'value' twice
```

This allowed creating `JobParameter` objects with `null` name, which could cause:
- `NullPointerException` in `equals()` and `hashCode()` methods
- Contract violation (name should never be null)
- Misleading error messages

## Changes Made
**File**: `org.springframework.batch.core.job.parameters.JobParameter`
- Changed to validate `name` instead of `value`
```java
// After (Fixed)
Assert.notNull(name, "name must not be null");   // ✅ now validates 'name'
Assert.notNull(value, "value must not be null"); // ✅ validates 'value'
```

## Testing
- [x] Added new test case for validation
- [x] All existing tests pass
- [x] Code formatting applied (`spring-javaformat:apply`)
- [x] Full build completed successfully

### DIFF
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/JobParameterTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/JobParameterTests.java
@@ -17,11 +17,9 @@ package org.springframework.batch.core;
 
 import java.util.Date;
 
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.batch.core.job.parameters.JobParameter;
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Good catch! Thank you for your contribution!

PR #5054, opened 2025-10-28
Remove unused id from persistence domain object

quaff:
`getId()` always return null.


fmbenhassine:
Good catch! Rebased, squashed and merged as 175fa1f3a1168edcd2405407484f3b7f0b46bcbb. Thank you for your PR!

PR #5065, opened 2025-10-30
Use NVARCHAR for Spring Batch SQL Server metadata

### LINKED ISSUE #5064: SQL Server DDL for metadata tables should use NVARCHAR to prevent performance degradation and deadlocks from implicit conversion
Hello Spring Batch Team,

  I'm reporting a significant performance issue and a potential for deadlocks when using Spring Batch with a SQL Server database. The root cause is a data type mismatch between the default
   Spring Batch schema and the default behavior of the Microsoft JDBC driver.

  The Problem

  The default DDL for SQL Server metadata tables (e.g., BATCH_JOB_INSTANCE's JOB_NAME, JOB_KEY) defines these columns as VARCHAR. However, the Microsoft JDBC Driver for SQL Server sends
  string parameters as NVARCHAR by default.

  This mismatch forces SQL Server to perform an implicit data type conversion (CONVERT_IMPLICIT) on the VARCHAR column for every comparison.

  Impact: Performance Degradation and Deadlocks

  This implicit conversion prevents efficient index usage. The query optimizer, despite showing an Index Seek operation, actually performs a costly range scan due to the CONVERT_IMPLICIT
  in the WHERE clause and the use of GetRangeThroughConvert to define a search range. This behavior leads to substantial performance degradation that worsens linearly with data growth.

  This severe performance degradation causes transactions to hold locks for much longer. Under high concurrency, this dramatically increases lock contention and the risk of deadlocks. We
  have confirmed these deadlocks occur even at the READ COMMITTED isolation level, as the range scan acquires broader range locks, leading to mutual waiting during INSERT attempts.

  Proposed Solution

  To resolve this at its root, I propose updating the schema-sqlserver.sql file to change all relevant string columns in the metadata tables from `VARCHAR` to `NVARCHAR`. This aligns the
  database schema with the JDBC driver's default behavior, ensuring efficient index usage and preventing these performance and deadlock issues.

  Thank you for your consideration.

Chienlin1014:
Spring Batch SQL Server metadata tables default to VARCHAR, but the Microsoft JDBC driver sends strings as NVARCHAR. This mismatch causes implicit conversions, slows queries, and increases deadlock risk.

Update schema-sqlserver.sql to use NVARCHAR for all string columns, aligning with the driver and improving performance.

 fix #5064


Chienlin1014:
reference： [International features of the JDBC driver](https://learn.microsoft.com/en-us/sql/connect/jdbc/international-features-of-the-jdbc-driver)

Chienlin1014:
Fixes #5064

fmbenhassine:
@Chienlin1014 Thank you for reporting this issue and for contributing a PR! LGTM 👍 

Can you please add a migration script for v6 similar to https://github.com/spring-projects/spring-batch/blob/a33f0e9a328b12c4042a20d577fd0d4460e5734b/spring-batch-core/src/main/resources/org/springframework/batch/core/migration/5.0/migration-sqlserver.sql that alters the tables and changes the columns type accordingly? With that I think this PR could be merged in the upcoming v6. Thank you upfront.

Chienlin1014:
@fmbenhassine  Thank you for the quick review and the clear guidance!  
I've just added the migration script for v6.0 under `src/main/resources/org/springframework/batch/core/migration/6.0/migration-sqlserver.sql`, following the same structure and comments style as the v5.0 script.  
It includes the corresponding `ALTER TABLE` statements to update the affected columns to `NVARCHAR` types.  

Please let me know if you'd like me to adjust anything.  
Thank you again for the review and for considering this PR for inclusion in v6.0!

fmbenhassine:
Thank you for the updates! Rebased and merged, thank you for reporting the issue and for contributing a fix 🙏

PR #5071, opened 2025-11-01
Add support for multiple files via pattern in MultiResourceItemReaderBuilder

### LINKED ISSUE #5056: Support specifying resource locations using a pattern in MultiResourceItemReaderBuilder
`MultiResourceItemReaderBuilder` currently allows setting resources via the `resources(Resource[])` method:

```java
return new MultiResourceItemReaderBuilder<User>()
    .delegate(fileReader)
    .resources(new Resource[] { r1, r2, r3, r4 })
    .name("multiFileReader")
    .build();
```

In many cases, multiple files need to be specified using a pattern.
While this can be achieved manually using the PathMatchingResourcePatternResolver class:

```java
var resourcePatternResolver = new PathMatchingResourcePatternResolver();
Resource[] resources = resourcePatternResolver.getResources("file:/home/logs/*.csv");

return new MultiResourceItemReaderBuilder<User>()
    .delegate(fileReader)
    .resources(resources)
    .name("multiFileReader")
    .build();
```

It would be more convenient if MultiResourceItemReaderBuilder itself could directly accept a resource pattern.
In older XML-based configurations in Spring Framework , resource arrays defined with patterns were automatically resolved by `PathMatchingResourcePatternResolver.`
Given that background, it feels natural to provide a similar capability in the builder API.

I’d like to propose adding a method like MultiResourceItemReaderBuilder.resourcesPattern(String) to support this use case:

```java
return new MultiResourceItemReaderBuilder<User>()
    .delegate(fileReader)
    .filesPattern("/home/logs/*.csv")
    .name("multiFileReader")
    .build();
```

Since this would be a relatively simple enhancement, I’d be happy to submit a Pull Request myself if the team agrees with the need for it.

benelog:
It resolves #5056 

The method name intentionally uses `filesPattern` instead of `resourcesPattern`, since in most cases the target resources are file-based. Internally, `MultiResourceItemReaderBuilder.build()` resolves resources using `PathMatchingResourcePatternResolver.getResources(String) ` with a "file:" prefix, which aligns naturally with this naming choice. 

I also attempted to add a test case to cover the catch block in the following code segment:

```java
try {
	Resource[] resources = patternResolver.getResources("file:" + this.filesPattern);
	reader.setResources(resources);
} catch (IOException e) {
	throw new IllegalArgumentException(
			"Unable to initialize resources by the pattern " + this.filesPattern, e);
}
```

However, I could not find a case that would actually trigger this block.
When no files are found for the specified pattern, `PathMatchingResourcePatternResolver` simply returns an empty `Resource[]` rather than throwing an exception.

I considered making the builder throw an exception when the resolved `Resource[]` has zero length, but decided it would be more appropriate to delegate such behavior to `MultiResourceItemReader` or its internal delegate. For example, `FlatFileItemReader` determines whether to throw an exception for missing files based on its `strict` property.


fmbenhassine:
Thanks for the PR!

I'd like to take a slightly different approach to the API shape though. The builder should stay agnostic to the resource type. Hardcoding the `file:` prefix limits it to filesystem locations even though `PathMatchingResourcePatternResolver` can already resolve `classpath:`, `classpath*:`, etc. and adding a separate `filesPattern` field/method alongside `resources` gives us two mutually-exclusive ways to configure the same thing.

I am thinking about overloading `resources` instead, something like `resources(String... locationPatterns)`, and resolving eagerly into the existing resources field. That way the builder stays agnostic to the resource type (the caller picks the prefix), there's no new state/precedence to reason about, and it's consistent with how `Resource...` and `String...` naturally coexist as overloads.

I will prototype this and amend your PR if it works as expected.

fmbenhassine:
merged as 0d96bd0b790992736385bd6f0f8bd36949a6f494, refined in 8ce6762a1c6dea6e30c5a78357ebc74112c29d31.

benelog:
Thank you for merging this and for taking the time to reshape it. I had only filesystem input in mind when I hardcoded the `file:` prefix, and I still expect `file:` to be the common case, but leaving the prefix open makes for a better design: it lets the builder benefit from the full range of the `Resource` abstraction instead of narrowing it to one implementation. It is also good to see the same shape carried over to #5488.

Issue #5077, opened 2025-11-06
ChunkOrientedStepBuilder: Default SkipPolicy should be NeverSkipItemSkipPolicy when only retry is configured (not AlwaysSkipItemSkipPolicy)


KILL9-NO-MERCY:
Hi Spring Batch team,

I think I've found an unexpected behavior change in Spring Batch 6 regarding the default skip policy when only retry is configured.

**Bug description**

When configuring only retry settings without any skip configuration, the default `SkipPolicy` is set to `AlwaysSkipItemSkipPolicy`. This causes all items that fail after exhausting retry attempts to be silently skipped instead of failing the step, which seems unintended.

**Environment**

- Spring Batch version: 6.0.0-RC2

**Steps to reproduce**
1. Configure a chunk-oriented step with retry but WITHOUT skip configuration:

2. Throw an exception from processor that exceeds retry limit

4. Observe that the item is skipped instead of failing the step

**Expected behavior**
When only retry is configured without any skip settings, items that fail after exhausting all retry attempts should **fail the step**, not be skipped.

The default `SkipPolicy` should be `NeverSkipItemSkipPolicy` (or equivalent) when skip is not explicitly configured.

**Root cause **

In `ChunkOrientedStepBuilder`:
```java
if (this.skipPolicy == null) {
    if (!this.skippableExceptions.isEmpty() || this.skipLimit > 0) {
        this.skipPolicy = new LimitCheckingExceptionHierarchySkipPolicy(this.skippableExceptions, this.skipLimit);
    }
    else {
        this.skipPolicy = new AlwaysSkipItemSkipPolicy(); // ← This seems wrong
    }
}
```

When neither `skippableExceptions` nor `skipLimit` is configured, it defaults to `AlwaysSkipItemSkipPolicy`, causing unexpected skip behavior.

**Comparison with Spring Batch 5**

In Spring Batch 5's `FaultTolerantStepBuilder`:
```java
if (skipPolicy == null) { // default == null
    if (skippableExceptionClasses.isEmpty() && skipLimit > 0) {
        logger.debug(String.format(
            "A skip limit of %s is set but no skippable exceptions are defined.",
            skipLimit));
    }
    skipPolicy = limitCheckingItemSkipPolicy; 
}
```

This would result in step failure when retry is exhausted without skip configuration.

**Suggested fix**
Change the default `SkipPolicy` to `NeverSkipItemSkipPolicy` when skip is not configured:
```java
[...]

PR #5080, opened 2025-11-08
Fix default skip policy when only retry is configured

JunggiKim:
Fix default skip policy when only retry is configured.

When configuring retry without skip in a fault-tolerant step, items that 
fail after exhausting retry attempts are currently silently skipped instead 
of failing the step, which can lead to silent data loss.

This PR changes the default SkipPolicy from AlwaysSkipItemSkipPolicy to 
NeverSkipItemSkipPolicy when no skip exceptions are explicitly configured.

**Changes:**
- Removed `skipLimit > 0` condition from skip policy initialization
- Updated default SkipPolicy from AlwaysSkipItemSkipPolicy to NeverSkipItemSkipPolicy
- Updated JavaDoc to reflect the new default behavior
- Added unit test to verify default SkipPolicy behavior

Resolves gh-5077


fmbenhassine:
Thank you for your PR! Apologies I fixed that in ce7e03acf9766983019be34e3b2a633756b5669f and forgot about your PR 😔 BTW, this PR updates the skip policy in `ChunkOrientedStepBuilder` but not in `ChunkOrientedStep` class itself. Anyway thank you for your time and effort! I see you have opened other PR which I am going to check.

Issue #5079, opened 2025-11-07
ChunkOrientedStep does not throw exception when skipPolicy.shouldSkip() returns false


KILL9-NO-MERCY:
Hi Spring Batch team,

I think, I’ve discovered a bug in `ChunkOrientedStep` where failed items are silently discarded when the skip policy rejects skipping.

## Bug description

When retry is exhausted in fault-tolerant mode, `ChunkOrientedStep` calls `skipPolicy.shouldSkip()` to determine whether the failed item should be skipped. However, when `skipPolicy.shouldSkip()` returns `false` (meaning the item should NOT be skipped), the code does not throw an exception. This causes the failed item to be silently lost, and the job continues as if nothing happened.

This affects three methods in `ChunkOrientedStep`:
- `doSkipInRead()` (line 528)
- `doSkipInProcess()` (line 656)
- `scan()` (line 736)

## Environment

- Spring Batch version: 6.0.0-RC2

## Steps to reproduce
1. Configure a fault-tolerant step with a skip policy that always returns `false` (never skip)
2. Configure retry with a limited number of attempts (e.g., retryLimit = 2)
3. Process items where one item consistently fails
4. After retry exhaustion, observe that the failed item is silently discarded instead of causing the job to fail

## Expected behavior

When `skipPolicy.shouldSkip()` returns `false`, the exception should be re-thrown to:
- Roll back the transaction
- Mark the step as FAILED
- Prevent silent data loss

The job should fail with a clear error indicating that the skip limit was exceeded or skip policy rejected skipping.

## Minimal Complete Reproducible example
```java

@Slf4j
@Configuration
public class IssueReproductionJobConfiguration {
    @Bean
    public Job issueReproductionJob(JobRepository jobRepository, Step issueReproductionStep) {
        return new JobBuilder(jobRepository)
                .start(issueReproductionStep)
                .build();
    }

    @Bean
    public Step issueReproductionStep(JobRepository jobRepository) {
        return new StepBuilder(jobRepository)
                .<String, String>chunk(3)
                .reader(issueReproductionReader())
                .processor(issueReproductionProcessor())
                .writer(issueReproductionWriter())
                .faultTolerant()
                .retryLimit(2)
                .skipPolicy(new NeverSkipItemSkipPolicy())  
                .build();
    }

    @Bean
    public ItemReader<String> issueReproductionReader() {
        return new ListItemReader<>(List.of("Item_1", "Item_2", "Item_3"));
    }

    @Bean
[...]

Issue #4868, opened 2025-06-09
Remove dependency to Spring Retry


fmbenhassine:
After the introduction of a core retry feature in Spring Framework (see https://github.com/spring-projects/spring-framework/pull/34716), Spring Batch should be updated to use that functionality n favour of Spring Retry.

HyunsangHan:
Here is what I understand about this task:
- Instead of using the `RetryTemplate` from [**spring-retry**](https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/support/RetryTemplate.java), we can replace it to the `RetryTemplate` provided by [**spring-framework (version 7.0 or higher)**](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/retry/RetryTemplate.java), and then remove the dependency on spring-retry altogether.
- Within spring-batch, retry logic is primarily handled through BatchRetryTemplate.

fmbenhassine:
Hi @HyunSangHan ,

Thank you for looking into this! However, this is a task that the team will take care of as it will require a substantial amount of work and API changes (related to fault-tolerance as described in #3950). I forgot to tag it accordingly when I created it..

Thanks again anyway!

HyunsangHan:
@fmbenhassine 
Appreciate the context. No worries! :)
Please feel free to reach out if you need any help.

fmbenhassine:
The retry functionality was moved to Spring Framework's Core Retry in #3950. However, the `org.springframework.classify.*` APIs from `spring-retry` are still used in the framework and it is not possible to make `spring-retry` an optional dependency.

Since the legacy implementation of the chunk-oriented processing model will be kept for backward compatibility during the entire lifetime of v6, this issue will be moved to v7.

FTR: At that time, `org.springframework.classify.*` should be copied in Spring Batch's code base and the dependency to `spring-retry` can be safely removed.

PR #5081, opened 2025-11-08
Fix silent data loss in ChunkOrientedStep when skip policy rejects

JunggiKim:
This change adds proper exception handling in ChunkOrientedStep to prevent silent data loss when skip policy rejects skipping after retry exhaustion.

**Changes:**
- Added else blocks in doSkipInRead(), doSkipInProcess(), and scan() methods
- Throw appropriate NonSkippable*Exception when skipPolicy.shouldSkip() returns false
- Added test case verifying NonSkippableProcessException is thrown

**Problem:**
When retry is exhausted and skipPolicy.shouldSkip() returns false, the current implementation silently discards items without throwing an exception. This causes silent data loss.

**Solution:**
Added else blocks to throw NonSkippableReadException, NonSkippableProcessException, or NonSkippableWriteException when skip policy rejects skipping.

Fixes gh-5079


fmbenhassine:
LGTM 👍 Rebased and merged as 946f78825414b872f3d27110ff53347a86d362e5. I just updated the test to use the `core.retry` feature from Spring Framework and not `SimpleRetryPolicy` from Spring Retry (see #4868)

PR #5095, opened 2025-11-20
Use fine-grained lock for JDBC DAO implementations

quaff:


quaff:
Inspired by https://www.baeldung.com/java-acquire-lock-by-key

PR #5100, opened 2025-11-21
Resolves minor logging issue when a step or job completes instantly

### LINKED ISSUE #5037: Minor logging issue when a step or job completes instantly
**Bug description**
We found such below log lines in our production environment, which misses the duration information at the end of the line.

... `Job: [FlowJob: [name=...]] completed with the following parameters: [...] and the following status: [FAILED] in `

Digging into a little bit of the code, I found that when a job is finished
https://github.com/spring-projects/spring-batch/blob/11ec7f12e8e4477ae802a02ee72f69f78afbf25b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/TaskExecutorJobLauncher.java#L221-L228

an info level log entry gets emitted by the framework. However if the start and the end date is essentially the same, the `BatchMetrics.formatDuration` method returns an empty `String` because of the `duration.isZero()` condition.

https://github.com/spring-projects/spring-batch/blob/11ec7f12e8e4477ae802a02ee72f69f78afbf25b/spring-batch-core/src/main/java/org/springframework/batch/core/observability/BatchMetrics.java#L69-L72

**Environment**
Java21 (temurin)
Spring Batch 5.2.2
File based H2 DB is used

**Steps to reproduce**
Start a lot of batch jobs which completes fast.

**Expected behavior**
Honestly I don't know, but at least a `0ms` would be better than nothing. Something like this:
`Job: [FlowJob: [name=...]] completed with the following parameters: [...] and the following status: [FAILED] in 0ms`

The below test fails for 5.2.2:
```
    @Test
    void testFormatDurationWhenCalculationReturnsZeroDuration() {
        var startDate = LocalDateTime.now();
        // create end date from the string representation of start date to ensure both dates are equal, but different references.
        // In reality there is another LocalDateTime.now() call, but that could return a different time, which could cause flaky tests.
        var endDate = LocalDateTime.parse(startDate.toString());
        var calculateDuration = BatchMetrics.calculateDuration(startDate, endDate);
        Assertions.assertNotNull(calculateDuration, "Calculated duration is a null reference!");
        var formattedDurationString = BatchMetrics.formatDuration(calculateDuration);
        Assertions.assertTrue(StringUtils.hasText(formattedDurationString), formattedDurationString);
    }
```

janossch:
Fixes #5037

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/observability/BatchMetrics.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/observability/BatchMetrics.java
@@ -66,8 +66,11 @@ public final class BatchMetrics {
 	 * @return A human readable duration
 	 */
 	public static String formatDuration(@Nullable Duration duration) {
-		if (duration == null || duration.isZero() || duration.isNegative()) {
+		if (duration == null || duration.isNegative()) {
 			return "";
+		}
+		if (duration.isZero()) {
+			return "0ms";
 		}
 		StringBuilder formattedDuration = new StringBuilder();
 		long hours = duration.toHours();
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 1d50d829907a580fe3aea5b6a17859a418e478b9. Thank you for your contribution!

PR #5102, opened 2025-11-24
Refactor the JpaPagingItemReader by making the entityManagerFactory f…

### LINKED ISSUE #5101: Refactor JpaPagingItemReader to be compliant with the requirements of constructor
As the new signature of the JpaPagingItemReader is the following: 

```java 
/**
 * Create a new {@link JpaPagingItemReader} instance.
 * @param entityManagerFactory the JPA entity manager factory.
 * @since 6.0
*/
public JpaPagingItemReader(EntityManagerFactory entityManagerFactory) {
	Assert.notNull(entityManagerFactory, "EntityManagerFactory must not be null.");
	this.entityManagerFactory = entityManagerFactory;
}
```

The same apply to the JpaItemWriter: 
```java
/**
 * Create a new {@link JpaItemWriter} instance.
 * @param entityManagerFactory the entity manager factory to use
 * @since 6.0
*/
public JpaItemWriter(EntityManagerFactory entityManagerFactory) {
	Assert.notNull(entityManagerFactory, "EntityManagerFactory must not be null");
	this.entityManagerFactory = entityManagerFactory;
}
```

We should make the field `EntityManagerFactory entitiyManagerFactory` final as we expect to get a non-null factory and remove the `setter` associated to this field.

Also the reader/writer should not change of entityManagerFactory during the process, and we should define a new one if we need to change the source entityManagerFactory.

maequise:
Fixes #5101

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JpaItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JpaItemWriter.java
@@ -49,7 +49,7 @@ public class JpaItemWriter<T> implements ItemWriter<T> {
 
 	protected static final Log logger = LogFactory.getLog(JpaItemWriter.class);
 
-	private EntityManagerFactory entityManagerFactory;
+	private final EntityManagerFactory entityManagerFactory;
 
 	private boolean usePersist = false;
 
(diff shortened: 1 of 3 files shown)


fmbenhassine:
Thank you for reporting #5101  and for opening this PR. We do not remove public APIs directly, we typically deprecate them first. See example in 77250aac181212bb98e62507007480604f14fba6.

I will amend that when merging this PR.

fmbenhassine:
Amended and merged as 4d70fe197db33af71935eed1107bdd8c9af5ebe7. I also included the same change in the JPA cursor reader as well.

maequise:
Thank you for the merge. And I take note for a next contribution.

PR #5113, opened 2025-11-27
Replace CopyOnWriteArrayList with ArrayList in PagingItemReaders

### LINKED ISSUE #5112: Replace CopyOnWriteArrayList with ArrayList in JdbcPagingItemReader and JpaPagingItemReader for performance and correctness
**Bug description**

JdbcPagingItemReader and JpaPagingItemReader create the internal results list as a CopyOnWriteArrayList:

```java
if (results == null) {
    results = new CopyOnWriteArrayList<>();
} else {
    results.clear();
}
```

However, paging readers are strictly single-threaded.
AbstractPagingItemReader#doRead() already serializes access with an internal lock:

```java
this.lock.lock();
try {
    ...
} finally {
    this.lock.unlock();
}
```

Because there is no concurrent access to results, using CopyOnWriteArrayList provides no benefit and introduces unnecessary write-cost overhead. A regular ArrayList is sufficient and more appropriate for this use case.

This is not a functional bug, but the current choice of collection does not match the actual access pattern of the readers.

**Environment**

- Spring Batch version: 6.0.0
- Java version: Java 21

**Expected behavior**
Since paging readers operate in a single-threaded, sequential manner, the internal list should use ArrayList instead of CopyOnWriteArrayList.

banseok1216:
Fixes #5112

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JdbcPagingItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JdbcPagingItemReader.java
@@ -25,7 +25,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
-import java.util.concurrent.CopyOnWriteArrayList;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Merged as 3f563357a97e06a9f99dd4799b6c2ecae8e4670d.

PR #5119, opened 2025-11-30
Remove redundant assert in StoredProcedureItemReader

banseok1216:
PR Description

This PR removes a redundant assertion in StoredProcedureItemReader.

The removed line was:
```java
Assert.state(refCursorPosition == 0 || refCursorPosition > 0, ...);
```
does not provide any real validation.The condition is always true for any non-negative value, and the preceding refCursorPosition >= 0 check already covers the only meaningful case.
The message also hints at parameter-bound validation that never happens.

This assertion was introduced in:
[8f70ed8](https://github.com/spring-projects/spring-batch/commit/8f70ed8a31dc108204baabbc92026c5a0ee36f99)


fmbenhassine:

PR #5131, opened 2025-12-05
Improve retrieval of running job executions

quaff:
This commit could save one unnecessary query.

See https://github.com/spring-projects/spring-batch/commit/57504927d912947ad1d15079b00d0969060db664


fmbenhassine:
Amazing! Thank you for the PR!

PR #5133, opened 2025-12-05
Improve `JdbcStepExecutionDao::getLastStepExecution` to reuse `JdbcJobExecutionDao::getJobExecution`

quaff:
This commit eliminate duplicate `StepExecution` mapping code from `JdbcStepExecutionDao`, make the code cleaner and keep the consistent style.

The drawback is it requires one more query, but it's acceptable since the query is by primary key.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcStepExecutionDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcStepExecutionDao.java
@@ -28,8 +28,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import org.springframework.batch.core.BatchStatus;
-import org.springframework.batch.core.ExitStatus;
 import org.springframework.batch.core.job.JobExecution;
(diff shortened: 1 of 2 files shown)


sean-huni:
Drive-by note in case this PR is ever considered a fix for portability: it changes which call is nested, but the call is still made from inside the open `ResultSet`.

```java
+   return new StepExecutionRowMapper(
+           jobExecutionDao.getJobExecution(rs.getLong("JOB_EXECUTION_ID")))
+       .mapRow(rs, 0);
```

That matters on databases which refuse more than one simultaneously-open portal. CockroachDB rejects the current code outright, so every Spring Batch **restart** fails there, and this refactor would leave that unchanged. Since `getJobExecution` issues more statements than the `getJobParameters` call it replaces, it would arguably nest more work rather than less.

None of which is an argument against the refactor itself. The duplicate mapping code is worth removing, and the extra primary-key query you call out as the drawback is a fair trade.

It just needs one more step to also close the portability gap: read the row (or only its id) first, close the `ResultSet`, then resolve the `JobExecution`. `getStepExecution(long)` already does exactly that sequentially and could be delegated to, which would remove the duplication you are targeting here as well.

Filed the underlying defect with a reproduction and the pgjdbc mechanism (`setMaxRows(1)` leaves the portal suspended until the next `Sync`) as #5470, and referenced this PR there.

PR #5140, opened 2025-12-07
Refine ResourcelessJobRepository lookups and add delete operations

### LINKED ISSUE #5139: Enhance `ResourcelessJobRepository` implementation for testing
## Background

I understand the design intent behind `ResourcelessJobRepository`, based on the following issue:

* [https://github.com/spring-projects/spring-batch/issues/4679](https://github.com/spring-projects/spring-batch/issues/4679)

 However, I believe there are a few targeted enhancements that would make this class much more useful in tests, without violating the original design goals.

In a single application context, `ResourcelessJobRepository` cannot run the same job multiple times. This limitation is acceptable as long as users understand it, but in tests it can become a constraint.

For example, the following test runs the same job with different `JobParameters` using `JobOperatorTestUtils`, but cannot rely on `ResourcelessJobRepository`:

```java
@SpringBootTest("spring.batch.job.enabled=false")
@SpringBatchTest
class HelloParamJobTest {

  @Autowired
  JobOperatorTestUtils testUtils;

  @BeforeEach
  void prepareTestUtils(@Autowired @Qualifier("helloParamJob") Job helloParamJob) {
    testUtils.setJob(helloParamJob);
  }

  @Test
  void startJob() throws Exception {
    JobParameters params = testUtils.getUniqueJobParametersBuilder()
        .addLocalDate("helloDate", LocalDate.of(2025, 7, 28))
        .toJobParameters();
    JobExecution execution = testUtils.startJob(params);
    assertThat(execution.getExitStatus()).isEqualTo(ExitStatus.COMPLETED);
  }

  @Test
  void startJobWithInvalidJobParameters() {
    JobParameters params = testUtils.getUniqueJobParametersBuilder()
        .addLocalDate("goodDate", LocalDate.of(2025, 7, 28))
        .toJobParameters();
    assertThatExceptionOfType(InvalidJobParametersException.class)
        .isThrownBy(() -> testUtils.startJob(params))
        .withMessageContaining("do not contain required keys: [helloDate]");
  }
}
```

## Known solutions

I am aware of several existing workarounds for this limitation:

* Register `ResourcelessJobRepository` as a prototype-scoped bean.
* Use an in-memory database together with a JDBC-based `JobRepository`.
* Use `@DirtiesContext` (or similar) to refresh the `ApplicationContext` before each test.

However, all of these come with trade-offs, such as:

* Additional complexity in object wiring and dependencies.
* Extra configuration overhead.
* Slower test execution due to frequent context refreshes.

In particular, as mentioned in this comment:

* [https://github.com/spring-projects/spring-batch/issues/5118#issuecomment-3604092261](https://github.com/spring-projects/spring-batch/issues/5118#issuecomment-3604092261)

calls to `ResourcelessJobRepository#getJobInstance(String, JobParameters)` have caused test scenarios that worked well with  Spring Batch v5.2 to become impossible when upgrading to v6.0.

In such cases, users may see an error like:

```text
Message: A job instance already exists and is complete for identifying parameters={JobParameter{name='batch.random', value=4546055881725385948}
```

This can be confusing because the job name and/or `JobParameters` are actually different, yet the repository still resolves them to the same `JobInstance`. This also feels misaligned with Spring Batch's conceptual model, where a `JobInstance` is uniquely identified by a job name and its `JobParameters`.

## Proposed enhancements

To address these issues while preserving the original design, I would like to propose the following enhancements to `ResourcelessJobRepository`:

### 1. Filter return values based on job name, IDs, and parameters

For the methods below, compare the incoming arguments (`jobName`, `instanceId`, `executionId`, `JobParameters`, etc.) with the values held in the internal `JobInstance` and `JobExecution` fields, and filter return values accordingly:

* `getJobInstances(String jobName, int start, int count)`
* `findJobInstances(String jobName)`
* `getJobInstance(long instanceId)`
* `getLastJobInstance(String jobName)`
* `getJobInstance(String jobName, JobParameters jobParameters)`
* `getJobInstanceCount(String jobName)`
* `getJobExecution(long executionId)`
* `getLastJobExecution(String jobName, JobParameters jobParameters)`
* `getLastJobExecution(JobInstance jobInstance)`
* `getJobExecutions(JobInstance jobInstance)`

Some methods already have comments like `// FIXME should return null if the id is not matching`, which suggest that this kind of filtering was already considered. Even if only a subset of these methods were updated, the observable behavior might be acceptable in practice. However, for conceptual consistency and to future-proof the implementation, I believe it would be better to have a systematic comparison of `jobName`, `jobInstanceId`, etc., across the class.

### 2. Add methods to delete the current JobInstance and JobExecution

Add the ability to drop the currently held `JobInstance` and `JobExecution` from `ResourcelessJobRepository`:

* `deleteJobInstance(JobInstance jobInstance)`
* `deleteJobExecution(JobExecution jobExecution)`

If these methods are introduced, tests could intentionally delete the just-run `JobInstance` or `JobExecution` to reuse the same `ResourcelessJobRepository` instance in a more flexible way. For example:

```java
@Autowired
JobOperatorTestUtils testUtils;

@Autowired
JobRepository repository;

@BeforeEach
void prepareTestUtils(@Autowired @Qualifier("helloParamJob") Job helloParamJob) {
  testUtils.setJob(helloParamJob);
}

@Test
void startJob() throws Exception {
  JobParameters params = testUtils.getUniqueJobParametersBuilder()
      .addLocalDate("helloDate", LocalDate.of(2025, 7, 28))
      .toJobParameters();
  JobExecution execution = testUtils.startJob(params);
  assertThat(execution.getExitStatus()).isEqualTo(ExitStatus.COMPLETED);

  // Explicitly clear the current JobInstance for the next test
  repository.deleteJobInstance(execution.getJobInstance());
}
```

These two enhancements together would:

* Make the implementation more faithful to the JobRepository interface contract.
  * Reduce surprising behavior where different jobs/parameters map to the same `JobInstance`.
* Make `ResourcelessJobRepository` more useful in test scenarios, especially when upgrading from Spring Batch 5.x to 6.x.
* Maintain the original in-memory, single-instance nature of `ResourcelessJobRepository`.

I would be happy to open a PR or further refine this proposal based on feedback from the maintainers.

benelog:
It resolves #5139

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/ResourcelessJobRepository.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/ResourcelessJobRepository.java
@@ -22,8 +22,10 @@ import java.util.List;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.batch.core.BatchStatus;
+import org.springframework.batch.core.job.DefaultJobKeyGenerator;
 import org.springframework.batch.core.job.JobExecution;
 import org.springframework.batch.core.job.JobInstance;
+import org.springframework.batch.core.job.JobKeyGenerator;
 import org.springframework.batch.core.job.parameters.JobParameters;
 import org.springframework.batch.infrastructure.item.ExecutionContext;
 import org.springframework.batch.core.step.StepExecution;
@@ -45,6 +47,7 @@ import org.springframework.batch.infrastructure.support.transaction.Resourceless
  *
  * @since 5.2.0
  * @author Mahmoud Ben Hassine
+ * @author Sanghyuk Jung
  */
 public class ResourcelessJobRepository implements JobRepository {
 
@@ -53,6 +56,8 @@ public class ResourcelessJobRepository implements JobRepository {
 	private @Nullable JobExecution jobExecution;
 
 	private long stepExecutionIdIncrementer = 0L;
+
+	private JobKeyGenerator jobKeyGenerator = new DefaultJobKeyGenerator();
 
 	/*
 	 * ===================================================================================
@@ -76,7 +81,7 @@ public class ResourcelessJobRepository implements JobRepository {
 
 	@Override
 	public List<JobInstance> getJobInstances(String jobName, int start, int count) {
-		if (this.jobInstance == null) {
+		if (this.jobInstance == null || !this.jobInstance.getJobName().equals(jobName)) {
 			return Collections.emptyList();
 		}
 		return Collections.singletonList(this.jobInstance);
@@ -91,7 +96,7 @@ public class ResourcelessJobRepository implements JobRepository {
 	 */
 	@Override
 	public List<JobInstance> findJobInstances(String jobName) {
-		if (this.jobInstance == null) {
+		if (this.jobInstance == null || !this.jobInstance.getJobName().equals(jobName)) {
 			return Collections.emptyList();
 		}
 		return Collections.singletonList(this.jobInstance);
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Rebased and merged as 7eaaa029e24a570802c87dd03aee77588cd4d899. I just amended a constructor with the key generator in case one wants to use a custom key generator Thank you for your contribution!

Issue #5138, opened 2025-12-05
Step execution no longer persisted after partitioner creates the context


brian-mcnamara:
**Bug description**
With [this change](https://github.com/spring-projects/spring-batch/commit/90d895955d951156849ba6fa018676273fdbe2c4#diff-1ccc98868257080253b51baded74a755478f3f85f754e0dc8ef05144ecd7dc02), a steps context is no longer persisted, causing the execution context to be lost in a partitioned remote worker execution environment. Specifically the call to jobRepository.addAll would persist the context.

Specifically, after the contexts are created, MessageChannelPartitionHandler.doHandle is responsible to create and send the message, when the remote worker receives the message, it loads the steps through the job repository.

**Environment**
Spring boot 4.0.0, batch 6.0.0, batch-integration 6.0.0, JDK21

**Steps to reproduce**

See https://github.com/brian-mcnamara/SpringBatch6/blob/main/src/main/java/com/example/batchpartitionbug/BatchPartitionBugApplication.java

(can be run with `./gradlew run`) Specifically note line 91 which should get the context from the partitioner on line 197

This bug is seen with a partitioned step using message channels for delivery and a persistence layer for the jobRepository. 

**Expected behavior**

The partitioner should initialize the step contexts on the controller and update the step in the repository. This enabling the worker to load the step from the repository and use the context created earlier

quaff:
My application failed to start due to `@Value("#{stepExecutionContext['xxx']}")` is null after upgrade to Spring Boot 4.0.

fmbenhassine:
Thank you for reporting this issue and for providing an example! This is a valid issue. It seems like we have a gap in our remote partitioning [tests](https://github.com/spring-projects/spring-batch/tree/main/spring-batch-samples/src/test/java/org/springframework/batch/samples/partition/remote).. I will check that. I planned the fix for the upcoming 6.0.1 release.

brian-mcnamara:
Thank you both for the quick turn around, and all your work!

PR #5147, opened 2025-12-08
Fix step execution context is not persisted and restored

quaff:
1. Step execution context is not persisted in `SimpleStepExecutionSplitter::split`
2. Step execution context is not restored in `SimpleJobRepository::getStepExecution`

Closes GH-5138


quaff:
@fmbenhassine Tests included, please review.

fmbenhassine:
LGTM 👍 Thank you very much for your contribution!

PR #5148, opened 2025-12-08
Fix typo in testing.adoc startJob method name

arey:
Corrected a typo in the method name from `srartJob()` to `startJob()`.


fmbenhassine:
Good catch! Every time I promise myself to NOT type code by hand and still do it again.. At least it proves that I don't use AI 😂 (or should I?)

Thank you for the PR!

PR #5153, opened 2025-12-09
Default event-based item readers to bean name and remove strict name assertions

### LINKED ISSUE #5149: Default reader/writer name to the bean name
now that Spring Batch is not tied to any persistence mechanism and now that the default is actually resourceless, could we provide some sensible defaults for the names? 

for example, StepBuilder and JobBuilder can both be used without specifying a name (I love this!) but some of the builders like FlatFileItemBuilder can not. 

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.batch.infrastructure.item.file.FlatFileItemReader]: Factory method 'reader' threw exception with message: A name is required when saveState is set to true.
```

if we dropped the assertion in the builder and then had the produced FlatFileItemReader implement BeanNameAware so that if the name is not explicitly provided we can default it to the bean name, it would be nice. people could provide their own names, but get a sensible default otherwise.

therepanic:
Added `BeanNameAware` to event item readers to use the bean name when no explicit  name is provided. Removed name assertions from the corresponding builders.

I've added a `BeanNameAware` implementation wherever the builders have a name. I believe this is what was meant in #5149. In any case, we can look into it.

Closes: #5149

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/data/MongoCursorItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/data/MongoCursorItemReader.java
@@ -28,6 +28,7 @@ import org.bson.codecs.DecoderContext;
 import org.jspecify.annotations.Nullable;
 import org.springframework.batch.infrastructure.item.ItemReader;
 import org.springframework.batch.infrastructure.item.support.AbstractItemCountingItemStreamItemReader;
+import org.springframework.beans.factory.BeanNameAware;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.mongodb.core.MongoOperations;
@@ -45,9 +46,11 @@ import org.springframework.util.StringUtils;
  * @author LEE Juchan
  * @author Mahmoud Ben Hassine
  * @author Jimmy Praet
+ * @author Andrey Litvitski
  * @since 5.1
  */
-public class MongoCursorItemReader<T> extends AbstractItemCountingItemStreamItemReader<T> implements InitializingBean {
+public class MongoCursorItemReader<T> extends AbstractItemCountingItemStreamItemReader<T>
+		implements InitializingBean, BeanNameAware {
 
 	private MongoOperations template;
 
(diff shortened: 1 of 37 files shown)


therepanic:
Can you tell me roughly why this problem occurs?
```
Error: Failures:
Error: VirtualThreadsSupportFunctionalTests.testParallelStepsWithVirtualThreads:128 expected: <exitCode=COMPLETED;exitDescription=> but was: <exitCode=FAILED;exitDescription=>
```

quaff:
> Can you tell me roughly why this problem occurs?
> 
> ```
> Error: Failures:
> Error: VirtualThreadsSupportFunctionalTests.testParallelStepsWithVirtualThreads:128 expected: <exitCode=COMPLETED;exitDescription=> but was: <exitCode=FAILED;exitDescription=>
> ```

It's a flaky test.

banseok1216:
I don’t think need to add BeanNameAware explicitly to the event-based item readers.

They already get BeanNameAware via the common hierarchy:

Object
└─ ItemStreamSupport implements ItemStream, BeanNameAware
└─ AbstractItemStreamItemReader
└─ AbstractItemCountingItemStreamItemReader
└─ MongoCursorItemReader / other readers...

ItemStreamSupport already sets a default name based on the short class name and only overrides it with the bean name when no explicit name has been set. So removing the strict name assertions in the builders makes sense, but adding BeanNameAware again on the concrete readers looks redundant unless there is a reader that does not extend ItemStreamSupport.

On a related note, I’m a bit unsure about removing the old assertion in the builders:

```java
if (this.saveState) {
    Assert.hasText(this.name, "A name is required when saveState is set to true");
}
```
When saveState is true, the reader is effectively promising to participate in restart by storing its position in the ExecutionContext under a name-based key. With the assertion gone, misconfiguration around names won’t fail anymore, but may instead surface later as subtle restart issues. It might be worth adding a test around this to ensure the implicit naming behaviour is still safe for restarts.

fmbenhassine:
Thanks for looking into this! As mentioned by @banseok1216 and explained in https://github.com/spring-projects/spring-batch/issues/5149#issuecomment-5702885562, there is no need to add `BeanNameAware` to these classes, since `ItemStreamSupport` (the common ancestor for all of them) already implements it and already defaults the name to the bean name when one isn't explicitly set (since 6.0). 

The issue comes from the builders enforcing a stricter, redundant assertion on top of that existing mechanism. A couple of other things I noticed:
- The fix needs to cover the writer side too (`FlatFileItemWriterBuilder`, `MultiResourceItemWriterBuilder`, `JsonFileItemWriterBuilder`), which have the same pattern but aren't touched here.
- `RepositoryItemReader.afterPropertiesSet()` has its own separate assertion that isn't addressed by the builder-side removal, so it would still fail at runtime for that reader.

The CI issue is not related to this PR.

Given the scope of getting this right across all readers and writers, I'll take care of it myself. Thanks again for raising it!

PR #5159, opened 2025-12-11
Introduce AbstractJobRepositoryFactoryBean for both JDBC and MongoDB repository

quaff:


fmbenhassine:
Fantastic! Thank you for updating these tests.

Issue #5161, opened 2025-12-11
OptimisticLockingFailureException in JobRepositoryTestUtils.removeJobExecutions() since Spring Batch 5.2.3


szopal24:
Since Spring Batch 5.2.3, calling JobRepositoryTestUtils.removeJobExecutions() in test cleanup methods throws OptimisticLockingFailureException when trying to delete job executions.

Environment
Spring Batch: 5.2.3, 5.2.4
Spring Boot: 3.4.5, 3.5.8
Java: 17
Database: PostgreSQL (with table prefix BOOT3_BATCH_)

**Test Code Example:**
```

    @Test
    public void testJob() throws Exception {
        JobExecution jobExecution = jobLauncherTestUtils.launchJob(defaultJobParameters());
        jobExecutionList.add(jobExecution);
        
        assertThat(jobExecution.getExitStatus()).isEqualTo(ExitStatus.COMPLETED);
    }

    @After
    public void cleanUp() {
        // This throws OptimisticLockingFailureException since 5.2.3
        jobRepositoryTestUtils.removeJobExecutions(jobExecutionList);
    }
```

Result:

```
org.springframework.dao.OptimisticLockingFailureException: Attempt to delete step execution id=95106 with wrong version (1)

	at org.springframework.batch.core.repository.dao.JdbcStepExecutionDao.deleteStepExecution(JdbcStepExecutionDao.java:386)
	at org.springframework.batch.core.repository.support.SimpleJobRepository.deleteStepExecution(SimpleJobRepository.java:316)
	at org.springframework.batch.core.repository.support.SimpleJobRepository.deleteJobExecution(SimpleJobRepository.java:324)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:360)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:196)
[...]

PR #5173, opened 2025-12-18
Improve JobRepositoryTestUtils to mitigate OptimisticLockingFailureException while removing JobExecution

quaff:
It's safe to remove stale version of JobExecution for testing purpose, this commit mitigate OptimisticLockingFailureException but cannot avoid since race condition may happen.

Fix GH-5161


lszopa:
In which version will this change be available?

quaff:
> In which version will this change be available?

Let's wait for response from project leader @fmbenhassine .

fmbenhassine:
Thank you for the PR, @quaff !

I will merge this in 6.0.2 and back port it to 5.2.5

fmbenhassine:
LGTM 👍  Rebased and merged as 1d9536cfbabf54b921f5a522beef62e5d0793a97.

PR #5175, opened 2025-12-18
Use Spring Framework's CollectionUtils to create HashSet/HashMap

quaff:
It handles initial capacity correctly.


quaff:
@fmbenhassine It's a simple and safe improvement, should be merged ASAP.

PR #5179, opened 2025-12-21
Add parameter converters for ZonedDateTime and OffsetDateTime

thswlsqls:
## Description

This PR adds JobParameters converters for `ZonedDateTime` and `OffsetDateTime` to support timezone-aware date/time parameters in Spring Batch jobs.

## Changes

- Add `ZonedDateTimeToStringConverter` and `StringToZonedDateTimeConverter`
- Add `OffsetDateTimeToStringConverter` and `StringToOffsetDateTimeConverter`
- Register new converters in `DefaultJobParametersConverter`
- Add unit tests for all new converters (4 test classes)
- Add integration tests in `DefaultJobParametersConverterTests` for JobParameters with ZonedDateTime/OffsetDateTime
- Update JavaDoc in `DefaultJobParametersConverter` to document the new supported types

## Testing

All tests have been verified to pass successfully. The following test results were obtained when running tests with pre-compiled classes.

### Test Execution Results

**Unit Tests for New Converters:**
```bash
./mvnw test -pl spring-batch-core -Dtest='*ZonedDateTime*Tests,*OffsetDateTime*Tests'
```

**Results:**
- `ZonedDateTimeToStringConverterTests`: 2 tests, all passing
- `StringToZonedDateTimeConverterTests`: 2 tests, all passing
- `OffsetDateTimeToStringConverterTests`: 2 tests, all passing
- `StringToOffsetDateTimeConverterTests`: 2 tests, all passing
- **Total: 8 unit tests, all passing**

**Integration Tests:**
```bash
./mvnw test -pl spring-batch-core -Dtest=DefaultJobParametersConverterTests
```

**Results:**
- `DefaultJobParametersConverterTests`: 22 tests (including 6 new integration tests), all passing
  - `testGetParametersWithZonedDateTime()` - Verifies ZonedDateTime conversion from Properties to JobParameters
  - `testGetParametersWithOffsetDateTime()` - Verifies OffsetDateTime conversion from Properties to JobParameters
  - `testGetPropertiesWithZonedDateTime()` - Verifies ZonedDateTime conversion from JobParameters to Properties
  - `testGetPropertiesWithOffsetDateTime()` - Verifies OffsetDateTime conversion from JobParameters to Properties
  - `testRoundTripWithZonedDateTime()` - Verifies round-trip conversion for ZonedDateTime
  - `testRoundTripWithOffsetDateTime()` - Verifies round-trip conversion for OffsetDateTime

**Full Test Suite:**
```bash
[...]

PR #5186, opened 2025-12-28
Replace custom converters with `DefaultFormattingConversionService`

### LINKED ISSUE #5178: Add ZonedDateTime and OffsetDateTime support to JobParametersConverter
**Expected Behavior**

`ZonedDateTime` and `OffsetDateTime` should be supported as JobParameters types, similar to `LocalDateTime`, `LocalDate`, and `LocalTime`.

Example usage:
```java
ZonedDateTime scheduleTime = ZonedDateTime.of(2023, 12, 25, 10, 30, 0, 0, ZoneId.of("Asia/Seoul"));
JobParameters parameters = new JobParametersBuilder()
    .addJobParameter("schedule.time", scheduleTime, ZonedDateTime.class, true)
    .toJobParameters();
```

**Current Behavior**

Spring Batch currently only provides converters for `LocalDateTime`, `LocalDate`, and `LocalTime`. 
`ZonedDateTime` and `OffsetDateTime` cannot be used as JobParameters because there are no converters available.

**Context**

**How has this issue affected you?**
When working with global services or multi-timezone applications, we need to pass timezone-aware date/time values as JobParameters, but currently only timezone-naive types (`LocalDateTime`, `LocalDate`, `LocalTime`) are supported.

**What are you trying to accomplish?**
- Execute batch jobs based on specific timezones in global services
- Require both UTC and local timezone in log analysis
- Include timezone information for each country in multi-country services

**What other alternatives have you considered?**
- Converting to `LocalDateTime` and storing timezone separately (loses timezone information)
- Using `String` type and parsing manually (error-prone, not type-safe)
- Using `Date` with timezone offset (legacy API, not recommended)

**Are you aware of any workarounds?**
Currently, there is no clean workaround. Users must convert to `LocalDateTime` and lose timezone information, or use `String` type which is not type-safe.

**Proposed Implementation:**
- Add `ZonedDateTimeToStringConverter` and `StringToZonedDateTimeConverter`
- Add `OffsetDateTimeToStringConverter` and `StringToOffsetDateTimeConverter`
- Register new converters in `DefaultJobParametersConverter`
- Add related test code

scordio:
This change replaces almost all existing custom converters with the Spring Framework's `DefaultFormattingConversionService` built-in features. The `Date` converters have been replaced with a custom formatter because the existing Spring Batch logic is incompatible with the framework's capabilities.

`DefaultFormattingConversionService` also supports `ZonedDateTime` and `OffsetDateTime` out of the box, as introduced in #5179.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DateToStringConverter.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DateToStringConverter.java
@@ -27,7 +27,11 @@ import org.springframework.core.convert.converter.Converter;
  *
  * @author Mahmoud Ben Hassine
  * @since 5.0.1
+ * @deprecated since 6.1 in favor of
+ * {@link ConversionServiceFactory#createConversionService()}. Scheduled for removal in
+ * 6.3 or later.
  */
+@Deprecated(since = "6.1", forRemoval = true)
 public class DateToStringConverter extends AbstractDateTimeConverter implements Converter<Date, String> {
 
 	@Override
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JdbcDefaultBatchConfiguration.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JdbcDefaultBatchConfiguration.java
@@ -16,14 +16,7 @@
 package org.springframework.batch.core.configuration.support;
 
 import org.springframework.batch.core.configuration.BatchConfigurationException;
-import org.springframework.batch.core.converter.DateToStringConverter;
-import org.springframework.batch.core.converter.LocalDateTimeToStringConverter;
-import org.springframework.batch.core.converter.LocalDateToStringConverter;
-import org.springframework.batch.core.converter.LocalTimeToStringConverter;
-import org.springframework.batch.core.converter.StringToDateConverter;
-import org.springframework.batch.core.converter.StringToLocalDateConverter;
-import org.springframework.batch.core.converter.StringToLocalDateTimeConverter;
-import org.springframework.batch.core.converter.StringToLocalTimeConverter;
+import org.springframework.batch.core.converter.ConversionServiceFactory;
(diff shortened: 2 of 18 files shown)


scordio:
Raised as a draft to give a glimpse of it. If the direction is accepted, I'm happy to clean up all other usages of the converters across the codebase.

spring-batch-core/src/main/java/org/springframework/batch/core/converter/DateToStringConverter.java
```diff
@@ -27,7 +27,11 @@
  *
  * @author Mahmoud Ben Hassine
  * @since 5.0.1
+ * @deprecated since 6.1 in favor of
+ * {@link ConversionServiceFactory#createConversionService()}. Scheduled for removal in
+ * 6.3 or later.
```
scordio: Since all existing converters are public, I proposed deprecating them for removal in 6.3. Let me know if I should make any adjustments to it.

fmbenhassine:
Thank you for the PR! It's always better to leverage what we can from Spring Framework indeed. I will plan this for 6.1.

I am just not sure yet if we really need the `ConversionServiceFactory` with a static factory method, but I will come back to that in detail when I start working on 6.1.

scordio:
> I am just not sure yet if we really need the `ConversionServiceFactory` with a static factory method

I created that because the `ConversionService` is instantiated five times in the codebase in exactly the same way, so I thought it would be better to centralize it instead of repeating the following:

```java
FormattingConversionService conversionService = new DefaultFormattingConversionService();
conversionService.addFormatterForFieldType(Date.class, new DateFormatter());

DateTimeFormatterRegistrar dateTimeFormatterRegistrar = new DateTimeFormatterRegistrar();
dateTimeFormatterRegistrar.setUseIsoFormat(true);
dateTimeFormatterRegistrar.registerFormatters(conversionService);
```

But I'm totally fine to inline it in those places or go in a different direction I haven't thought of, just let me know your preference 🙂

fmbenhassine:
Rebased and merged as bf0d8458771d1c491c363452424c79c951d1259c.

PR #5189, opened 2025-12-29
Fix race condition in StepContribution.filterCount

### LINKED ISSUE #5188: StepContribution counters are not thread-safe during parallel chunk processing
# Bug Description

When `ChunkOrientedStep` runs in concurrent mode (with `TaskExecutor` configured), `StepContribution.incrementFilterCount()` and `StepContribution.incrementProcessSkipCount()` suffer from race conditions, causing counts to be under-counted. This is because multiple worker threads concurrently call these methods on the same `StepContribution` instance, and both `filterCount` and `processSkipCount` fields use non-thread-safe `long` types with `+=` or `++` operations.

## Environment

- Spring Batch version: 6.0.1
- Java version: 22
- OS: macOS (also reproducible on Linux/Windows)

## Steps to Reproduce

1. Configure a `ChunkOrientedStep` with a `TaskExecutor` to enable concurrent mode
2. Use an `ItemProcessor` that filters all items (returns null)
3. Process a large number of items
4. Check `StepExecution.getFilterCount()` and `StepExecution.getProcessSkipCount()` after execution

## Expected Behavior

`filterCount` and `processSkipCount` should accurately reflect the number of filtered/skipped items.

## Root Cause Analysis

The issue is in `StepContribution.java`:

```java
// Current implementation - NOT thread-safe
private long filterCount = 0;
private long processSkipCount = 0;

public void incrementFilterCount(long count) {
    filterCount += count; // Race condition!
}

public void incrementProcessSkipCount() {
    processSkipCount++; // Race condition!
}
```

In `ChunkOrientedStep.processChunkConcurrently()`, multiple tasks are submitted to the `TaskExecutor`, all sharing the same contribution object:

```java
Future itemProcessingFuture = this.taskExecutor.submit(
    () -> processItem(item, contribution) // Same contribution shared
);
```

Each worker thread calls `contribution.incrementFilterCount()` or `contribution.incrementProcessSkipCount()` when processing items, causing the race condition.

### The Problem with `long +=` and `++` operations

These operations are **not atomic** and consist of multiple steps:

1. Read current value
2. Increment by count
3. Write back to memory

Multiple threads performing these steps simultaneously can cause lost updates:

```
Thread 1: Read (100) → Increment → Write (101)
Thread 2: Read (100) → Increment → Write (101)  ← Lost update!
Expected: 102, Actual: 101
```

## Proposed Solution

Change `filterCount` and `processSkipCount` to `AtomicLong`:

```java
private final AtomicLong filterCount = new AtomicLong(0);
private final AtomicLong processSkipCount = new AtomicLong(0);

public void incrementFilterCount(long count) {
    filterCount.addAndGet(count);
}

public void incrementProcessSkipCount() {
    processSkipCount.incrementAndGet();
}

public long getFilterCount() {
    return filterCount.get();
}

public long getProcessSkipCount() {
    return processSkipCount.get();
}
```

### Approach 1: Partial AtomicLong Application (Current Implementation)

This approach changes only filterCount and processSkipCount to AtomicLong in the StepContribution class, which maintains lock-free performance and API compatibility with minimal code changes, but having only some fields as AtomicLong while others remain long may reduce code consistency and cause confusion for other developers.

### Approach 2: Aggregate Results in Main Thread

Modify `processChunkConcurrently()` to collect processing results and update counts in the main thread only.

```java
record ProcessingResult<O>(@Nullable O item, boolean filtered, boolean skipped) {}

private void processChunkConcurrently(...) {
    List<Future<ProcessingResult<O>>> tasks = new LinkedList<>();

    for (...) {
        Future<ProcessingResult<O>> future = this.taskExecutor.submit(
            () -> processItemWithResult(item)  // Don't pass contribution
        );
        tasks.add(future);
    }

    // Aggregate results in main thread - no race condition
    for (Future<ProcessingResult<O>> future : tasks) {
        ProcessingResult<O> result = future.get();
        if (result.filtered()) ...
        if (result.skipped())...
        if (result.item() != null) ...
    }
}
```

I believe **Approach 1 (AtomicLong)** is the simplest and most practical solution, but I'm open to feedback. If you prefer **Approach 2** or have other suggestions, please let me know and I'll revise accordingly.

KMGeon:
Change filterCount field from long to AtomicLong to ensure thread-safety when incrementing filter count in concurrent chunk processing.

Resolves #5188

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DefaultJobParametersConverter.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DefaultJobParametersConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2025 the original author or authors.
+ * Copyright 2006-present the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -64,6 +64,10 @@ import org.springframework.util.StringUtils;
  * {@link java.time.format.DateTimeFormatter#ISO_LOCAL_TIME} format</li>
  * <li>{@link java.time.LocalDateTime}: in the
  * {@link java.time.format.DateTimeFormatter#ISO_LOCAL_DATE_TIME} format</li>
+ * <li>{@link java.time.ZonedDateTime}: in the
+ * {@link java.time.format.DateTimeFormatter#ISO_ZONED_DATE_TIME} format</li>
+ * <li>{@link java.time.OffsetDateTime}: in the
+ * {@link java.time.format.DateTimeFormatter#ISO_OFFSET_DATE_TIME} format</li>
  * </ul>
  *
  * @author Dave Syer
@@ -85,6 +89,10 @@ public class DefaultJobParametersConverter implements JobParametersConverter {
 		conversionService.addConverter(new StringToLocalTimeConverter());
 		conversionService.addConverter(new LocalDateTimeToStringConverter());
 		conversionService.addConverter(new StringToLocalDateTimeConverter());
+		conversionService.addConverter(new ZonedDateTimeToStringConverter());
+		conversionService.addConverter(new StringToZonedDateTimeConverter());
+		conversionService.addConverter(new OffsetDateTimeToStringConverter());
+		conversionService.addConverter(new StringToOffsetDateTimeConverter());
 		this.conversionService = conversionService;
 	}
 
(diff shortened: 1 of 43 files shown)


quaff:
What about other counts?

KMGeon:
> What about other counts?

Thank you for your feedback! I've added `processSkipCount` to `AtomicLong` as well in the latest commit.

Regarding other count fields (`readSkipCount`, `writeSkipCount`, `writeCount`), they are only modified by single-threaded operations in `ChunkOrientedTasklet`, so they don't have the same race condition issue that `filterCount` and `processSkipCount` have in concurrent chunk processing scenarios.

Let me know if you think those should be converted as well for consistency.

fmbenhassine:
LGTM 👍 Rebased and merged as cc06132c31f760156de69e16bf93828f74a48c21.

Thank you for reporting the issue and contributing a fix!

Issue #4855, opened 2025-06-02
Make MapJobRegistry smart enough to auto register jobs defined in the application context


fmbenhassine:
As of v5.2, it is required to populate the  job registry with a different component (like a bean post processor, or smart initializing bean or something similar) before being able to use it with the `JobOperator`.

This feature request is to make the `MapJobRegistry` smart enough to auto register jobs defined in the application context. This would remove the need for a distinct component to populate the registry and therefore simplifies the configuration. While this could be done by creating a `SmartMapJobRegistry` that extends `MapJobRegistry`, I would keep things simple along the lines of #4847 and make `MapJobRegistry` itself a `SmartInitializingSingleton`.

Other jobs defined outside the application context could be registered manually by the user (same as before).

PR #5200, opened 2026-01-07
Lookup available `JobRegistry` from context for `JobOperatorFactoryBean`

quaff:


fmbenhassine:
Thank you for the PR! However, as mentioned in https://github.com/spring-projects/spring-batch/issues/5229#issuecomment-3847561454 , this introduces a breaking change and is not consistent with how other optional components are customized (why would we use an `ObjectProvider` for the registry but not the for task executor or the job parameters converter?)

The way to fix the issue is by looking for a registry from the context before creating a default one (similar to what we do for the [datasource](https://github.com/spring-projects/spring-batch/blob/4fd95415a2392be12e095e00da3aae606a6709fe/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JdbcDefaultBatchConfiguration.java#L130-L144) and the [transaction manager](https://github.com/spring-projects/spring-batch/blob/4fd95415a2392be12e095e00da3aae606a6709fe/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/JdbcDefaultBatchConfiguration.java#L147-L161) in the JDBC config. Can you please update the PR accordingly? Thank you upfront.

As a side note, regarding https://github.com/spring-projects/spring-boot/pull/48992#issuecomment-3809331647, this is not an oversight. The class is designed to accept mandatory dependencies as constructor parameters, while optional dependencies can be customized by overriding getters (the registry is optional starting from v6).

quaff:
Updated.

quaff:
Another question, is registry job the responsibility of `JobRegistry` implementation or `DefaultBatchConfiguration` ?

I suggest latter, and remove logic from `MapJobRegistry`:
https://github.com/spring-projects/spring-batch/blob/a3a517f0ceecb9ebcc9d7a48862287ade214ce03/spring-batch-core/src/main/java/org/springframework/batch/core/configuration/support/MapJobRegistry.java#L64-L74

fmbenhassine:
Job registration is typically the responsibility of a configuration class (like `DefaultBatchConfiguration`). The reason is it currently done in the `MapJobRegistry` implementation is for people who do **not** use `DefaultBatchConfiguration` and still want to get a populated registry by default when defined as a bean (this actually eliminate the need for a `JobRegistryBeanPostProcessor` or a `JobRegistrySmartInitializingSingleton` as we had in v5, see #4855).

The PR LGTM 👍 Thank you for the updates! I will merge it for the upcoming 6.0.3.

fmbenhassine:
Rebased and merged as ac73b3ab136806b5d1ae331b16465d1890185c90. Thank you for your contribution!

Issue #5201, opened 2026-01-07
SynchronizedItemStreamReader update call is not threadsafe


leonschenk:
**Bug description**
JpaCursorItemReader clears the entityManager upon update. This breaks threadsafety of the update call in case of the use of SynchronizedItemStreamReader, because update method is not synchronized.

**Environment**
Any version of SpringBatch
JpaCursorItemReader based upon jpa implementation Hibernate 6.6
Parallel execution of a JpaCursorItemReader

**Steps to reproduce**
Configure a job with a step containing:
 1. synchronized JpaCursorItemReader with a jpql query, based upon a hibernate entityManager
 2. asynchronous taskExecutor with at least 2 threads
 3. a database that contains abundant items to select with the query (more items than the chunk size)
 4. Debug the testcase, set breakpoints upon read/update of the SynchronizedItemStreamReader
 5. Run the first thread until the update step, it will have read and processed chunk size of items by that time. The lock will be released.
 6. Advance the second thread until hibernate ScrollableResultImpl:131 (					currentRow = getRowReader().readRow( getRowProcessingState() );)
 7. Second thread now has registered the jdbcValuesSourceProcessingState in the persistenceContext.
 8. Advance the first thread until after update, such that the entityManager is cleared.
 9. Advance the second thread now. It will throw NoSuchElementException when trying to deregister the jdbcValuesSourceProcessingState.

`
Caused by: java.util.NoSuchElementException
at org.hibernate.internal.util.collections.StandardStack.pop(StandardStack.java:60) ~[hibernate-core-6.6.33.Final.jar:6.6.33.Final]
at org.hibernate.sql.results.spi.LoadContexts.deregister(LoadContexts.java:43) ~[hibernate-core-6.6.33.Final.jar:6.6.33.Final]
at org.hibernate.internal.ScrollableResultsImpl.prepareCurrentRow(ScrollableResultsImpl.java:144) ~[hibernate-core-6.6.33.Final.jar:6.6.33.Final]
at org.hibernate.internal.ScrollableResultsImpl.next(ScrollableResultsImpl.java:52) ~[hibernate-core-6.6.33.Final.jar:6.6.33.Final]
at org.hibernate.query.internal.ScrollableResultsIterator.hasNext(ScrollableResultsIterator.java:33) ~[hibernate-core-6.6.33.Final.jar:6.6.33.Final]
[...]

PR #5203, opened 2026-01-07
Fix non threadsafe call to update in SynchronizedItemStreamReader

leonschenk:
Related to #5201


fmbenhassine:
Thank you for raising the issue and contributing a PR!

Since the `update` method is called at chunk-boundaries, it is safe to synchronize it. There is no need to introduce a new flag for that. Fixed in 6f15e7209b8b91af229042493a4070232c08cf3e.

leonschenk:
@fmbenhassine thank you for considering this pr!

PR #5235, opened 2026-01-24
Avoid ByteArrayResource fallback when JsonItemReaderBuilder resource s null

### LINKED ISSUE #5234: JsonItemReaderBuilder should not substitute null resource with ByteArrayResource
Environment
Spring Batch 6.0.2

**Expected Behavior**

`JsonItemReaderBuilder` should not replace a `null` `resource` with an empty `ByteArrayResource`. It should align with the “resource can be injected later” pattern used with `FlatFileItemReader`, especially when the reader is used as a delegate of `MultiResourceItemReader`.

```java
if (this.resource == null) {
    logger.debug("The resource is null. This is only a valid scenario when "
            + "injecting it later as in when using the MultiResourceItemReader");
}

JsonItemReader<T> reader = new JsonItemReader<>(this.jsonObjectReader);
reader.setResource(this.resource);
return reader;
```

---

**Current Behavior**

When `resource == null`, the builder replaces it with an empty `ByteArrayResource` and creates the reader with that placeholder.

```java
if (this.resource == null) {
    this.resource = new ByteArrayResource(new byte[0]);
}
JsonItemReader<T> reader = new JsonItemReader<>(this.resource, this.jsonObjectReader);
```

---

**Context**

Keep the resource null in the builder and let the reader validate it at doOpen time (as FlatFileItemReader does), instead of inserting an empty placeholder resource.

banseok1216:
Closes #5234

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JsonItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JsonItemReader.java
@@ -55,11 +55,20 @@ public class JsonItemReader<T> extends AbstractItemCountingItemStreamItemReader<
 
 	private static final Log LOGGER = LogFactory.getLog(JsonItemReader.class);
 
-	private Resource resource;
+	private @Nullable Resource resource;
 
 	private JsonObjectReader<T> jsonObjectReader;
 
 	private boolean strict = true;
+
+	/**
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Thank you for doing this! I think the reason was related to nullability refinement in v6 (non nullable resource) and there was a TODO to check if it's possible to change that.

It's great that you made things consistent with the flat file item reader.

Merged as 61286ec6b22011bcb1d7d6f7048887e6070e015a

Issue #5247, opened 2026-01-31
Incorrect rollbackCount in scan mode in ChunkOrientedStep


KMGeon:
## Bug Description                                                                                                                             
                                                                                                                                                 
  This issue was originally raised in #5210. While #5210 focused on the duplicate data insertion problem, further testing revealed that          
  `rollbackCount` was also incorrect.                                                                                                            
                                                                                                                                                 
  ## Problems Found                                                                                                                              
                                                                                                                                                 
  1. **Duplicate data insertion** - Scan mode executes in the same transaction as failed write                                                   
  2. **Incorrect rollbackCount** - No rollback before scan → `rollbackCount` is not incremented                                                  
                                                                                                                                                 
  ## Steps to Reproduce                                                                                                                          
                                                                                                                                                 
  See #5210 for example projects comparing Spring Batch 5 and 6 behavior.                                                                        
                                                                                                                                                 
  ## Expected Behavior                                                                                                                           
                                                                                                                                                 
  - Scan mode should execute in a **new transaction** after rollback (same as `FaultTolerantChunkProcessor` in Spring Batch 5)                   
  - `rollbackCount` should reflect actual transaction rollbacks                                                                                  
                                                                                                                                                 
  ## Actual Behavior                                                                                                                             
                                                                                                                                                 
  | Issue | Description |                                                                                                                        
  |-------|-------------|                                                                                                                        
  | Duplicate data | Scan executes without rollback, causing items to be written twice |                                                         
  | Incorrect rollbackCount | No rollback before scan → `rollbackCount` stays 0 |                                                                
                                                                                                                                                 
  ---                                                                                                                                            
                                                                                                                                                 
  ## Solution                                                                                                                                    
                                                                                                                                                 
  ### Before (Bug)                                                                                                                               
  [TX 1] write(1,2,3) → fail → scan(1,2,3) → commit                                                                                              
                                ↑                                                                                                                
                      item "1" written twice (duplicate!)                                                                                        
                                                                                                                                                 
  ### After (Fixed)                                                                                                                              
  [TX 1] write(1,2,3) → fail → save state → ROLLBACK                                                                                             
                                                ↓                                                                                                
  [TX 2] scan(1,2,3) → write each item → commit                                                                                                  
                                ↑                                                                                                                
                      item "1" written once (correct!)                                                                                           
                                                                                                                                                 
  ### Key Changes                                                                                                                                
                                                                                                                                                 
[...]

PR #5248, opened 2026-02-01
Fix duplicate writes in scan mode by deferring scan
to new transaction

### LINKED ISSUE #5210: [Spring Batch 6.0.1] Transaction is not rolled back before chunk scanning in ChunkOrientedStep with skip policy.
**Bug description**
When using `ChunkOrientedStep` (fault-tolerant) with a given skip policy, the current transaction is not rolled back before chunk scanning starts. As a result, the persistence context is not cleared. This causes any Hibernate entities created and added to the persistence context during the writing of the full chunk to remain in the persistence context. Consequently, this can lead to a false belief that an entity existed before the chunk writing began, potentially causing logical errors inside the writer during chunk scanning.

This is a regression in spring-batch 6.x, the issue was not present in spring-batch 5.x

**Environment**
- **Spring Batch version**: 6.0.1
- **Hibernate** 7+
- **Java version**: 17+
- **Database**: H2, PostgreSQL, MySQL (issue is not database-specific)

**Steps to reproduce**
1. Configure a `ChunkOrientedStep` with fault tolerance enabled and a skip policy.
2. Write a chunk that creates Hibernate-managed entities.
3. Trigger a scenario where chunk scanning starts (e.g., an exception occurs during chunk processing).
4. Observe that the transaction is not rolled back and the persistence context is not cleared before the chunk scanning starts.

**Expected behavior**
When chunk scanning starts, the current transaction should be rolled back, and the persistence context should be cleared. This ensures that no Hibernate-managed entities from the chunk writing remain in the persistence context, avoiding logical errors in the writer.

Below is a minimal example to reproduce the issue. In this example, we demonstrate that item 1 (created during initial chunk writing) is still present in the persistence context after chunk scanning has started (errors are thrown on writing of items 2 and 3).

Full code: [spring-batch-mcve-issue.zip](https://github.com/user-attachments/files/24595267/spring-batch-mcve-issue.zip)

```java
@Bean
public ListItemReader<String> reader() {
    return new ListItemReader<>(Arrays.asList("1", "2", "3"));
}

@Bean
public ItemWriter<String> writer(EntityManager em) {
    return items -> {
        System.out.println(); System.out.println("Writing chunk of items..." + items); System.out.println();
        for (String number : items) {
            System.out.println("Writing item: " + number);
            writeItem(em, number);
            System.out.println();
        }
    };
}

private void writeItem(EntityManager em, String number) {
    if (number.equals("2") || number.equals("3")) {
        System.out.println("Simulating error on item: " + number);
        throw new RuntimeException("Simulated write error for item: " + number);
    } else {
        Delivery existingDelivery = em.find(Delivery.class, number);
        if (existingDelivery != null) {
            System.out.println("Delivery with number " + number + " already exists but shouldn't: " + existingDelivery);
        } else {
            System.out.println("Delivery with number " + number + " does not exist.");
            em.persist(new Delivery(number));
        }
    }
}

@Bean
public Step step(JobRepository jobRepository, PlatformTransactionManager tm, ListItemReader<String> reader, ItemWriter<String> writer) {
    return new StepBuilder("step", jobRepository)
            .<String, String>chunk(3)
            .transactionManager(tm)
            .reader(reader)
            .writer(writer)
            .faultTolerant()
            .skip(Exception.class)
            .skipLimit(1)
            .build();
}

@Entity
public class Delivery {

    @Id
    private String number;

    public Delivery() {
    }

    public Delivery(String number) {
        this.number = number;
    }

    public String getNumber() {
        return number;
    }

    public void setNumber(String number) {
        this.number = number;
    }

    @Override
    public String toString() {
        return "Delivery{" +
                "number='" + number + '\'' +
                '}';
    }
}
```

**Output and indication of where rollback is expected**
```
Writing chunk of items...[items=[1, 2, 3], skips=[]]

Writing item: 1
Delivery with number 1 does not exist.

Writing item: 2
Simulating error on item: 2
<-- Transaction is NOT rolled back here, persistence context is NOT cleared -->
Writing chunk of items...[items=[1], skips=[]]

Writing item: 1
Delivery with number 1 already exists but shouldn't: Delivery{number='1'}

Writing chunk of items...[items=[2], skips=[]]

Writing item: 2
Simulating error on item: 2

Writing chunk of items...[items=[3], skips=[]]

Writing item: 3
Simulating error on item: 3
```

KMGeon:
Fixes #5210, #5247. When chunk write fails in fault-tolerant mode, scan() was called within the same transaction,
causing duplicate data insertion. Changes: Defer scan execution to new transaction after rollback, Add ChunkTracker.scanMode and pendingChunk,
Skip JobRepository update when rollback is pending.

### DIFF
--- a/spring-batch-core/src/test/resources/schema.sql
+++ b/spring-batch-core/src/test/resources/schema.sql
@@ -1,1 +1,2 @@
 create table person_target (id int primary key, name varchar(5));
+create table delivery (id int auto_increment primary key, item_number varchar(255));
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkOrientedStep.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkOrientedStep.java
@@ -54,6 +54,7 @@ import org.springframework.batch.core.step.skip.NeverSkipItemSkipPolicy;
 import org.springframework.batch.core.step.skip.NonSkippableProcessException;
 import org.springframework.batch.core.step.skip.NonSkippableReadException;
 import org.springframework.batch.core.step.skip.NonSkippableWriteException;
+import org.springframework.batch.core.step.skip.SkipLimitExceededException;
 import org.springframework.batch.core.step.skip.SkipPolicy;
 import org.springframework.batch.infrastructure.item.Chunk;
 import org.springframework.batch.infrastructure.item.ExecutionContext;
@@ -133,7 +134,7 @@ public class ChunkOrientedStep<I, O> extends AbstractStep {
 	 */
 	private final int chunkSize;
 
-	private final ThreadLocal<ChunkTracker> chunkTracker = ThreadLocal.withInitial(ChunkTracker::new);
+	private final ThreadLocal<ChunkTracker<O>> chunkTracker = ThreadLocal.withInitial(ChunkTracker::create);
 
 	private final CompositeChunkListener<I, O> compositeChunkListener = new CompositeChunkListener<>();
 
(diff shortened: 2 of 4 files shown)


spring-batch-core/src/test/resources/schema.sql
```diff
@@ -1 +1,2 @@
-create table person_target (id int primary key, name varchar(5));
\ No newline at end of file
+create table person_target (id int primary key, name varchar(5));
+create table delivery (id int auto_increment primary key, item_number varchar(255));
```
KMGeon: Added `auto_increment` to the `delivery` table for testing duplicate insertion.                                                                
                                                                                                                                                 
  Without `auto_increment`:                                                                                                                      
  - Duplicate insert → Primary key conflict → Error                                                                                              
  - Cannot verify if duplicate data was inserted                                                                                                 
                                                                                                                                                 
  With `auto_increment`:                                                                                                                         
  - Duplicate insert → Different id auto-generated → Success                                                                                     
  - Can verify duplicate by `SELECT COUNT(*) WHERE item_number = '1'`                                                                            
                                                                                                                                                 
  This is required to test the GH-5210 duplicate insertion bug.

fmbenhassine:
That's great! And the addition of the "scan mode" flag in the chunk tracker is neat idea 👌 

Rebased and merged as b24a85eb56925d200544dddeecf8c28173779277. Thank you for your contribution, much appreciated 🙏

Issue #5252, opened 2026-02-01
StepExecutionDao violate JobRepository API contract


marschall:
**Bug description**
The StepExecutionDao classes that are used by `SimpleJobRepository` / `SimpleJobExplorer` to implement the `getStepExecutionCount` `throws NoSuchStepException` violate their API contract.

Specifically: `org.springframework.batch.core.repository.JobRepository#getStepExecutionCount(JobInstance, String)` returns `0` rather than throwing `NoSuchStepException`. `org.springframework.batch.core.job.SimpleStepHandler#shouldStart(StepExecution, JobExecution, Step)` currently relies on the broken behavior and breaks if `NoSuchStepException` is thrown.

**Environment**
Spring Batch version: 6.0.2
Java version: 25.0.1
Database: H2 2.4.240

**Steps to reproduce**
Steps to reproduce the issue.

**Expected behavior**

The implemented behavior matches the API contract. Looking at the callers, they do or should all threat `NoSuchStepException` as `0` so it is debatable whether the API contract should be updated instead.

**Minimal Complete Reproducible example**

In `org.springframework.batch.core.repository.support.AbstractJobRepositoryIntegrationTests`

```java
	/*
	 * Step execution count throws if no executions.
	 */
	@Test
	void testGetStepExecutionCountThrows() throws Exception {
		job.setRestartable(true);
		StepSupport step = new StepSupport("restartedStep");

		ExecutionContext executionContext = new ExecutionContext();
		JobInstance jobInstance = jobRepository.createJobInstance(job.getName(), jobParameters);
		// first execution
		JobExecution firstJobExec = jobRepository.createJobExecution(jobInstance, jobParameters, executionContext);
		assertThrows(NoSuchStepException.class, () -> jobRepository.getStepExecutionCount(jobInstance, step.getName()));
	}
```

quaff:
`JobRepository::getStepExecutionCount` API doesn't state whether `NoSuchStepException` will be thrown or not like `JobRepository::getJobInstanceCount` did.
https://github.com/spring-projects/spring-batch/blob/c121edd04b428cdf2a0bb8392602d6486213f358/spring-batch-core/src/main/java/org/springframework/batch/core/repository/JobRepository.java#L138-L139

---
> `org.springframework.batch.core.job.SimpleStepHandler#shouldStart(StepExecution, JobExecution, Step)` currently relies on the broken behavior and breaks if `NoSuchStepException` is thrown.

[...]

PR #5257, opened 2026-02-02
Improve `SimpleStepHandler::shouldStart` to handle NoSuchStepException correctly

quaff:
We should treat no such step as step was not executed before.

See https://github.com/spring-projects/spring-batch/issues/5252


fmbenhassine:
This was fixed as part of db57cf5f3e8a2f6d7bd026f31536bce25c26c366.

Thank you for the PR anyway 🙏

PR #5275, opened 2026-02-07
Fix MongoJobInstanceDao to comply with JobInstanceDao contract

### LINKED ISSUE #5274: MongoJobInstanceDao: getJobNames() violates interface contract and getJobInstances() performs pagination in memory
**Bug description**

`MongoJobInstanceDao` has two issues in the current `main` branch:

### 1. `getJobNames()` violates the `JobInstanceDao` interface contract

The [`JobInstanceDao#getJobNames()`](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JobInstanceDao.java#L129) contract states:

> "Retrieve the names of all jobs for which job instances exist, **sorted alphabetically**."

The JDBC implementation correctly uses `SELECT DISTINCT JOB_NAME FROM BATCH_JOB_INSTANCE ORDER BY JOB_NAME`.

However, the MongoDB implementation loads **all documents** into memory without deduplication or sorting:

```java
public List<String> getJobNames() {
    return this.mongoOperations
        .findAll(JobInstance.class, COLLECTION_NAME)
        .stream()
        .map(JobInstance::getJobName)
        .toList();
}
```

Three problems:
- **`findAll()`** loads every document into memory — potential OOM in production with many job instances
- **No deduplication** — returns duplicate job names (JDBC uses `DISTINCT`)
- **No sorting** — interface contract requires alphabetical order

### 2. `getJobInstances(String, int, int)` performs pagination in memory instead of delegating to MongoDB

The current implementation fetches **all** matching documents and then paginates in Java:

```java
public List<JobInstance> getJobInstances(String jobName, int start, int count) {
    List<...> jobInstances = this.mongoOperations
        .find(query.with(Sort.by(sortOrder)), ...).stream().toList(); // loads ALL
    if (jobInstances.size() <= start) {
        return Collections.emptyList();
    }
    return jobInstances.subList(start, Math.min(jobInstances.size(), start + jobInstances.size()))
        .stream()
        .map(this.jobInstanceConverter::toJobInstance)
        .limit(count)
        .toList();
}
```

Two problems:
- **All documents loaded into memory** — pagination should use MongoDB's `skip()` and `limit()` instead
- **`Math.min(jobInstances.size(), start + jobInstances.size())`** always evaluates to `jobInstances.size()` (since `start >= 0`), making the upper bound calculation ineffective

**Environment**

- Spring Batch: main branch (6.x)
- Java: 17+
- Database: MongoDB

**Expected behavior**

- `getJobNames()` should return **distinct**, **alphabetically sorted** job names without loading all documents — consistent with the JDBC implementation and the `JobInstanceDao` interface contract
- `getJobInstances()` should delegate pagination to MongoDB using `skip()` and `limit()` instead of performing it in memory

The existing integration test `MongoJobInstanceDaoIntegrationTests#testGetJobNames` does not verify deduplication or sorting, so it does not catch the issue. The following test demonstrates the problem:

```java
@Test
void testGetJobNamesShouldReturnDistinctAndSortedNames() {
    // Create multiple instances of the same job
    dao.createJobInstance("batchJob", new JobParameters());
    dao.createJobInstance("batchJob", new JobParametersBuilder().addLong("id", 1L).toJobParameters());
    dao.createJobInstance("alphaJob", new JobParameters());

    List<String> jobNames = dao.getJobNames();

    // Should be deduplicated
    assertEquals(2, jobNames.size());
    // Should be sorted alphabetically
    assertEquals("alphaJob", jobNames.get(0));
    assertEquals("batchJob", jobNames.get(1));
}
```

combe4259:
Closes gh-5274

`MongoJobInstanceDao` has two issues:

1. `getJobNames()` uses `findAll()` which loads all documents into memory without deduplication or sorting. The `JobInstanceDao` interface contract requires distinct, alphabetically sorted names. Replaced with `findDistinct()` with sorting.

2. `getJobInstances()` fetches all matching documents and paginates in Java. Replaced with MongoDB `skip()` and `limit()`.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/mongodb/MongoJobInstanceDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/mongodb/MongoJobInstanceDao.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024-2025 the original author or authors.
+ * Copyright 2024-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -112,20 +112,14 @@ public class MongoJobInstanceDao implements JobInstanceDao {
 
 	@Override
 	public List<JobInstance> getJobInstances(String jobName, int start, int count) {
-		Query query = query(where("jobName").is(jobName));
-		Sort.Order sortOrder = Sort.Order.desc("jobInstanceId");
-		List<org.springframework.batch.core.repository.persistence.JobInstance> jobInstances = this.mongoOperations
-			.find(query.with(Sort.by(sortOrder)),
-					org.springframework.batch.core.repository.persistence.JobInstance.class, COLLECTION_NAME)
-			.stream()
-			.toList();
-		if (jobInstances.size() <= start) {
-			return Collections.emptyList();
-		}
-		return jobInstances.subList(start, Math.min(jobInstances.size(), start + jobInstances.size()))
+		Query query = query(where("jobName").is(jobName))
+			.with(Sort.by(Sort.Order.desc("jobInstanceId")))
+			.skip(start)
+			.limit(count);
+		return this.mongoOperations
+			.find(query, org.springframework.batch.core.repository.persistence.JobInstance.class, COLLECTION_NAME)
 			.stream()
 			.map(this.jobInstanceConverter::toJobInstance)
-			.limit(count)
 			.toList();
 	}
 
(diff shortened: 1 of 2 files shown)


combe4259:
@fmbenhassine 
Hi, Could you please review this PR when you have a chance?

I’ve updated MongoJobInstanceDao to better comply with the JobInstanceDao contract by improving the getJobNames() and getJobInstances() implementations.

I’d really appreciate any feedback
 Thank you

fmbenhassine:
Hi @combe4259 

Thank you for raising this valid issue and for contributing a PR!

The fix LGTM and goes in the right direction, however there's a build failure with this change set:

```
[ERROR] Failures:
[ERROR]   MongoJobInstanceDaoIntegrationTests.testGetLastInstancesPastEnd:285 expected: <0> but was: <2>
```

Can you please check? Thank you upfront.

combe4259:
Hi @fmbenhassine

Thank you for the review!

The failure was caused by MongoDB's `limit(0)` behaving as "no limit" rather than "limit to zero results", so `getJobInstances(jobName, 0, 0)` was returning all matching documents instead of an empty list. Fixed by adding an early return when `count` is zero.

All tests in `MongoJobInstanceDaoIntegrationTests` are now passing.

fmbenhassine:
LGTM now 👍 Rebased and merged as 518611c8a8e91f28fe106ac0036e436e1c8d09af.

Thank you for your contribution !

PR #5282, opened 2026-02-12
Fix make JsonObjectReader close operations null-safe

### LINKED ISSUE #5281: NullPointerException in   JacksonJsonObjectReader.close() and   GsonJsonObjectReader.close()
**Bug description**
When `JsonItemReader.open()` fails during resource.getInputStream( , the `inputStream` field in `JsonObjectReader` implementations is never initialized. Later, when `close()` is called in a finally block, a `NullPointerException` occurs in the close phase, which gets wrapped in an `ItemStreamException`.

The root cause is that both `JacksonJsonObjectReader.close()` and `GsonJsonObjectReader.close()` do not perform a null check before calling `inputStream.close()`. Even though `JsonItemReader.doOpen()` checks `resource.isReadable()` before calling `jsonObjectReader.open()`, the `Resource.isReadable()` javadoc explicitly states: "Note that actual content reading may still fail when attempted." This means `inputStream` may still be `null` when `getInputStream()` throws an exception during `open()`.

**Environment**
- Spring Batch version: 6.0.2

**Steps to reproduce**
1. Create a `JsonItemReader` with a `Resource` that throws an `IOException` when `getInputStream()` is called
2. Call `itemReader.open(new ExecutionContext())`
3. The open fails with `ItemStreamException`
4. In the finally block, call `itemReader.close()`
5. `NullPointerException` is thrown from `JacksonJsonObjectReader.close()` or `GsonJsonObjectReader.close()`

**Expected behavior**
The `close()` method should be null-safe and gracefully handle the case where `inputStream` was never initialized due to a failed `open()` operation. No additional exception should be thrown during cleanup.

```java
package org.springframework.batch.infrastructure.item.json;

import java.io.IOException;
import java.io.InputStream;

import org.springframework.batch.infrastructure.item.ExecutionContext;
import org.springframework.batch.infrastructure.item.ItemStreamException;
import org.springframework.core.io.AbstractResource;
import org.springframework.core.io.Resource;

import static org.junit.jupiter.api.Assertions.assertThrows;

class JsonObjectReaderNpeTests {

    @Test
    void testJacksonJsonItemReaderIntegrationNpe() throws Exception {
        Resource resource = new FailingResource();
        JacksonJsonObjectReader<String> jsonObjectReader = new JacksonJsonObjectReader<>(String.class);
        JsonItemReader<String> itemReader = new JsonItemReader<>(resource, jsonObjectReader);

        try {
            assertThrows(ItemStreamException.class, () -> itemReader.open(new ExecutionContext()));
        }
        finally {
            itemReader.close();
        }
    }

    @Test
    void testGsonJsonItemReaderIntegrationNpe() throws Exception {
        Resource resource = new FailingResource();
        GsonJsonObjectReader<String> jsonObjectReader = new GsonJsonObjectReader<>(String.class);
        JsonItemReader<String> itemReader = new JsonItemReader<>(resource, jsonObjectReader);

        try {
            assertThrows(ItemStreamException.class, () -> itemReader.open(new ExecutionContext()));
        }
        finally {
            itemReader.close();
        }
    }

    private static class FailingResource extends AbstractResource {

        @Override
        public String getDescription() {
            return "fail";
        }

        @Override
        public InputStream getInputStream() throws IOException {
            throw new IOException("Connection failed during open");
        }

    }

}
```

**Stack trace**

```
org.springframework.batch.infrastructure.item.ItemStreamException: Error while closing item reader

    at org.springframework.batch.infrastructure.item.support.AbstractItemCountingItemStreamItemReader.close(AbstractItemCountingItemStreamItemReader.java:143)
    at org.springframework.batch.infrastructure.item.json.JsonObjectReaderNpeTests.testJacksonJsonItemReaderIntegrationNpe(JsonObjectReaderNpeTests.java:27)
Caused by: java.lang.NullPointerException: Cannot invoke "java.io.InputStream.close()" because "this.inputStream" is null
    at org.springframework.batch.infrastructure.item.json.JacksonJsonObjectReader.close(JacksonJsonObjectReader.java:112)
    at org.springframework.batch.infrastructure.item.json.JsonItemReader.doClose(JsonItemReader.java:125)
    at org.springframework.batch.infrastructure.item.support.AbstractItemCountingItemStreamItemReader.close(AbstractItemCountingItemStreamItemReader.java:140)
    ... 1 more
```

**Fix**
Add null check in `JacksonJsonObjectReader.close()` and `GsonJsonObjectReader.close()` methods before closing `inputStream`.

**Reference**
```java
	/**
	 * Indicate whether non-empty contents of this resource can be read via
	 * {@link #getInputStream()}.
	 * <p>Will be {@code true} for typical resource descriptors that exist
	 * since it strictly implies {@link #exists()} semantics as of 5.1.
	 * Note that actual content reading may still fail when attempted.
	 * However, a value of {@code false} is a definitive indication
	 * that the resource content cannot be read.
	 * @see #getInputStream()
	 * @see #exists()
	 */
	default boolean isReadable() {
		return exists();
	}
```

banseok1216:
Closes #5281

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/GsonJsonObjectReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/GsonJsonObjectReader.java
@@ -97,11 +97,14 @@ public class GsonJsonObjectReader<T> implements JsonObjectReader<T> {
 		return null;
 	}
 
-	@SuppressWarnings("DataFlowIssue")
 	@Override
 	public void close() throws Exception {
-		this.inputStream.close();
-		this.jsonReader.close();
+		if (inputStream != null) {
+			inputStream.close();
+		}
+		if (jsonReader != null) {
+			jsonReader.close();
+		}
 	}
 
 	@SuppressWarnings("DataFlowIssue")
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JacksonJsonObjectReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JacksonJsonObjectReader.java
@@ -106,11 +106,14 @@ public class JacksonJsonObjectReader<T> implements JsonObjectReader<T> {
 		return null;
 	}
 
-	@SuppressWarnings("DataFlowIssue")
 	@Override
 	public void close() throws Exception {
-		this.inputStream.close();
-		this.jsonParser.close();
+		if (inputStream != null) {
+			inputStream.close();
+		}
+		if (jsonParser != null) {
+			jsonParser.close();
+		}
 	}
 
 	@SuppressWarnings("DataFlowIssue")


fmbenhassine:
LGTM 👍 Thank you for raising the issue and for contributing a fix!

Issue #5285, opened 2026-02-13
Inaccurate JavaDoc of BatchStatus.isLessThanOrEqualTo


yuseogi0218:
**Bug description**

The JavaDoc of `BatchStatus.isLessThanOrEqualTo(BatchStatus other)` is slightly misleading.

It currently states:

> @return {@code true} if this is less than {@code other}.

However, the method actually checks for `less than or equal to`.

**Expected behavior**

Update the JavaDoc to accurately reflect that the method returns true when this status is less than or equal to the given status.

> @return {@code true} if this is less than or equalt to {@code other}.

yuseogi0218:
PR submitted: #5286

hqqw2h-lgtm:
I'd like to work on this. Will fix the JavaDoc to correctly state 'less than or equal to'.

fmbenhassine:
Resolved with #5286

PR #5286, opened 2026-02-13
GH-5285: Clarify JavaDoc of BatchStatus.isLessThanOrEqualTo

yuseogi0218:
The JavaDoc of `BatchStatus.isLessThanOrEqualTo(BatchStatus other)`
was slightly misleading as it only mentioned "less than" while the
method actually performs a "less than or equal to" comparison.

This update clarifies the documentation to accurately reflect the
method behavior.

No functional changes.


fmbenhassine:
Thank you for raising the issue and contributing a fix!

PR #5296, opened 2026-02-19
GH-5295: Fix parameter value truncation in CommandLineJobOperator

### LINKED ISSUE #5295: CommandLineJobOperator.parse() truncates parameter values containing '='
**Bug description**

`CommandLineJobOperator.parse()` uses `String.split("=")` without a limit, which splits on every `=` character. This causes two problems:

1. Parameter values containing `=` get truncated (e.g. URLs with query strings)
2. Parameters without `=` throw `ArrayIndexOutOfBoundsException`

The deprecated `CommandLineJobRunner` and Spring Boot's `JobLauncherApplicationRunner` both use `StringUtils.splitArrayElementsIntoProperties` which correctly splits only on the first `=`.

**Environment**

Spring Batch 6.0.x (all versions since `CommandLineJobOperator` was introduced in 6.0)

**Steps to reproduce**

Run a job via `CommandLineJobOperator` with a parameter containing `=` in the value:

```
java ... CommandLineJobOperator com.example.MyJobConfig start myJob url=http://example.com?id=123
```

**Expected behavior**

The parameter should be parsed as `url` → `http://example.com?id=123`.

**Current behavior**

`split("=")` produces `["url", "http://example.com?id", "123"]`, so only `http://example.com?id` is stored as the value. The rest is lost.

**Relevant code**

https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobOperator.java#L406-L413

For comparison, `JobLauncherApplicationRunner` handles this correctly:
https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-batch/src/main/java/org/springframework/boot/batch/autoconfigure/JobLauncherApplicationRunner.java#L148

Wordbe:
`CommandLineJobOperator.parse()` uses `String.split("=")` without a limit, which splits on every `=` character. This truncates parameter values that contain `=` (e.g. URLs with query strings) and throws `ArrayIndexOutOfBoundsException` for parameters without `=`.

Replaced with `StringUtils.splitArrayElementsIntoProperties`, consistent with `JobLauncherApplicationRunner` and the deprecated `CommandLineJobRunner`.

Resolves #5295

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobOperator.java
@@ -34,6 +34,7 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 import static org.springframework.batch.core.launch.support.ExitCodeMapper.JVM_EXITCODE_COMPLETED;
 import static org.springframework.batch.core.launch.support.ExitCodeMapper.JVM_EXITCODE_GENERIC_ERROR;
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 36c4fc1838c6d837e588f30baa9d698483977009.

Thank you for reporting the issue and for contributing a PR! Much appreciated 🙏

Issue #4858, opened 2025-06-03
Add ability to use bean names for jobs and steps


fmbenhassine:
As of v5.2, both `AbstractJob` and `AbstractStep` implement `BeanNameAware`. So it could be possible to use the bean name as a default job/step name in order to avoid duplication:

``` diff
@Bean
public Job job(JobRepository jobRepository, Step step) {
--	return new JobBuilder("job", jobRepository)
++	return new JobBuilder(jobRepository)
                           .start(step)
                           .build();
}
```

Obviously, the current constructor that takes the job/step name should remain to be able to specify a different name if needed.

cc @joshlong

PR #5299, opened 2026-02-19
Fix step name being null when not defined as bean

### LINKED ISSUE #5278: Step name is null when the step is not defined as a bean and not assigned a name
The name of the step should default to the class name if the step is not defined as a bean, but it is currently `null`, see logs below:

```java
package org.springframework.batch.samples.helloworld;

import org.springframework.batch.core.job.Job;
import org.springframework.batch.core.job.JobExecution;
import org.springframework.batch.core.job.builder.JobBuilder;
import org.springframework.batch.core.job.parameters.JobParameters;
import org.springframework.batch.core.launch.JobOperator;
import org.springframework.batch.core.step.Step;
import org.springframework.batch.core.step.builder.StepBuilder;
import org.springframework.batch.infrastructure.repeat.RepeatStatus;

@Configuration
@EnableBatchProcessing
public class HelloWorldJobConfiguration {

		ApplicationContext context = new AnnotationConfigApplicationContext(HelloWorldJobConfiguration.class);
		JobOperator jobOperator = context.getBean(JobOperator.class);
		jobOperator.start(job, new JobParameters());
	}

	@Bean
	public Job job(JobRepository jobRepository) {
		Step step = new StepBuilder(jobRepository).tasklet((contribution, chunkContext) -> {
			System.out.println("Hello world!");
			return RepeatStatus.FINISHED;
		}).build();
		return new JobBuilder(jobRepository).start(step).build();
	}

}
```

prints:

```
07:47:25.970 [main] INFO  o.s.b.c.l.s.TaskExecutorJobLauncher - Job: [SimpleJob: [name=job]] launched with the following parameters: [{}]
07:47:25.973 [main] INFO  o.s.batch.core.job.SimpleStepHandler - Executing step: [null]
Hello world!
07:47:25.978 [main] INFO  o.s.batch.core.step.AbstractStep - Step: [null] executed in 4ms
07:47:25.979 [main] INFO  o.s.b.c.l.s.TaskExecutorJobLauncher - Job: [SimpleJob: [name=job]] completed with the following parameters: [{}] and the following status: [COMPLETED] in 5ms
```

EDIT:

This is more than cosmetic: with a JDBC-backed `JobRepository`, `SimpleJobRepository.createStepExecution()` rejects a null step name with an `IllegalArgumentException`, since `BATCH_STEP_EXECUTION.STEP_NAME` is `NOT NULL` - so affected jobs fail outright rather than merely logging "[null]".

therepanic:
In this commit, if Step is not a bean, we give it a class name, thereby preventing null in the class name, which is prohibited by contract.

Closes: gh-5278

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/StepBuilderHelper.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/StepBuilderHelper.java
@@ -42,6 +42,7 @@ import org.springframework.batch.infrastructure.support.ReflectionUtils;
  * @author Michael Minella
  * @author Taeik Lim
  * @author Mahmoud Ben Hassine
+ * @author Andrey Litvitski
  * @since 2.2
  */
 // FIXME remove once default constructors (required by the XML namespace) are removed
@@ -137,6 +138,9 @@ public abstract class StepBuilderHelper<B extends StepBuilderHelper<B>> {
 	}
 
 	protected void enhance(AbstractStep step) {
+		if (step.getName() == null) {
+			step.setName(step.getClass().getSimpleName());
+		}
 		step.setJobRepository(properties.getJobRepository());
 
 		ObservationRegistry observationRegistry = properties.getObservationRegistry();
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/step/StepBuilderHelperTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/step/StepBuilderHelperTests.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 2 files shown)


therepanic:
Test crashed because of flaky tests, I think

fmbenhassine:
Got a build failure in `ChunkOrientedStepObservabilityIntegrationTests` when testing this PR. I wanted to include it in 6.0.3, but I will check for a later iteration.

fmbenhassine:
> Got a build failure in `ChunkOrientedStepObservabilityIntegrationTests` when testing this PR. I wanted to include it in 6.0.3, but I will check for a later iteration.

In that test, the steps are defined as beans without being assigned a name, so the step name is expected to be the bean name (as in #4858), but with this PR, the step name is now the class name instead of the bean name. In other terms, this change breaks the new feature in #4858. This is also visible on the hello world sample which now prints:

```
10:56:49.754 [main] INFO  o.s.b.c.s.AbstractStep - Executing step: [org.springframework.batch.core.step.tasklet.TaskletStep]
```

instead of 

```
10:57:15.790 [main] INFO  o.s.b.c.s.AbstractStep - Executing step: [step]
```

Can you please check? I will check on my side as well. Thank you upfront.

fmbenhassine:
> I will check on my side as well.

I checked this and I found that trying to set the step name anywhere in the builder is too early as it will prevent to use the bean name as a step name if the step is defined as a bean (ie break #4858)

`SimpleStepHandler.handleStep()` is the single choke point where every execution path first consults `Step#getName()` to create/look up StepExecutions, and it only runs once the whole ApplicationContext has finished initializing. So by then, `setBeanName()` has already run for any step that is a bean. Therefore, falling back there, only when the name is still `null`, cannot clobber an explicit name or a bean name.

I pushed a fix along these lines, see 9802737ecc2fd56b0f81839a9a2a6f5ad032c404. Thank you for your contribution anyway 🙏

PR #5315, opened 2026-03-04
Fix CommandLineJobOperator parsing for '=' in values

### LINKED ISSUE #5295: CommandLineJobOperator.parse() truncates parameter values containing '='

answndud:
## Summary
- parse command line job parameters on the first `=` only
- ignore malformed parameters without `=` instead of failing with `ArrayIndexOutOfBoundsException`
- add tests for values containing `=` and malformed parameters

## Why
`CommandLineJobOperator.parse` currently uses `String.split("=")`, which truncates values that contain `=` and can fail on malformed input.

Closes #5295

## Testing
- Attempted locally: `./mvnw -Dmaven.user.home=/tmp/m2 -pl spring-batch-core -Dtest=org.springframework.batch.core.launch.support.CommandLineJobOperatorTests test`
- Local build is blocked by Error Prone/NullAway JSpecify requirement (JDK 22+).

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobOperator.java
@@ -34,6 +34,7 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.springframework.core.log.LogAccessor;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 import static org.springframework.batch.core.launch.support.ExitCodeMapper.JVM_EXITCODE_COMPLETED;
 import static org.springframework.batch.core.launch.support.ExitCodeMapper.JVM_EXITCODE_GENERIC_ERROR;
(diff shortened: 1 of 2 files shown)


nikitanagar08:
I would like to work on this issue. I understand the problem:

**Current behavior:**
- `String.split("=")` splits on ALL `=` characters, truncating values that contain `=`
- Malformed parameters cause `ArrayIndexOutOfBoundsException`

**Proposed fix:**
- Parse only on the FIRST `=` character
- Handle malformed parameters gracefully instead of throwing exception
- Add tests for both cases

I will start implementing this fix.

fmbenhassine:
Closing in favor #5296

PR #5317, opened 2026-03-05
Restore backward-compatible default job key format

### LINKED ISSUE #5283: Spring Batch 6 breaks compatibility with existing job keys
Note: I'm not actually sure whether this is intentional or not. If it is, I didn't see any mention of it, but it feels like something that should explicitly be mentioned in the Spring Batch 6 migration guide. The docs do mention changes to the Job Parameter APIs, but they don't seem to indicate that these changes will affect persisted data.

``DefaultJobKeyGenerator`` was refactored here: https://github.com/86dh/spring-batch/commit/2d9bbaee2c46a942fb972f3993020876159f10a6#diff-6e38cfa7139610dfc635ee6285d477356902cd01916529efa2fbca15ab6b1b01

The refactor changes the format of generated job keys, breaking backward compatibility with keys stored in the job repository database for existing job instances.

Old format: ``key={value=..., type=..., identifying=...}``
New format: ``JobParameter{name='...', value=..., type=..., identifying=...}``

Jobs with the same parameters will generate different keys, potentially preventing applications on Spring Batch 6 from finding existing job executions.

answndud:
## Summary
- restore the default job key serialization format used before the JobParameter refactor
- avoid relying on JobParameter.toString for key generation to keep key values stable
- add a regression test in DefaultJobKeyGeneratorTests for backward-compatible key generation
- add a JDBC DAO regression test to verify lookup of existing job instances created with the legacy key format

## Testing
- ./mvnw -pl spring-batch-core -Dtest=DefaultJobKeyGeneratorTests,JdbcJobInstanceDaoTests test

Fixes #5283

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/DefaultJobKeyGenerator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/DefaultJobKeyGenerator.java
@@ -50,10 +50,15 @@ public class DefaultJobKeyGenerator implements JobKeyGenerator {
 		for (String key : keys) {
 			JobParameter<?> jobParameter = source.getParameter(key);
 			if (jobParameter != null && jobParameter.identifying()) {
-				stringBuffer.append(jobParameter);
+				stringBuffer.append(key).append("=").append(asLegacyParameterValue(jobParameter)).append(";");
 			}
 		}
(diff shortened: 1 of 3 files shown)


answndud:
Closing this since issue #5283 was resolved through migration guide updates rather than a code change. I do not want to keep a code fix open when the maintainer direction is documentation guidance instead.

PR #5323, opened 2026-03-07
Introduce addListener method in StepBuilder and JobBuilder for better clarity

### LINKED ISSUE #5321: Introduce addListener method in StepBuilder and JobBuilder for better clarity
### Discussed in https://github.com/spring-projects/spring-batch/discussions/5194

<sup>Originally posted by **jon890** December 31, 2025</sup>
First of all, I would like to express my sincere gratitude for your hard work in maintaining and evolving the Spring Batch framework.

## Motivation

As a backend engineer heavily using Spring Batch with Java, I've observed that the current `.listener()` method in `StepBuilder` and `JobBuilder` can sometimes be ambiguous.

In the Spring ecosystem and general Java conventions, a method named after a property (without a prefix) often implies setting a single value, potentially overwriting previous ones. However, in Spring Batch, calling `.listener()` multiple times actually **appends** the listeners to an internal list rather than replacing them.

## The Problem
The current naming doesn't explicitly communicate its "additive" behavior:
- Developers might mistakenly think that subsequent calls to `.listener()` will override previous ones.
- For those coming from other frameworks or standard Java Bean conventions, `addListener` is a more intuitive name for cumulative behavior.

## Proposed Change
I suggest adding `addListener()` as an alias or a preferred method for adding listeners in the Fluent DSL.

```java
// Current
public StepBuilder listener(StepExecutionListener listener) { ... }

// Proposed
public StepBuilder addListener(StepExecutionListener listener) {
    return listener(listener); // Delegate to the existing method
}
```

### Benefits
- **Improved Readability**: addListener makes it clear that the listener is being added to a collection.
- **API Consistency**: It aligns with many other modern Java libraries where add* is used for collection-based properties and set* or the property name itself is used for single values.
- **Self-Documenting Code** : It reduces the need for developers to check the internal implementation or Javadoc to confirm if the method is additive.

### Backward Compatibility
To maintain backward compatibility, the existing .listener() method should remain. We could potentially mark it as the secondary option or keep both to support different coding styles (Fluent vs. Explicit).</div>

nikitanagar08:
Fixes #5321

## Motivation

The current `.listener()` method in `StepBuilder` and `JobBuilder` can be ambiguous. A method named after a property (without a prefix) often implies setting a single value, potentially overwriting previous ones. However, in Spring Batch, calling `.listener()` multiple times actually **appends** the listeners to an internal list rather than replacing them.

## Proposed Change

Added `addListener()` methods as aliases that delegate to the existing `listener()` methods. The new method names make the additive behavior explicit and improve API consistency.

## Benefits

- **Improved Readability**: `addListener` makes it clear that the listener is being added to a collection
- **API Consistency**: Aligns with modern Java libraries where `add*` is used for collection-based properties
- **Self-Documenting Code**: Reduces the need for developers to check the internal implementation or Javadoc

## Changes

- `StepBuilderHelper.addListener(StepExecutionListener)` - delegates to `listener()`
- `StepBuilderHelper.addListener(Object)` - delegates to `listener()` for annotation-based config
- `JobBuilderHelper.addListener(JobExecutionListener)` - delegates to `listener()`
- `JobBuilderHelper.addListener(Object)` - delegates to `listener()` for annotation-based config

## Backward Compatibility

The existing `.listener()` methods remain for full backward compatibility.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilderHelper.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/builder/JobBuilderHelper.java
@@ -155,6 +155,28 @@ public abstract class JobBuilderHelper<B extends JobBuilderHelper<B>> {
 	}
 
 	/**
+	 * Add a {@link JobExecutionListener} to the job.
+	 * This method makes the additive nature of adding listeners explicit.
+	 * @param listener the job execution listener to add
+	 * @return this for fluent chaining
+	 * @since 6.1
+	 */
+	public B addListener(JobExecutionListener listener) {
+		return listener(listener);
+	}
+
+	/**
+	 * Add a job execution listener using annotation-based configuration.
+	 * This method makes the additive nature of adding listeners explicit.
+	 * @param listener the object that has a method configured with listener annotation
+	 * @return this for fluent chaining
+	 * @since 6.1
+	 */
+	public B addListener(Object listener) {
+		return listener(listener);
+	}
+
+	/**
 	 * Set a flag to prevent restart an execution of this job even if it has failed.
 	 * @return this to enable fluent chaining
 	 */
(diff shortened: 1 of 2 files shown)


fmbenhassine:
#5321 was closed as not planned: https://github.com/spring-projects/spring-batch/issues/5321#issuecomment-5639249948.

Therefore, I am closing this PR. Thank you for your time anyway!

Issue #5276, opened 2026-02-10
V6 migration script for SQL Server fails


sbklahr:
**Bug description**
I'm migrating from Spring Batch 5.2.4 to 6.02 with an existing SQL server database.

The initial schema was created using [schema-sqlserver.sql in version 5.2.4](https://github.com/spring-projects/spring-batch/blob/e74e9289b6d26c7444667a945c6d7f67b16049d4/spring-batch-core/src/main/resources/org/springframework/batch/core/schema-sqlserver.sql).  When executing [migration-sqlserver.sql in version 6.0.2](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/resources/org/springframework/batch/core/migration/6.0/migration-sqlserver.sql) the following sql error is thrown:

> Failed SQL: (5074) The object "JOB_INST_UN" is dependent on column "JOB_NAME"

when executing `ALTER TABLE BATCH_JOB_INSTANCE ALTER COLUMN JOB_NAME NVARCHAR(100) NOT NULL;` from the migration script.

**Environment**
Spring Batch 6.0.2, SQL Server 15.0.4455

**Steps to reproduce**
Execute the initial schema for [schema-sqlserver.sql in version 5.2.4](https://github.com/spring-projects/spring-batch/blob/e74e9289b6d26c7444667a945c6d7f67b16049d4/spring-batch-core/src/main/resources/org/springframework/batch/core/schema-sqlserver.sql). Then execute [migration-sqlserver.sql in version 6.0.2](https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/resources/org/springframework/batch/core/migration/6.0/migration-sqlserver.sql)

**Expected behavior**
The migration should be executed without errors.

PR #5331, opened 2026-03-11
Bugfix for issue: V6 migration script for SQL Server fails #5276

beatjost:
Dropped and re-added unique constraint on BATCH_JOB_INSTANCE table to optimize performance.


fmbenhassine:
LGTM 👍 Rebased and merged as bcefb68f69c381f8e57e5b272cc0293e438d33bd.

Issue #5245, opened 2026-01-29
Add configuration property to set the jobInstanceIncrementer name


trisberg:
Please do a quick search on Github issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

I would like to have a property, similar to setting the table prefix, for the sequence name used for the jobInstanceIncrementer. This would allow us to share the same schema between Batch 5 and Batch 6 jobs.

**Current Behavior**

The BATCH_JOB_SEQ incrementer was renamed to BATCH_INSTANCE_JOB_SEQ. This makes it impossible for Batch 5 and Batch 6 jobs to share the same schema.

**Context**

This came up when attempting to run Batch 6 jobs in Spring Cloud Data Flow and also with Tanzu Data Flow. We would like to be able to support both Boot 3.x/Batch 5.x and  Boot 4.x/Batch 6.x apps within Data Flow.

quaff:
If no one already work on this, I'd like to pick.

quaff:
Should other incrementer name be configurable?

fmbenhassine:
@quaff thank you for the PR. Yes, this is a good opportunity to make other incrementers configurable as well.

quaff:
> [@quaff](https://github.com/quaff) thank you for the PR. Yes, this is a good opportunity to make other incrementers configurable as well.

I'm hesitating that it will bring unnecessary complexity, cannot imagine why people would like to custom other incrementer's name.

fmbenhassine:
I have seen interest in customizing not only the prefix, but even the entire table name: #4902.

So if we provide a way to customise only the job instance incrementer name, I immediately see "why is it possible to customise the job instance sequence name but not other sequences" coming very quickly. #5258 LGTM, and I don't see any (accidental) complexity coming from it even with two additional setters for other properties.

trisberg:
@fmbenhassine Since the [Boot PR](https://github.com/spring-projects/spring-boot/pull/49401) was closed, can we re-open this issue and see if we can implement the suggested solution of using the Environment?

fmbenhassine:
@trisberg I would not reopen this issue as it added the possibility to configure the sequence name as requested. 

> see if we can implement the suggested solution of using the Environment?

We don't use (and never used) the environment to configure Spring Batch, and I am not keen to introduce that now, especially in a patch release. We need to find another solution. Let's discuss this internally.

PR #5332, opened 2026-03-11
Add support to configure job instance incrementer name through environment variable / system property

trisberg:
In order to be able to run both Batch 5 and 6 jobs we need a way to use the same database schema.

This PR introduces an environment variable `SPRING_BATCH_JDBC_SCHEMA_CLASSIC` in order to override the name for the job instance incrementer used. The classic Batch 5 schema used JOB_SEQ while the Batch 6 schema uses JOB_INSTANCE_SEQ.

See issue: #5245


fmbenhassine:
LGTM. Rebased, squashed and merged as 4580c595673668bdd111fb5986203e7d3e6c9a53. Refined in 80e4349b3028c14299661578f5e375a9e7dcb734 (some minor changes).

I made the exception to introduce this in a patch release. But I will plan a more comprehensive feature to configure Spring Batch through the environment for non Boot users (similar to other projects in the portfolio, for example like Spring Integration, which [has this capability](https://docs.spring.io/spring-integration/reference/configuration/global-properties.html)).

Thank you for your contribution, @trisberg !

PR #5334, opened 2026-03-12
Make JFR optional to support minimized JREs without jdk.jfr module

### LINKED ISSUE #5326: java.lang.ClassNotFoundException: jdk.jfr.Event with minimized JRE
When I use Spring Batch with minimized JRE (without jdk.jfr module) following exception is thrown at `TaskExecutorJobOperator.start` method:

```
<#f190dde2> java.lang.ClassNotFoundException: jdk.jfr.Event
	... 188 more
Wrapped by: <#57495acc> java.lang.NoClassDefFoundError: jdk/jfr/Event
	at org.springframework.batch.core.launch.support.TaskExecutorJobOperator.start(TaskExecutorJobOperator.java:112)
[truncated]
```

Is JFR required for Spring Batch?

nikitanagar08:
Fixes #5326

**Problem:**
TaskExecutorJobOperator.start() creates a JobLaunchEvent which extends jdk.jfr.Event. On minimized JREs without the jdk.jfr module, this causes NoClassDefFoundError: jdk/jfr/Event.

**Solution:**
Wrap the JFR event creation in a try-catch block to handle the case when JFR is not available, allowing Spring Batch to work on minimized JREs without JFR support.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/TaskExecutorJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/TaskExecutorJobOperator.java
@@ -109,7 +109,17 @@ public class TaskExecutorJobOperator extends SimpleJobOperator {
 			JobExecutionAlreadyRunningException, JobRestartException, InvalidJobParametersException {
 		Assert.notNull(job, "Job must not be null");
 		Assert.notNull(jobParameters, "JobParameters must not be null");
-		new JobLaunchEvent(job.getName(), jobParameters.toString()).commit();
+		try {
+			new JobLaunchEvent(job.getName(), jobParameters.toString()).commit();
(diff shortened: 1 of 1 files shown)


nikitanagar08:
This PR fixes #5326 - ClassNotFoundException when using minimized JREs without the jdk.jfr module. The fix makes JFR optional by wrapping JFR event creation in a try-catch block to handle NoClassDefFoundError, allowing Spring Batch to work on minimized JREs without JFR support.

fmbenhassine:
Thank you for the PR. But that's not how we should address the issue.

We should either update the docs with a JDK requirement (if that is the case for SF), or gate all JFR events with a feature flag, or probably introduce a more elaborate interface for observability providers. I will wait for feedback on #5326 to decide which solution to adopt.

PR #5335, opened 2026-03-12
Fix command line parameter parsing to support values containing commas

### LINKED ISSUE #5329: Comma no more supported in command line parameter value since v5
With v4, it was possible to pass parameter values containing a comma on the command line.

That doesn’t seem any more possible with the new command line parameter format.

The comma is interpreted as a separator and so part of the value is considered as a  type, resulting in « Unable to parse job parameter » / ClassNotFoundException.

nikitanagar08:
Fixes #5329

**Problem:**
The parser split on ALL commas, breaking when values contained commas (e.g., "items=apple,banana,orange" would parse "banana" as a type name).

**Solution:**
Added intelligent parsing that checks if a token looks like a class name before treating it as type. If not, all commas are treated as part of the value.

**Tests:**
- Added tests for values with commas
- All existing tests pass (24 tests)

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DefaultJobParametersConverter.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/converter/DefaultJobParametersConverter.java
@@ -175,7 +175,36 @@ public class DefaultJobParametersConverter implements JobParametersConverter {
 		if (tokens.length == 0) {
 			return "";
 		}
-		return tokens[0];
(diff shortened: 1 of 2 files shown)


nikitanagar08:
This PR fixes #5329 - Command line parameter values containing commas are now supported. The fix adds intelligent parsing that distinguishes between commas that are separators vs commas that are part of the value. All 797 core tests pass with backward compatibility maintained.

PR #5351, opened 2026-03-18
Restore JsonLineMapper compatibility with Jackson 2

### LINKED ISSUE #5330: Unable to use JsonLineMapper with Jackson 2
**Bug description**
If an application is using the JsonLineMapper but is stuck with using Jackson 2.x, it will break.

**Environment**
Java 25, Spring Batch 6, H2 database

**Steps to reproduce**
* Create a Java application that has spring-batch-infrastructure as a dependency, typically a spring batch application
* Add a jackson-databind v2 dependency
* Have a usage of the JsonLineMapper in the application
* Execute the application

```
Exception in thread "main" java.lang.NoClassDefFoundError: tools/jackson/databind/json/JsonMapper
	at org.springframework.batch.infrastructure.item.file.mapping.JsonLineMapper.<init>(JsonLineMapper.java:51)
	at LineMapperUser.main(LineMapperUser.java:5)
Caused by: java.lang.ClassNotFoundException: tools.jackson.databind.json.JsonMapper
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:490)
	... 2 more
```

Reproducer project linked hereunder.

**Expected behavior**
As per the [migration guide](https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-6.0-Migration-Guide#jackson-3-support), a migration to Jackson 3 is recommended but should not be mandatory.

Reproducer project: 

[reproducer-json-line-mapper-vs-jackson-2.zip](https://github.com/user-attachments/files/25902595/reproducer-json-line-mapper-vs-jackson-2.zip)

Potentially related issue: https://github.com/spring-projects/spring-batch/issues/1207

**Work around**

There is a workaround by having a custom LineMapper like this:

```java
import com.fasterxml.jackson.databind.ObjectMapper;

public class Jackson2JsonLineMapper implements LineMapper<Map<String, Object>> {
  private final ObjectMapper jsonMapper = new ObjectMapper();
  @Override
  @SuppressWarnings("unchecked")
  public Map<String, Object> mapLine(String line, int lineNumber) throws Exception {
    return this.jsonMapper.readValue(line, Map.class);
  }
}
```

answndud:
Fixes #5330

`JsonLineMapper` currently instantiates Jackson 3's `JsonMapper` by default, which makes the default constructor unusable when only Jackson 2 is present on the classpath.

This change updates `JsonLineMapper` to:
- keep the existing Jackson 3 constructor
- add a Jackson 2 `ObjectMapper` constructor
- fall back to Jackson 2 when Jackson 3 is not available
- add regression tests for both the explicit Jackson 2 constructor and the default-constructor fallback path

Tests:
- `./mvnw -pl spring-batch-infrastructure -Dtest=org.springframework.batch.infrastructure.item.file.mapping.JsonLineMapperTests test`

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/file/mapping/JsonLineMapper.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/file/mapping/JsonLineMapper.java
@@ -15,11 +15,17 @@
  */
 package org.springframework.batch.infrastructure.item.file.mapping;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Map;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import tools.jackson.databind.json.JsonMapper;
 
+import org.jspecify.annotations.Nullable;
 import org.springframework.batch.infrastructure.item.file.LineMapper;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 
 /**
  * Interpret a line as a JSON object and parse it up to a Map. The line should be a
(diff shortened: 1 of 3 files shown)


quaff:
I don't think it's a good idea to mix usage of two version of Jackson, it's better to write your own `JsonLineMapper` if you insist on using Jackson2 for Spring Batch 6.

fmbenhassine:
I think it is not appropriate to make a single class work with two different Jackson versions. To use Jackson 2 with Spring Batch 6, one should use a custom mapper. I will update the migration guide as needed.

Thank you for the PR anyway.

PR #5353, opened 2026-03-20
Fix #5340: Allow restart of jobs with JobParameterIncrementer from command line

### LINKED ISSUE #5340: batch job using param incrementer can’t be restarted anymore from command line
Supersedes spring-projects/spring-boot/issues/49569 (please check spring-boot team feedback on it)

With Spring Boot 4.0.3, batches using `JobParameterIncrementer` can’t be restarted anymore from the command line. Instead of restarting the job instance, İt starts a new instance of the job.

Without `JobParameterIncrementer`, there is no such issue with restart.

Spring Batch 6 doesn’t forbid to restart a job with `JobParameterIncrementer`. Such a job can still be restarted as expected when using `JobOperator#restart` (but not with `JobOperator#start()`).

`JobOperator` seems a little inconsistent to me with Spring Batch 6 major changes around `JobParameterIncrementer`. I don’t know if this restart issue should be fixed on Spring Boot `JobLaunchApplicationRunner` or on Spring Batch side.

Here is a [demo.zip](https://github.com/user-attachments/files/25918290/demo.zip) to illustrate the issue.

nikitanagar08:
This PR addresses #5340 by fixing the restart behavior for jobs using `JobParameterIncrementer` when launched from the command line.

## Problem
Previously, when a job with a `JobParameterIncrementer` was launched from the command line, the `start()` method would always use the incrementer to create a new instance, even when there was a failed/stopped job instance that could be restarted.

## Solution
Modified `SimpleJobOperator.start()` to:
1. First try to run the job with the provided parameters (which restarts failed/stopped instances)
2. Only use the incrementer when `JobInstanceAlreadyCompleteException` is thrown

## Behavior Change
- **Before**: Jobs with incrementers always created new instances from command line
- **After**: Failed/stopped jobs are restarted; only completed jobs trigger the incrementer

This maintains backward compatibility while enabling the expected restart behavior.

---

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
@@ -172,8 +172,10 @@ public class SimpleJobOperator extends TaskExecutorJobLauncher implements JobOpe
 
 	/**
 	 * Start a new instance of a job with the specified parameters. If the job defines a
-	 * {@link JobParametersIncrementer}, then the incrementer will be used to calculate
-	 * the next parameters in the sequence and the provided parameters will be ignored.
+	 * {@link JobParametersIncrementer} and there is no restartable job instance for the
+	 * provided parameters, then the incrementer will be used to calculate the next
+	 * parameters in the sequence. Otherwise, the job will be launched (or restarted) with
+	 * the provided parameters.
 	 * @param job the {@link Job} to start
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Thank you for the PR. As explained in #5340, it is intended to not attempt a restart behind a start operation.

Moreover, this change uses an exception for a normal execution flow (we should not attempt to run the job and then start the next instance if the first attempt throws an exception).

For these reasons, I am closing this PR. Thank you for your time anyway!

PR #5359, opened 2026-03-26
Refine ChunkOrientedStepBuilder to fix generics

### LINKED ISSUE #5358: Generic ItemReader in ChunkOrientedStepBuilder
I was using **SimpleStepBuilder** which gets an` ItemReader<? extends I> reader`. This became deprecated and replaced with **ChunkOrientedStepBuilder** which is not polymorphic (`ItemReader<I> reader`). This sounds like a regression as far as I am concerned. What is your point of vue on this ?

Thanks

Guillaume

quaff:
Closes GH-5358


fmbenhassine:
Merged as 10214165994b549a2901d5331a4a0c00c54fef92. Thank you for your contribution!

PR #5369, opened 2026-04-01
Fix SimpleJobOperator jobRegistry state message

klopfdreh:
Just a small fix until `SimpleJobOperator` is removed


fmbenhassine:
Good catch! thank you for your contribution!

PR #5370, opened 2026-04-04
Fix outdated documentation in Scalability reference page

### LINKED ISSUE #5355: Fix outdated documentation in Scalability reference page
### Summary

There are two documentation issues on the [Scalability reference page](https://docs.spring.io/spring-batch/reference/scalability.html).

---

### 1. Reference to removed `throttle-limit` / `throttleLimit`

The following sentences reference `throttle-limit` (XML) and `throttleLimit` (Java), both of which were removed in Spring Batch 6:

**Java config section:**
> Similar to the multi-threaded step's `throttleLimit` method, the `gridSize` method prevents the task executor from being saturated with requests from a single step.

**XML config section:**
> Similar to the multi-threaded step's `throttle-limit` attribute, the `grid-size` attribute prevents the task executor from being saturated with requests from a single step.

Since both `throttleLimit` and `throttle-limit` no longer exist in Spring Batch 6, these comparisons are misleading and should be updated. A suggested replacement for both:

> The `gridSize` method (or `grid-size` attribute) prevents the task executor from being saturated with requests from a single step.

---

### 2. Broken link and outdated reference to partition job sample

The following sentence contains a broken link and references XML-based configuration that no longer exists:

> The unit test suite for Spring Batch Samples (see `partition*Job.xml` configuration) has a simple example that you can copy and extend.

The linked path (`https://github.com/spring-projects/spring-batch/tree/main/spring-batch-samples/src/main/resources/jobs`) no longer contains partition job XML files.

Also, the placement is inconsistent with other sections — Local Chunking and Remote Step both place the sample link at the end of their respective sections, whereas this reference currently sits in the middle of the Partitioning section, before the PartitionHandler, Partitioner, and Binding Input Data to Steps subsections.

Would it make sense to remove the sentence from its current location and add the following at the end of the Binding Input Data to Steps subsection, consistent with the pattern used in other sections?

> You can find a complete example in the [Partitioning Sample](https://github.com/spring-projects/spring-batch/tree/main/spring-batch-samples/src/main/java/org/springframework/batch/samples/partitioning)

---

### Affected file

`spring-batch-docs/modules/ROOT/pages/scalability.adoc`

---

I'd be happy to submit a pull request for this if the proposed direction looks good to you.

ShinjoWang:
## Summary                                                                                                                                                                 
- Remove references to `throttleLimit` (Java) and `throttle-limit` (XML) in the Partitioning section, which were removed in Spring Batch 6
- Remove broken link to `partition*Job.xml` that no longer exists
- Add Partitioning Sample link at end of "Binding Input Data to Steps" subsection, consistent with Local Chunking and Remote Step sections

Closes #5355

### DIFF
--- a/spring-batch-docs/modules/ROOT/pages/scalability.adoc
+++ b/spring-batch-docs/modules/ROOT/pages/scalability.adoc
@@ -344,9 +344,8 @@ public Step step1Manager(JobRepository jobRepository) {
 }
 ----
 +
-Similar to the multi-threaded step's `throttleLimit` method, the `gridSize`
-method prevents the task executor from being saturated with requests from a single
-step.
+The `gridSize` method prevents the task executor from being saturated with requests
+from a single step.
 
 XML::
 +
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Good catch! Thank you for reporting the issue and contributing a PR!

PR #5379, opened 2026-04-15
Fix skip policy with JPA ItemWriter in scan mode

### LINKED ISSUE #5377: Incorrect handling of skip policy with JPA infrastructure
As of 6.0.3, it is impossible to skip invalid records with a JPA setup. The issue does not happen with a JDBC setup. This was initially discussed in #5303 and seems to be a bug in Spring Batch (to be confirmed). Here is a minimal example:

[demo-gh5303.zip](https://github.com/user-attachments/files/26631042/demo-gh5303.zip)

The sample is expected to insert some items in a table using a `JpaItemWriter`. When all items are valid, they are inserted correctly. If an item violates a database constraint and I try to skip it, the job fails with the following exception:

```
[main] ERROR org.springframework.batch.core.step.AbstractStep -  COMMONS-LOGGING Encountered an error executing step myJobStep in job job
org.springframework.transaction.UnexpectedRollbackException: Transaction silently rolled back because it has been marked as rollback-only
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:803)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:757)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:149)
	at org.springframework.transaction.support.TransactionOperations.executeWithoutResult(TransactionOperations.java:67)
	at org.springframework.batch.core.step.item.ChunkOrientedStep.doExecute(ChunkOrientedStep.java:369)
	at org.springframework.batch.core.step.AbstractStep.execute(AbstractStep.java:251)
	at org.springframework.batch.core.job.SimpleStepHandler.handleStep(SimpleStepHandler.java:128)
	at org.springframework.batch.core.job.AbstractJob.handleStep(AbstractJob.java:397)
	at org.springframework.batch.core.job.SimpleJob.doExecute(SimpleJob.java:129)
	at org.springframework.batch.core.job.AbstractJob.execute(AbstractJob.java:293)
	at org.springframework.batch.core.launch.support.TaskExecutorJobLauncher$1.run(TaskExecutorJobLauncher.java:220)
	at org.springframework.core.task.SyncTaskExecutor.execute(SyncTaskExecutor.java:88)
	at org.springframework.batch.core.launch.support.TaskExecutorJobLauncher.launchJobExecution(TaskExecutorJobLauncher.java:211)
	at org.springframework.batch.core.launch.support.TaskExecutorJobLauncher.run(TaskExecutorJobLauncher.java:109)
	at org.springframework.batch.core.launch.support.SimpleJobOperator.start(SimpleJobOperator.java:201)
	at org.springframework.batch.core.launch.support.TaskExecutorJobOperator.start(TaskExecutorJobOperator.java:117)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:359)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:190)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:158)
	at org.springframework.transaction.interceptor.TransactionInterceptor$1.proceedWithInvocation(TransactionInterceptor.java:133)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:371)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:130)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:222)
	at jdk.proxy2/jdk.proxy2.$Proxy60.start(Unknown Source)
	at com.example.demo.DemoApplication.main(DemoApplication.java:47)
```

In this case, it is expected that the job succeeds and only the invalid item is skipped (this works as expected  with JDBC).

MinChul-Son:
## Problem

Closes #5377

When using a JPA-based `ItemWriter` with a skip policy, the step fails with:

```
Transaction silently rolled back because it has been marked as rollback-only
```

This only happens with JPA; JDBC-based writers work correctly.

### Root Cause

When a chunk write fails and the step enters **scan mode**, `ChunkOrientedStep` was processing all pending items in **a single transaction**. JPA/Hibernate marks the entire transaction as rollback-only after a flush failure. This causes the next item write in the same transaction to throw `"Transaction silently rolled back"`, which is not in the skip list and gets promoted to a `NonSkippableWriteException`, failing the step.

JDBC does not have this problem because a caught SQL exception does not poison the surrounding transaction.

A secondary issue also existed: when `scan()` catches a skippable exception and returns normally, `doExecute()` detects `transactionStatus.isRollbackOnly()` and returns from the lambda without calling `transactionStatus.setRollbackOnly()` explicitly. Since JPA sets the transaction as **globally** rollback-only (via `ConnectionHolder`) but not **locally**, `AbstractPlatformTransactionManager.commit()` calls `processRollback(defStatus, unexpected=true)` and throws `UnexpectedRollbackException`.

### `ChunkOrientedStep`

**One item per scan transaction**

Replaced `ChunkTracker.pendingChunk: Chunk<O>` with `pendingScanItems: LinkedList<O>`. Instead of passing the entire pending chunk to `scan()` at once, the `doExecute()` while loop now dequeues **one item at a time** via `pollNextScanItem()`, so each scan item runs in its own independent transaction. A JPA rollback-only marker from a failed item no longer affects subsequent items.

**Prevent `UnexpectedRollbackException`**

In `doExecute()`, when a rollback-only state is detected, `transactionStatus.setRollbackOnly()` is now called explicitly before returning from the lambda. This converts the global rollback-only (set by JPA) to a local one, so `AbstractPlatformTransactionManager` performs a clean rollback instead of throwing `UnexpectedRollbackException`.

**`beforeChunk` / `afterChunk` consistency (concurrent mode)**

The concurrent scan path was calling `compositeChunkListener.beforeChunk(new Chunk<>())` with an empty chunk. This is now aligned with the sequential path — both pass the actual single-item chunk, and `afterChunk` is guarded by `!status.isRollbackOnly()`.

### Tests (`ChunkOrientedStepScanModeIntegrationTests`)

Four new test cases are added:

| Test | What it verifies |
|---|---|
| `testSkipPolicyWithJpaLikeRollbackOnlyBehaviorInSequentialMode` | Items after a failing item are processed in a fresh transaction (ThreadLocal simulation) |
| `testSkipPolicyWithJpaLikeRollbackOnlyBehaviorInConcurrentMode` | Same, concurrent mode |
| `testUnexpectedRollbackExceptionPreventedInSequentialScanMode` | No `UnexpectedRollbackException` when `ConnectionHolder` is globally rollback-only (direct `ConnectionHolder.setRollbackOnly()` simulation of JPA behavior) |
| `testUnexpectedRollbackExceptionPreventedInConcurrentScanMode` | Same, concurrent mode |

The `globalRollbackOnlyWriter` helper directly calls `ConnectionHolder.setRollbackOnly()` on failure, which is exactly what `JpaTransactionManager`/Hibernate does internally, and is the minimal reproduction of the `UnexpectedRollbackException` path.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkOrientedStep.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkOrientedStep.java
@@ -90,6 +90,7 @@ import static org.springframework.batch.core.observability.BatchMetrics.METRICS_
  * @author Mahmoud Ben Hassine
  * @author Andrey Litvitski
  * @author xeounxzxu
+ * @author Minchul Son
  * @since 6.0
  */
 public class ChunkOrientedStep<I, O> extends AbstractStep {
@@ -375,6 +376,11 @@ public class ChunkOrientedStep<I, O> extends AbstractStep {
 
 				// Skip update during rollback to avoid OptimisticLockingFailureException
 				if (transactionStatus.isRollbackOnly()) {
+					// Explicitly mark as locally rollback-only to prevent
+					// UnexpectedRollbackException when the transaction manager
+					// (eg JpaTransactionManager) has marked it as globally rollback-only
+					// (eg after a JPA flush failure) but not locally rollback-only.
+					transactionStatus.setRollbackOnly();
 					chunkTransactionEvent.transactionStatus = BatchMetrics.STATUS_ROLLED_BACK;
 					chunkTransactionEvent.commit();
 					return;
@@ -408,17 +414,26 @@ public class ChunkOrientedStep<I, O> extends AbstractStep {
 		try {
 			if (tracker.isScanMode()) {
 				logger.info("Executing scan in new transaction after rollback");
-				Chunk<O> pendingChunk = tracker.getPendingChunk();
-				if (pendingChunk != null) {
+				O item = tracker.pollNextScanItem();
+				if (item != null) {
+					Chunk<O> singleItemChunk = new Chunk<>(item);
 					ChunkScanEvent chunkScanEvent = new ChunkScanEvent(stepExecution.getStepName(),
 							stepExecution.getId());
 					chunkScanEvent.begin();
-					scan(pendingChunk, contribution);
+					compositeChunkListener.beforeChunk(singleItemChunk);
+					scan(singleItemChunk, contribution);
+					if (!status.isRollbackOnly()) {
+						compositeChunkListener.afterChunk(singleItemChunk);
+					}
 					chunkScanEvent.skipCount = contribution.getSkipCount();
 					chunkScanEvent.commit();
+				}
+				if (!tracker.hasPendingScanItems()) {
 					logger.info("Chunk scan completed");
 					tracker.exitScanMode();
-					stepExecution.incrementCommitCount();
+					if (!status.isRollbackOnly()) {
+						stepExecution.incrementCommitCount();
+					}
 				}
 				return;
 			}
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Thank you for the PR! LGTM 👍 Rebased and merged as 05b2484cff8a6eb438e6c6859c35e9bceda11efe.

I introduced two changes:

- I noticed that the 3rd modification "Add TransactionStatus parameter to scan()" that you mentioned in https://github.com/spring-projects/spring-batch/issues/5377#issuecomment-4244334449 is not part of this PR, so I amended it.
- The tests added in this PR simulate the behaviour of JPA while being in an integration test class. In an integration test, we typically do not simulate things, but rather setup the entire stack and test against real implementations. For this reason, I omitted those tests and added a full JPA-based integration in 2e9ca2ffd584d391e37906b437aa595b34b47647.

PR #5407, opened 2026-05-28
Remove unused ClassUtils import from JdbcPagingItemReader

enduf7686:
## Summary

Remove unused `ClassUtils` import from `JdbcPagingItemReader`.

## Motivation

The import is no longer referenced and can be safely removed to keep the code clean.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JdbcPagingItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JdbcPagingItemReader.java
@@ -39,7 +39,6 @@ import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
 
 /**
  * <p>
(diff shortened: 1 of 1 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JdbcPagingItemReader.java
```diff
@@ -75,6 +74,7 @@
  * @author Mahmoud Ben Hassine
  * @author Stefano Cordio
  * @author Jimmy Praet
+ * @author Duyeor Kim
```
sudhir-tech: Could we keep this PR scoped to the unused import removal? The author Javadoc change seems unrelated to the stated purpose of this PR. If the author attribution is required, could you clarify the reason for adding it here?

enduf7686: Thanks for the review! You're right—I'll remove the @author tag to keep this PR focused solely on removing unused imports. I'll push an update shortly.

fmbenhassine: The "author" tag can remain. We typically do this for anyone who authored the file in some way, even for small changes.

fmbenhassine:
Rebased and merged as e5cee992daaeee0f733642955e8831ffd2fdccd9.

I dropped the commit that removes the author tag, since you deserve it!

Thank you for your contribution 🙏

PR #5408, opened 2026-06-01
GH-5185 - Move async processing APIs to core

### LINKED ISSUE #5185: Move async processing APIs to spring-batch-core
Hello, I've been studying the new Local Chunking feature added in Spring Batch 6.0, and I'm curious about why this feature is located in spring-batch-integration rather than spring-batch-core.

## Definition of Integration

According to the [[Spring Batch Integration documentation](https://docs.spring.io/spring-batch/reference/spring-batch-integration.html)]

> "Many users of Spring Batch may encounter requirements that are outside the scope of Spring Batch but that may be efficiently and concisely implemented by using Spring Integration."

In other words, the spring-batch-integration module is intended to implement requirements that are outside the scope of Spring Batch by leveraging Spring Integration (external messaging systems).

The key features in this module include:
* **Remote Chunking**: Delegating chunk processing to remote workers via message queues (RabbitMQ, Kafka, etc.)
* **Remote Partitioning**: Executing partitions on external nodes
* **Spring Integration integration**: Message-based Job execution, event handling, etc.

All of these deal with integration with external systems (message brokers, remote nodes).

## Characteristics of Local Chunking 

Local Chunking, introduced in [[#5021](https://github.com/spring-projects/spring-batch/issues/5021)]

> "The idea is to introduce an item writer similar to ChunkMessageChannelItemWriter, but instead of sending chunk requests over the wire to remote workers, it would submit them to local worker threads from a task executor."

Key characteristics:
* Operates within a single JVM
* Uses TaskExecutor (thread pool) to delegate chunks to local worker threads
* No dependency on external systems (message brokers, remote nodes)
* TaskExecutor is a basic Spring Framework component, unrelated to Spring Integration

## Question

Local Chunking:
* Does not require integration with external systems
* Has no dependency on Spring Integration
* Only uses TaskExecutor, which is already used in spring-batch-core for Multi-threaded Step, etc.

**So why is this feature located in spring-batch-integration rather than spring-batch-core?**

The item-level concurrency introduced in [[#4955](https://github.com/spring-projects/spring-batch/issues/4955)](ChunkOrientedStep's `processChunkConcurrently`) is in spring-batch-core, but Local Chunking (chunk-level concurrency), which similarly only uses TaskExecutor, is in spring-batch-integration. This seems inconsistent.

Understanding the background of this design decision would be very helpful for understanding Spring Batch.

Thank you!

won-seoop:
Closes #5185

This moves the async item processing APIs and the local chunking item writer to `spring-batch-core`:

- `AsyncItemProcessor`
- `AsyncItemWriter`
- `ChunkTaskExecutorItemWriter`

The existing `spring-batch-integration` types are kept as deprecated wrappers for source compatibility, and samples/docs now import the core package.

Verification:

- `./mvnw -pl spring-batch-core -Dtest=AsyncItemProcessorTests,AsyncItemWriterTests,ChunkTaskExecutorItemWriterTests test`
- `./mvnw -pl spring-batch-integration,spring-batch-samples -am -DskipTests compile`

Note: on local JDK 21, NullAway JSpecify mode requires `-XDaddTypeAnnotationsToSymbol=true`; the second compile was run with a temporary local Maven compiler configuration change and NullAway downgraded to WARN to get past an existing unrelated `spring-batch-infrastructure` NullAway failure.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/AsyncItemProcessor.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/AsyncItemProcessor.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2006-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.batch.core.step.item;
+
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+
+import org.springframework.batch.core.step.StepExecution;
+
+import org.jspecify.annotations.Nullable;
+import org.springframework.batch.core.listener.ItemProcessListener;
+import org.springframework.batch.core.scope.context.StepContext;
+import org.springframework.batch.core.scope.context.StepSynchronizationManager;
+import org.springframework.batch.infrastructure.item.ItemProcessor;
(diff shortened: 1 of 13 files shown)


won-seoop:
Closing this in favor of a new PR from branch GH-5185 with a DCO-signed commit.

PR #5409, opened 2026-06-01
GH-5185: Move async processing APIs to core

### LINKED ISSUE #5185: Move async processing APIs to spring-batch-core

won-seoop:
Closes #5185

This pull request moves async item processing and local chunking support into `spring-batch-core`:

- Adds `AsyncItemProcessor`, `AsyncItemWriter`, and `ChunkTaskExecutorItemWriter` to `spring-batch-core`
- Keeps the existing `spring-batch-integration` types as deprecated wrappers for source compatibility
- Updates samples and documentation to use the core package
- Adds unit tests for the moved core types

Validation:

- `./mvnw -pl spring-batch-core -Dtest=AsyncItemProcessorTests,AsyncItemWriterTests,ChunkTaskExecutorItemWriterTests test`
- `./mvnw -pl spring-batch-integration,spring-batch-samples -am -DskipTests compile`

Note: local validation was run on JDK 21 with temporary compiler settings for NullAway JSpecify support; those local settings are not part of this pull request.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkTaskExecutorItemWriter.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkTaskExecutorItemWriter.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2025-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.batch.core.step.item;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.batch.core.BatchStatus;
+import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.job.JobInterruptedException;
+import org.springframework.batch.core.listener.StepExecutionListener;
+import org.springframework.batch.core.scope.context.StepSynchronizationManager;
+import org.springframework.batch.core.step.StepContribution;
+import org.springframework.batch.core.step.StepExecution;
+import org.springframework.batch.core.step.skip.NonSkippableReadException;
+import org.springframework.batch.core.step.skip.SkipLimitExceededException;
+import org.springframework.batch.core.step.skip.SkipListenerFailedException;
+import org.springframework.batch.infrastructure.item.Chunk;
+import org.springframework.batch.infrastructure.item.ItemWriter;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.retry.RetryException;
+
+/**
+ * This item writer submits chunk requests to local workers from a {@link TaskExecutor}.
+ *
+ * <p>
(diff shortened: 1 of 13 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/step/item/ChunkTaskExecutorItemWriter.java
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+
+	private void markContributionFailed(StepContribution contribution, Exception exception) {
+		contribution.setExitStatus(ExitStatus.FAILED.addExitDescription(exception));
+	}
+
+	@Override
+	public void beforeStep(StepExecution stepExecution) {
+		this.stepExecution = stepExecution;
```
pkernevez, line 135: I put the comment here as I don't really know where to put it as the class is on transit.
There is an existing bug here or at least an issue with the documentation.
When starting a new step we have to reset the state too:
```suggestion
		this.stepExecution = stepExecution;
        this.responses.clear();
        this.sequence = 0;
```
Without this, we mixe the state of the previous runs.
Or in the [documentation](https://docs.spring.io/spring-batch/reference/scalability.html#localChunking) show an example with a Scope more volatile than a singleton:
```java
@Bean
public ChunkTaskExecutorItemWriter<Vet> itemWriter(ChunkProcessor<Vet> chunkProcessor) {
    ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
    taskExecutor.setCorePoolSize(4);
    taskExecutor.setThreadNamePrefix("worker-thread-");
    taskExecutor.setWaitForTasksToCompleteOnShutdown(true);
    taskExecutor.afterPropertiesSet();
    return new ChunkTaskExecutorItemWriter<>(chunkProcessor, taskExecutor);
}
```

fmbenhassine: @pkernevez That's correct, thank you for raising this. I see that the PR was updated with the suggested fix.

fmbenhassine:
LGTM 👍 Thank you for your contribution! There will be some minor updates that I will add in a refinement commit.

fmbenhassine:
rebased and merged as eafd2e819d0f855fe60bd5e8fa3362f0aebcaa9d. Refined in 826e7d4ea89ca11c2ed77dd81b0a40994d0f0bc6.

PR #5410, opened 2026-06-01
GH-5386: Document transactions in multi-threaded steps

### LINKED ISSUE #5386: Missing documentation about transaction management in multi-threaded steps
**Bug description**
When a ChunkOrientedStep is configured with a TaskExecutor, the item processor is executed outside any transaction. This is a regression from Spring Batch 5.2.5 (where the processor ran inside the chunk transaction) and contradicts the [documentation](https://docs.spring.io/spring-batch/reference/step/chunk-oriented-processing/configuring.html), which states:
```
chunk: The Java-specific name of the dependency that indicates that this is an item-based step and 
the number of items to be processed before the transaction is committed.
```

**Environment**
Migrating from Spring Boot 3.5.13 / Spring Batch 5.2.5 (working) to Spring Boot 4.0.5 / Spring Batch 6.0.3 (broken)
Java 25, PostgreSQL

**Steps to reproduce**
Clone https://github.com/pkernevez/pb-hibernate-proxy/tree/springbatch-transaction-issue (/!\ branch: springbatch-transaction-issue) and run IssueSpringBatchTest.  
The test fails because the processor is annotated with @Transactional(TxType.MANDATORY) and no transaction is active when it runs.  
Commenting out .taskExecutor(executor) in BatchConfiguration makes the test pass.  
To reproduce from scratch: build a ChunkOrientedStep with a chunk size, a TransactionManager, and a TaskExecutor, then annotate the processor with @Transactional(TxType.MANDATORY).
```
        return new StepBuilder("currencyStep", jobRepository)
                .<Long, CurrencyEntity>chunk(10)
                .transactionManager(transactionManager)
//                .taskExecutor(executor)
                .reader(currencyReader)
                .processor(processor)
                .writer(writer)
                .build();
```

**Expected behavior**
* In normal mode: one transaction per chunk, wrapping reader, processor, and writer for all items in that chunk.
* In faultTolerant() mode after a rollback: one transaction per item (scan mode).

**Investigation**
I can't understand the current implementation regarding SpringBatch 5, the documentation and my experience.
This is my understanding without scanMode.

Without a `TaskExecutor`:
1. `doExecute` iterate on the chunk starts a transaction before each
2. For each chunk we process `processNextChunk` that delegates to `processChunkSequentially`
3. We iterate on the item and process them in the transaction started in 1
All is ok.

With a `TaskExecutor`:
1. `doExecute` iterate on the chunk starts a transaction before each in the `main thread`
2. For each chunk we process `processNextChunk` that delegates to `processChunkConcurrently`
3. We iterate on the item and push them in the task executor.
4. They are all executed in other threads without transations  ❌
5. Wait for all the item execution before returning to 1 for the next chunk

Consequences
* Correctness: the processor runs without a transaction and items in a chunk no longer share a Hibernate session — breaking consistency guarantees.
* Performance (caching): reference entities loaded by item N in the Session are not cached in the session for items N+1..N+chunkSize.
* Performance (parallelism): previously, each worker processed a full chunk sequentially in its own transaction; a slow item only stalled one worker.  
Now a slow item blocks the main thread (and the whole batch) until all futures of the current chunk finish, even when the executor pool has spare capacity.

Proposed implementation
1. `doExecute` iterate on the chunk starts a transaction before each in the `main thread`
2. For each chunk we process `processNextChunk` that delegates to `processChunkConcurrently`
3. We prepare the chunk to be executed in another thread, and push them in the task executor.
4. In the `main thread` we directly return to 1 to prepare the next chunk to push in the executor
4bis. In the thread of the task executor, we start a new transaction then iterate on the item of the chunk
5. Wait in 'main-thread` for the completion of all the chunk/Future 

Clone the repo https://github.com/pkernevez/pb-hibernate-proxy/tree/springbatch-transaction-issue
/!\ the right branch is `springbatch-transaction-issue`

won-seoop:
This PR updates the multi-threaded step documentation to describe the transaction boundary of item processor calls in Spring Batch 6.

The new text explains that processor calls run in worker threads and therefore do not participate in the chunk transaction configured on the step, since Spring transactions are thread-bound. It also points users to local chunking when they need worker-thread chunk processing with delegate-managed transaction boundaries.

Closes #5386

Validation:
- `./mvnw -pl spring-batch-docs validate`

Additional check attempted:
- `./mvnw antora -pl spring-batch-docs` failed while Antora invoked `./mvnw javadoc:aggregate`; the failure was an Error Prone plugin internal exception in existing `spring-batch-core` sources (`NestedInstanceOfConditions` in `FaultTolerantStepBuilder.java`), unrelated to this documentation-only change. The same command also fails under JDK 21 earlier due to NullAway requiring JDK 22+ for JSpecify mode.

### DIFF
--- a/spring-batch-docs/modules/ROOT/pages/scalability.adoc
+++ b/spring-batch-docs/modules/ROOT/pages/scalability.adoc
@@ -90,6 +90,15 @@ called from multiple threads at the same time. This means that the `ItemProcesso
 thread-safe. If you are using stateful components in the processing, you must ensure that
 they are properly synchronized for concurrent access.
 
+Since Spring transactions are thread-bound, the item processor calls that run in worker
+threads do not participate in the chunk transaction configured on the step. If an item
+processor needs transaction management, it should start its own transaction, for example
+with a propagation of `REQUIRED`. A processor that requires an existing transaction, for
+example with a propagation of `MANDATORY`, fails unless a transaction is already active in
+the worker thread. If you need an entire chunk to be processed in a worker thread, where
+the delegate can manage a transaction around the chunk, consider using
+<<localChunking,local chunking>> instead.
+
 The reading and writing of items is still done in serial by the main thread executing the step,
 so the `ItemReader` and `ItemWriter` do not have to be thread-safe or synchronized. However,
 the throughput of the step may be limited by the speed of reading and writing. If this is
@@ -608,4 +617,3 @@ public StepLocator stepLocator(BeanFactory beanFactory) {
 ----
 
 You can find a complete example in the https://github.com/spring-projects/spring-batch/tree/main/spring-batch-samples/src/main/java/org/springframework/batch/samples/remotestep[Remote Step Sample].
-


fmbenhassine:
LGTM 👍 Rebased and merged as 73ecb60bbbcc884f006d96469a1763114b14e687.

PR #5414, opened 2026-06-05
Fix incorrect Javadoc reference in JobExecution.isStopped()

leestana01:
The `isStopped()` Javadoc refers to `{@link StepExecution}`, but the method operates on the `JobExecution` instance (`this`). The sibling `isRunning()` and `isStopping()` methods correctly refer to `{@link JobExecution}`. Update `isStopped()` to match.

Javadoc only — no behavior change.


fmbenhassine:
Good catch! thank you for contributing a fix!

PR #5430, opened 2026-06-13
Propagate step exit description from FlowJob to JobExecution

### LINKED ISSUE #3740: FlowJob implementation doesn't set the exception message in JobExecution.ExitStatus.ExitDescription when exception are thrown
Whenever occurs an exception in a standard spring batch program (using simple steps without using flows), the exception message is set in JobExecution.ExitStatus.ExitDescription; however when spring batch work with Flows, the ExitDescription is not set.
As far as I have seen `org.springframework.batch.core.job.flow.FlowJob` implementation doesn't update the JobExecutionStatus with an exit description in `org.springframework.batch.core.job.flow.JobFlowExecutor.updateJobExecutionStatus(FlowExecutionStatus status)`, it just set the exit status, meanwhile `org.springframework.batch.core.job.SimpleJob` implementation does it.

seonwooj0810:
Fixes #3740

## Root cause

When a step inside a `FlowJob` fails, the `StepExecution`'s exit status already contains the failure description (typically the stack trace appended by `AbstractStep.execute()` via `ExitStatus.addExitDescription(ex)`). `SimpleJob` propagates that description to the `JobExecution` directly in `doExecute()`:

```java
execution.setExitStatus(stepExecution.getExitStatus());
```

`FlowJob` does not. Instead, it relies on `JobFlowExecutor.updateJobExecutionStatus(FlowExecutionStatus)`, which only carries the final flow status *name*:

```java
public void updateJobExecutionStatus(FlowExecutionStatus status) {
    execution.setStatus(findBatchStatus(status));
    exitStatus = exitStatus.and(new ExitStatus(status.getName()));
    execution.setExitStatus(exitStatus);
}
```

`executeStep` previously returned only the step's exit *code* and discarded its exit *description*. The executor's tracked `exitStatus` therefore stayed at `ExitStatus.EXECUTING` (description `""`) until it was merged with `new ExitStatus(status.getName())` — also without a description. The result: when a `FlowJob` ends in `FAILED`, the resulting `JobExecution.getExitStatus().getExitDescription()` is empty, even though the failing step has the exception detail. Tooling and listeners that inspect the job-level exit description (e.g. job-level logging, downstream notification jobs) lose the cause.

## Change

In `JobFlowExecutor.executeStep`, accumulate the just-executed step's exit description into the executor's tracked `ExitStatus` via `ExitStatus.addExitDescription`. The subsequent `updateJobExecutionStatus` then merges the description through to the `JobExecution`.

`ExitStatus.addExitDescription` already:
- returns `this` when the new description is blank, so successful steps with empty descriptions cost nothing;
- deduplicates equal descriptions, so the same failure is not duplicated;
- concatenates distinct descriptions with `; `, so multiple failing steps along a flow path are preserved without losing earlier context.

This mirrors `SimpleJob`'s behavior while staying compatible with arbitrary flow shapes — including transitions, decisions, and split flows where multiple step executions may carry descriptions.

## Test evidence

Added `FlowJobFailureTests#testStepFailureExitDescriptionPropagatedToJobExecution`. The test installs a step that sets `ExitStatus.FAILED.addExitDescription("boom: simulated step failure")` and asserts that the resulting `JobExecution.getExitStatus().getExitDescription()` contains that text.

- Without the fix: the assertion fails because the description is empty (`''`).
- With the fix: passes.

All existing tests in `org.springframework.batch.core.job.flow.**` and `org.springframework.batch.core.job.**` continue to pass:

```
mvnw -pl spring-batch-core test \
     -Dtest='org.springframework.batch.core.job.**'
[INFO] Tests run: 150, Failures: 0, Errors: 0, Skipped: 42
[INFO] BUILD SUCCESS
```

(Skipped tests are pre-existing infrastructure-dependent tests.)

## Verification done

1. Confirmed no in-flight PR — `gh pr list --search "FlowJob ExitDescription"` and `gh pr list --search "JobFlowExecutor updateJobExecutionStatus"` both empty.
2. No claim comments in the issue thread (zero comments since 2020-07-08).
3. Fix touches `.java` files only.
4. Verified the bug pattern still exists on current `main` (`d8447d8bc`) by grepping `JobFlowExecutor.updateJobExecutionStatus` and inspecting `executeStep`.
5. Verified the bug behavior by running the new test against unmodified `JobFlowExecutor` (fails with empty description) and against the patched version (passes).
6. Issue thread acknowledges the asymmetry between `SimpleJob.doExecute()` and `JobFlowExecutor.updateJobExecutionStatus()` — the description here matches the analysis the issue author already laid out.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/JobFlowExecutor.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/job/flow/JobFlowExecutor.java
@@ -82,6 +82,12 @@ public class JobFlowExecutor implements FlowExecutor {
 			stepExecution.getExecutionContext().put("batch.restart", true);
 		}
 
+		// Preserve the step's exit description (e.g., the stack trace recorded by
+		// AbstractStep when a step fails) so it survives the flow's final
+		// updateJobExecutionStatus() call, which only carries the FlowExecutionStatus
+		// name and would otherwise discard it.
+		exitStatus = exitStatus.addExitDescription(stepExecution.getExitStatus().getExitDescription());
+
 		return stepExecution.getExitStatus().getExitCode();
 	}
 
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/job/flow/FlowJobFailureTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/job/flow/FlowJobFailureTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2023 the original author or authors.
+ * Copyright 2010-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package org.springframework.batch.core.job.flow;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


seonwooj0810:
Gentle ping — open since mid-June.

`SimpleJob` copies the failing step's exit status onto the `JobExecution`, so the failure description survives; `FlowJob` goes through `JobFlowExecutor.updateJobExecutionStatus(FlowExecutionStatus)`, which only carries the status *name*, so the exit description is lost (#3740). The PR propagates the step's exit description in the flow path too.

Does propagating it from the failing `StepExecution` look right to you, or would you prefer `JobFlowExecutor` carry the description through the `FlowExecutionStatus` instead? Still merges cleanly, checks green.

seonwooj0810:
Merged current `main` to resolve a conflict (only the license header in `FlowJobFailureTests`); CI is green again. This fixes #3740, open since 2020: when a step fails inside a `FlowJob`, the step's exit description is now propagated to the `JobExecution`, matching what `SimpleJob` already does. Happy to adjust if you'd prefer a different approach.

PR #5433, opened 2026-06-20
GH-5432: Fix JsonFileItemWriter append with custom callbacks

### LINKED ISSUE #5432: Fix JsonFileItemWriter to produce valid JSON when append allowed with custom header/footer callback
**Bug description**

When org.springframework.batch.infrastructure.item.json.JsonFileItemWriter is configured with setAppendAllowed(true) and used in multiple runs (open → write → close) against the same target file _with a custom header or footer callback_, the resulting file is not a valid JSON array. So mostly the same as https://github.com/spring-projects/spring-batch/issues/5272 but with custom header and footer callbacks.

**Environment**

Spring Batch 5.2.6, but should be the same for 6.0.4

**Steps to reproduce**

Take the test class from https://github.com/spring-projects/spring-batch/issues/5272 and add the following test case:

```java
    @Test
    void appendAllowedEntriesShouldNotCorruptJsonArray() throws Exception {
        JsonFileItemWriter<String> writer = new JsonFileItemWriter<>(this.resource,
                                                                     new JacksonJsonObjectMarshaller<>());

        writer.setAppendAllowed(true);
        writer.setHeaderCallback(hw -> hw.write("{\"entries\":["));
        writer.setFooterCallback(fw -> fw.write("]}"));

        writer.open(new ExecutionContext());
        writer.close();

        writer.open(new ExecutionContext());
        writer.write(Chunk.of("foo"));
        writer.close();

        writer.open(new ExecutionContext());
        writer.write(Chunk.of("bar"));
        writer.close();
   }
```

This test creates the following invalid `test.json`:

```json
{"entries": "foo",
 "bar"]}
```

**Expected behavior**

I would expect a valid json. Please let me know if this behavior should be supported or not. 

---

To me it looks like the new reopen method is removing this character somehow https://github.com/spring-projects/spring-batch/blob/91a81b6870657210b6eeb7c072958c0a134ed21c/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JsonFileItemWriter.java#L137 . The reopen mechanism is sadly not opt-out if you want to append to an existing file but there seem to be some problems if the file already contains half-finished json (eg in this example because of the header callback but it also applies for other use cases).

DragonFSKY:
Closes #5432.

This PR updates `JsonFileItemWriter#reopen` to keep the array start when append mode reopens an empty file that was created with custom header/footer callbacks. Without that, the truncation step can leave the custom wrapper incomplete before the next append.

The existing root-array behavior is unchanged, and the regression test covers an empty close followed by subsequent appends with a custom wrapper.

Tests:
- `./mvnw -pl spring-batch-infrastructure test`

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JsonFileItemWriter.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/json/JsonFileItemWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2025 the original author or authors.
+ * Copyright 2018-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -138,15 +138,25 @@ public class JsonFileItemWriter<T> extends AbstractFileItemWriter<T> {
 		try (RandomAccessFile raf = new RandomAccessFile(file, "rw")) {
 			long pos = raf.length();
 			boolean stopFound = false;
+			boolean trailingWhitespaceOnly = true;
 			while (--pos >= 0) {
 				raf.seek(pos);
 				int current = raf.readByte();
 				if (!stopFound && current == JSON_ARRAY_STOP) {
 					stopFound = true;
 				}
+				else if (!stopFound && current == JSON_ARRAY_START && trailingWhitespaceOnly) {
+					this.hasExistingItems = false;
+					raf.setLength(pos > 0 ? pos + 1 : pos);
+					break;
+				}
+				else if (!stopFound && !Character.isWhitespace(current)) {
+					trailingWhitespaceOnly = false;
+				}
 				else if (stopFound && !Character.isWhitespace(current)) {
 					this.hasExistingItems = current != JSON_ARRAY_START;
-					raf.setLength(this.hasExistingItems ? pos + 1 : pos);
+					boolean hasContentBeforeArrayStart = current == JSON_ARRAY_START && pos > 0;
+					raf.setLength(this.hasExistingItems || hasContentBeforeArrayStart ? pos + 1 : pos);
 					break;
 				}
 			}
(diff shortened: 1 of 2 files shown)


drexhage:
Hi! Thank you for taking a look at it. What I also feel bad about is if the file already contains other content that also has a pair of `[]`. In this case it might find the json array stop at the array that has nothing to do with it.


This test:
```java
	@Test
	void appendAllowedWithExistingContent() throws Exception {
		JsonFileItemWriter<String> writer = new JsonFileItemWriter<>(this.resource,
				new JacksonJsonObjectMarshaller<>());
		var os = this.resource.getOutputStream();
		os.write("{\"existing\": [1, 2], \"entries\":[".getBytes());
		writer.setAppendAllowed(true);
		writer.setFooterCallback(footerWriter -> footerWriter.write("]}"));

		writer.open(new ExecutionContext());
		writer.close();

		writer.open(new ExecutionContext());
		writer.write(Chunk.of("foo"));
		writer.close();

		System.out.println(this.resource.getContentAsString(StandardCharsets.UTF_8));
	}
```

produces this json:
```json
{"existing": [1, 2,
 "foo"]}
```
completely removing the "entries" field.

I think it should instead stop searching for the json array stop if it already found a unclosed json array start.

DragonFSKY:
Thanks for the detailed reproducer and suggestion. I agree this exposes an edge case in the backward reopen scan.

I’ve updated the PR to stop at the trailing custom entries-array opener before considering an earlier unrelated `]`, and added regression coverage based on your example.

drexhage:
LGTM! Thank you

drexhage (approved):

vauvenal5:
Whats the idea behind:
```java
raf.setLength(this.hasExistingItems || hasContentBeforeArrayStart ? pos + 1 : pos);
```

What is the usecase behind overriding the JSON_ARRAY_START element if it is at position 0?

DragonFSKY:
You’re right—there wasn’t a use case for removing `[` at position 0. I kept that behavior from the original reopen logic, where it caused the default header to be written again, but preserving the existing `[` is sufficient. I’ve simplified both paths to always truncate at `pos + 1`; the existing writer tests still pass. Sorry for the unnecessary condition.

fmbenhassine:
Thank you for the fix and for the follow up!

Rebased and merged as 7979fa08024ffe1287ef2502e139324d727ee51f.

PR #5434, opened 2026-06-21
Fix typo in Javadoc of `JobOperator.start()`

mjd507:
I think it‘s a hasty mistake :)


fmbenhassine:
Good catch! Thank you for contributing a fix!

PR #5440, opened 2026-06-23
GH-5439 Surface the real reason a job execution cannot be restarted

### LINKED ISSUE #5439: SimpleJobOperator.restart(JobExecution) masks the real failure reason with a misleading "job execution already running" message
### Summary

`SimpleJobOperator.restart(JobExecution)` wraps every exception thrown by `run(...)` into a `JobRestartException` whose message is hardcoded to "job execution already running", discarding the real, accurate reason the job could not be restarted.

### Details

`TaskExecutorJobLauncher.createJobExecution(...)` already throws specific, accurate exceptions for non-restartable conditions. For example, when the previous execution (or one of its step executions) ended in `UNKNOWN` status:

```java
  throw new JobRestartException("Cannot restart job from UNKNOWN status. "
      + "The last execution ended with a failure that could not be rolled back, "
      + "so it may be dangerous to proceed. Manual intervention is probably necessary.");
```

But SimpleJobOperator.restart(JobExecution) catches the broad Exception and rebuilds a new message:

```java
  try {
      return run(job, parameters);
  }
  catch (Exception e) {
      throw new JobRestartException(
          String.format(ILLEGAL_STATE_MSG, "job execution already running", jobName, parameters), e);
  }
```

So the accurate message is demoted to the cause, and the top-level message wrongly says "job execution already running" — even though the job is not running (it's UNKNOWN, or already complete, or had invalid parameters).

## Impact

Users and logs see "job execution already running" when the real problem is an UNKNOWN-status execution requiring manual intervention — sending people down the wrong debugging path. The same masking also mislabels JobInstanceAlreadyCompleteException and InvalidJobParametersException.

## Expected

restart(JobExecution) should surface the real reason: propagate the launcher's JobRestartException as-is, and only use the "already running" message for the genuinely running case (JobExecutionAlreadyRunningException).

## How to reproduce

Restart a JobExecution whose last execution ended in UNKNOWN status (e.g. when a step's final metadata save fails under a concurrent stop). The resulting JobRestartException reports "job execution already running" instead of the UNKNOWN-status reason.

## Suggested fix

Replace the broad catch (Exception e) with per-type handling: wrap only JobExecutionAlreadyRunningException as "already running", rethrow JobRestartException unchanged, and wrap JobInstanceAlreadyCompleteException / InvalidJobParametersException with their own messages.

kyungrae:
Resolves #5439

## Problem

`SimpleJobOperator.restart(JobExecution)` wraps *every* exception thrown by `run(...)` into a `JobRestartException` with a hardcoded `"job execution already running"` message:

  ```java
  try {
      return run(job, parameters);
  }
  catch (Exception e) {
      throw new JobRestartException(
          String.format(ILLEGAL_STATE_MSG, "job execution already running", jobName, parameters), e);
  }
```

The launcher (TaskExecutorJobLauncher.createJobExecution) already throws accurate, specific exceptions for the various non-restartable conditions. For example, when the last execution (or one of its step executions) ended in UNKNOWN status:

> Cannot restart job from UNKNOWN status. The last execution ended with a failure that could not be rolled back, so it may be dangerous to proceed. Manual intervention is probably necessary.

Because the operator catches the broad Exception and rebuilds the message, that accurate reason is demoted to the cause and the top-level message wrongly states the job is "already running" — even though it is not (it is UNKNOWN, already complete, or had invalid parameters). Users and logs are then sent down the wrong debugging path.

Note: although SimpleJobOperator is deprecated in favour of TaskExecutorJobOperator, the latter inherits this method via super.restart(...), so the active operator is affected.

## Change

Replace the broad catch (Exception e) with per-type handling:

- JobExecutionAlreadyRunningException → keep the "job execution already running" message(the only genuinely-running case);
- JobInstanceAlreadyCompleteException / InvalidJobParametersException → wrap while preserving their own message;
- JobRestartException → propagate unchanged so its specific reason (e.g. the UNKNOWN-status message above) survives.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
@@ -245,9 +245,12 @@ public class SimpleJobOperator extends TaskExecutorJobLauncher implements JobOpe
 		try {
 			return run(job, parameters);
 		}
-		catch (Exception e) {
+		catch (JobExecutionAlreadyRunningException e) {
 			throw new JobRestartException(
 					String.format(ILLEGAL_STATE_MSG, "job execution already running", jobName, parameters), e);
+		}
+		catch (JobInstanceAlreadyCompleteException | InvalidJobParametersException e) {
+			throw new JobRestartException(e.getMessage(), e);
 		}
 
 	}
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/launch/support/TaskExecutorJobOperatorTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/launch/support/TaskExecutorJobOperatorTests.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.batch.core.launch.support;
 
+import java.time.LocalDateTime;
+
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased, squashed and merged as 3c875a47eec3cb18c755e681bec7487d33f2fef2.

PR #5441, opened 2026-06-23
GH-5435: Add MongoDB runtime hints

### LINKED ISSUE #5435: Missing GraalVM runtime hints for MongoDB persistence model classes and resources
hi graalvm native image compilation fails in a new boot 4.1 + mognodb configuration 

if I create a `RuntimeHintsRegistrar` as follows it seems to work 

```
    static class Hints implements RuntimeHintsRegistrar {

        @Override
        public void registerHints(@NonNull RuntimeHints hints, @Nullable ClassLoader classLoader) {
            for (var c : new Class[]{
                    org.springframework.batch.core.repository.persistence.JobInstance.class,
                    org.springframework.batch.core.repository.persistence.ExecutionContext.class,
                    org.springframework.batch.core.repository.persistence.ExitStatus.class,
                    org.springframework.batch.core.repository.persistence.StepExecution.class,
                    org.springframework.batch.core.repository.persistence.JobExecution.class,
                    org.springframework.batch.core.repository.persistence.JobParameter.class,
            }) {
                hints.reflection().registerType(c, MemberCategory.values());
            }

            var prefix = "org/springframework/batch/core/";
            for (var r : new String[]{
                    "schema-mongodb", //
                    "schema-drop-mongodb"}) {
                for (var suffix : "jsonl,js".split(",")) {
                    var path = prefix + r + "." + suffix;
                    var resource = new ClassPathResource(path);
                    if (resource.exists()) {
                        hints.resources().registerResource(resource);
                    }
                }
            }
        }
    }
```

DragonFSKY:
Closes #5435.

Register the four MongoDB schema resources used by Spring Batch and the MongoDB persistence model types needed for Spring Data binding. The reflection hints use Spring Framework's `BindingReflectionHintsRegistrar`, limiting access to the binding contract instead of enabling every member category.

One focused test verifies the resource hints and required field/constructor access for all six persistence types.

Tests:
- `./mvnw --no-transfer-progress -pl spring-batch-core -Dtest=CoreRuntimeHintsTests test`
- `./mvnw --no-transfer-progress --quiet -pl spring-batch-core test` (819 tests, 0 failures/errors)
- `git diff --check`

I did not run a native-image build locally because the issue does not include a reproducer.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/aot/CoreRuntimeHints.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/aot/CoreRuntimeHints.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2025 the original author or authors.
+ * Copyright 2022-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -98,6 +98,13 @@ public class CoreRuntimeHints implements RuntimeHintsRegistrar {
 				"java.util.concurrent.locks.ReentrantLock$FairSync",
 				"java.util.concurrent.locks.ReentrantLock$NonfairSync",
 				"java.util.concurrent.ConcurrentHashMap$Segment");
+		Set<Class<?>> mongoDbPersistenceTypes = Set.of(
+				org.springframework.batch.core.repository.persistence.ExecutionContext.class,
+				org.springframework.batch.core.repository.persistence.ExitStatus.class,
+				org.springframework.batch.core.repository.persistence.JobExecution.class,
+				org.springframework.batch.core.repository.persistence.JobInstance.class,
+				org.springframework.batch.core.repository.persistence.JobParameter.class,
+				org.springframework.batch.core.repository.persistence.StepExecution.class);
 
 		// resource hints
 		hints.resources().registerPattern("org/springframework/batch/core/schema-h2.sql");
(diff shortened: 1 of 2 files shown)


spring-batch-core/src/main/java/org/springframework/batch/core/aot/CoreRuntimeHints.java
```diff
@@ -153,6 +158,13 @@ public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
 		jdkTypes.stream()
 			.map(TypeReference::of)
 			.forEach(type -> hints.reflection().registerType(type, MemberCategory.values()));
+		new BindingReflectionHintsRegistrar().registerReflectionHints(hints.reflection(),
```
fmbenhassine, line 161: There is no need to use a new `BindingReflectionHintsRegistrar` here, we can use the `hints` reference already in scope. I will amend that on merge.

spring-batch-core/src/test/java/org/springframework/batch/core/aot/CoreRuntimeHintsTests.java
```diff
@@ -0,0 +1,57 @@
...
+					"org/springframework/batch/core/schema-mongodb.js",
+					"org/springframework/batch/core/schema-drop-mongodb.jsonl",
+					"org/springframework/batch/core/schema-drop-mongodb.js")
+			.forEach(path -> assertThat(runtimeHints).matches(resource().forResource(path)));
+		Stream
+			.of(JobInstance.class, ExecutionContext.class, ExitStatus.class, StepExecution.class, JobExecution.class,
+					JobParameter.class)
+			.forEach(type -> assertThat(runtimeHints).matches(reflection().onType(type)
+				.withMemberCategories(MemberCategory.ACCESS_DECLARED_FIELDS,
+						MemberCategory.INVOKE_DECLARED_CONSTRUCTORS)));
```
fmbenhassine, line 54: In `CoreRuntimeHints`, hints for those type are registered for all member categories (`MemberCategory.values()`), so the assertion should include all `MemberCategory` values as well.

I will amend that on merge too.

fmbenhassine:
Rebased and merged as 2f7c88c060eb0902402c41f915f3505263cc0f93.

Issue #5360, opened 2026-03-27
Performance in SimpleJobRepository.update(StepExecution) since Spring Batch 6.x


leechoongyon:
Please do a quick search on Github issues first, there might be already a duplicate issue for the one you are about to create.
  If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

  **Bug description**

  `SimpleJobRepository.update(StepExecution)` was changed in Spring Batch 6.x to call `getStepExecution(stepExecution.getId())`, which triggers a query against `BATCH_JOB_EXECUTION_PARAMS` on every chunk commit. On databases such as Oracle, which do not automatically create an index for foreign key constraints, `JOB_EXECUTION_ID` has no index on `BATCH_JOB_EXECUTION_PARAMS`, resulting in a full table scan on every chunk commit. As historical job execution data accumulates, this causes a severe and growing performance regression.

 For reference, in Spring Batch 5.x, `update(StepExecution)` called `checkForInterruption()` → `synchronizeStatus()`, which issued only a single lightweight query:

  ```sql
  SELECT VERSION FROM BATCH_JOB_EXECUTION WHERE JOB_EXECUTION_ID = ?
  ```

  In Spring Batch 6.x, `getStepExecution(stepExecution.getId())` triggers the following **8 queries on every chunk commit**:

  | # | Table | Description |
  |---|-------|-------------|
  | 1 | `BATCH_JOB_EXECUTION` + `BATCH_STEP_EXECUTION` | fetch `JOB_EXECUTION_ID` |
  | 2 | `BATCH_JOB_INSTANCE` + `BATCH_JOB_EXECUTION` | fetch `JOB_INSTANCE_ID` |
  | 3 | `BATCH_JOB_INSTANCE` | fetch job instance |
  | 4 | `BATCH_JOB_EXECUTION_PARAMS` | fetch job parameters ⚠️ **full table scan on Oracle** |
  | 5 | `BATCH_JOB_EXECUTION` | fetch job execution |
  | 6 | `BATCH_STEP_EXECUTION` | fetch step execution |
  | 7 | `BATCH_STEP_EXECUTION_CONTEXT` | fetch step execution context |
  | 8 | `BATCH_JOB_EXECUTION_CONTEXT` | fetch job execution context |

[...]

Issue #5120, opened 2025-12-01
StepExecution Update in SimpleJobOperator.stop() Causes JobExecution.BatchStatus.UNKNOWN after graceful stop


KILL9-NO-MERCY:
Hello Spring Batch Team,

I am reporting an issue where using JobOperator.stop() to gracefully stop a running ChunkOrientedStep results in an OptimisticLockingFailureException and setting UNKNOWN state

## Bug description 
In Spring Batch version 6.0.0, calling SimpleJobOperator.stop(jobExecution) on an executing ChunkOrientedStep causes an Optimistic Locking version conflict.

This happens because the SimpleJobOperator.stop() method, after calling stoppableStep.stop() at line #374 proceeds to explicitly call jobRepository.update(stepExecution). 

This update prematurely increments the database version of the StepExecution. 

Consequently, the main batch execution thread, which holds an outdated version of the StepExecution in memory, fails with an OptimisticLockingFailureException during its final persistence call in AbstractStep.execute().

## Environment
Spring Batch Version: 6.0.0
Spring Boot 4.0.0

## Steps to reproduce
1) Start a Spring Batch application with a long-running ChunkOrientedStep.
2) While the step is actively processing a chunk (inside the chunk transaction), call JobOperator.stop(jobExecution) from a separate thread or API endpoint.
3) The SimpleJobOperator.stop() call updates the DB, increasing the StepExecution version.(at line 375)
~4) The batch execution thread(Chunk processing thread) detects the terminateOnly flag and attempts a graceful exit from the chunk processing loop. (at ChunkOrientedStep.doExecute() line 362)~ ChunkOrientedStep doExecute() completed(not stopped - this is related https://github.com/spring-projects/spring-batch/issues/5114) 
5) The AbstractStep.execute() method attempts to save the final status of the step.(at line 327)
6) The job fails with an OptimisticLockingFailureException. and JobExecution.BatchStatus & ExitStatus set UNKNOWN 
7) so this JobExecution cannot restarted

## Expected behavior 
When JobOperator.stop() is called, the job should safely stop and transition to the STOPPED status without causing an OptimisticLockingFailureException or setting UNKNOWN status for restartability

## Actual Stack Trace
```java
org.springframework.dao.OptimisticLockingFailureException: Attempt to update step execution id=9 with wrong version (1), where current version is 2
[...]

Issue #5114, opened 2025-11-27
stop() does not prevent upcoming steps to be executed anymore


andre-bugay:
It seems like Spring Batch 6 cannot stop a Job anymore.
After calling stop(), all steps are executed and later the job is marked as FAILED.

In Spring Batch 5 the flow was:
`STARTED` -> `STOPPING` -> mark step executions as terminateOnly -> `STOPPED`

In Spring Batch 6 it is:
`STARTED` -> `STOPPING` -> `STOPPED` -> `FAILED`

If I am right then the root cause of this change is the following new line in 

https://github.com/spring-projects/spring-batch/blob/c8a0528bf1ee3ff8015ae1ddaaef368355f32ed3/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java#L348

Directly afterwards the 
`jobRepository.update(jobExecution);`
checks
https://github.com/spring-projects/spring-batch/blob/c8a0528bf1ee3ff8015ae1ddaaef368355f32ed3/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java#L139

This will always be false as the endTime was set just before.
The jobState will be set from `STOPPING` to `STOPPED` directly.

**Consequence**
Inside `SimpleJobRepository#update(StepExecution)` -> `checkForInterruption(stepExecution)` the check in
https://github.com/spring-projects/spring-batch/blob/c8a0528bf1ee3ff8015ae1ddaaef368355f32ed3/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java#L186-L188
will never be true and the steps are not marked for terminateOnly.

Is this intended and how can I prevent running the unstarted steps once the job is stopped?

KILL9-NO-MERCY:
Hi, I also have reviewed the root cause of this issue. I'd like to share my findings for your reference.

As you pointed out, the commit e5fbc2a introduced:
```java
jobExecution.setEndTime(LocalDateTime.now());
```

This change causes the following logic within SimpleJobRepository to execute:
```java
if (jobExecution.getStatus() == BatchStatus.STOPPING && jobExecution.getEndTime() != null) {
    if (logger.isInfoEnabled()) {
       logger.info("Upgrading job execution status from STOPPING to STOPPED since it has already ended.");
    }
    jobExecution.upgradeStatus(BatchStatus.STOPPED);
}
```
[...]

Issue #4023, opened 2021-10-31
Spring batch terminate in started status after sigterm


esoni:
Hi All,
i have a job that run on k8s. During a shutdown for a cluster update k8s send SIGTERM to my container.
My goal is to 
- handle sigterm in a function
- in this function i want to call stop job and wait the termination
I m able to add a shutdown hook using 
`Runtime.getRuntime().addShutdownHook(new Thread() {`
but in the meanwhile spring batch automatically terminate when i receive SIGTERM and the job remains in STARTED status on the db.
How can customize this behavior ? i want to stop correctly the job when i receive sigterm

Jacopo47:
Hi everyone, 
any update on this?

We're experiencing the same situation and we're evaluating how to tackle this. 
But we're wondering if this functionality should be handled by the third part or if should be covered by the batch framework.
Because, as shared by @esoni, this bug leads to inconsistencies since job status on DB is STARTED even if no job's execution is running on services

fmbenhassine:
> During a shutdown for a cluster update k8s send SIGTERM to my container.

Are you using Docker or another container runtime? If using Docker, are you using the `shell` entry point form? I am asking because the shell form of docker's ENTRYPOINT [does not send Unix signals to the sub-process running in the container](https://docs.docker.com/engine/reference/builder/#entrypoint):

```
The shell form prevents any CMD or run command line arguments from being used, but has the disadvantage
that your ENTRYPOINT will be started as a subcommand of /bin/sh -c, which does not pass signals. 

This means that the executable will not be the container’s PID 1 - and will not receive Unix signals
- so your executable will not receive a SIGTERM from docker stop <container>.
```

So in order to correctly intercept Unix signals by the Spring Batch job running in a container, the ENTRYPOINT form should be `exec`. To my knowledge, the only way offered by the JVM to intercept external signals is shutdown hooks. So if we really want to to stop a job when receiving a SIGTERM, we should add a shutdown hook that calls `JobOperator.stop`. However, this approach is not guaranteed to work because shutdown hooks are not guaranteed to be called by the JVM. Here is an excerpt from the Javadoc of `Runtime.addShutdownHook` method:

```
In rare circumstances the virtual machine may abort, that is, stop running without shutting down cleanly.
This occurs when the virtual machine is terminated externally, for example with the SIGKILL signal on
Unix or the TerminateProcess call on Microsoft Windows.
```

Moreover, shutdown hooks are expected to run "quickly":

```
[...]

PR #5448, opened 2026-06-29
GH-5308: Serialize step execution updates with concurrent stop requests

### LINKED ISSUE #5308: GracefulShutdownFunctionalTests.testStopJob fails intermittently due to a race condition
The resolution of #5217 has reduced the risk of race conditions on graceful shutdown, but does not seem to remove it completely (as the test is still failing intermittently).

This issue to investigate further.

kyungrae:
## Problem

When `JobOperator.stop(jobExecution)` is called while a step is running, the stopping thread persists the step execution's stopped state (`jobRepository.update(stepExecution)`) on its own thread — concurrently with the worker thread still committing chunks for the same BATCH_STEP_EXECUTION row. Both issue optimistic-locking updates:

```SQL
UPDATE BATCH_STEP_EXECUTION SET ..., VERSION = ? WHERE STEP_EXECUTION_ID = ? AND VERSION = ?
```

so one matches 0 rows and fails with `OptimisticLockingFailureException`. It is timing- and vendor-sensitive: frequent on MySQL (`REPEATABLE READ`), occasional on PostgreSQL (`READ COMMITTED`), almost never on in-memory HSQLDB — which is why CI rarely catches it and `GracefulShutdownFunctionalTests` was `@Disabled`.

A contributing factor: SimpleJobRepository.update(StepExecution) re-read the row version via stepExecutionDao.synchronizeStatus(stepExecution) while the job was stopping. Under MySQL REPEATABLE READ that read returns the stale snapshot version, overwriting the correct in-memory version and guaranteeing the stopping thread's update loses.

## Solution

Guard every update to a step execution's metadata with a **per-execution lock**, held across the surrounding transaction's commit, shared between the worker and the stopping thread:

- `AbstractStep` keeps one `Semaphore` per running step execution and exposes `StoppableStep.callUnderLock(StepExecution, Runnable)`. The worker's start/chunk/final updates and the operator's stop update all run under it, so they serialize and neither observes a stale version.
- `TaskletStep` and `ChunkOrientedStep` take this shared lock around their chunk transactions.
- `SimpleJobRepository.update(StepExecution)` no longer re-reads the version while stopping — under the lock the shared in-memory execution already holds the current version (the stale re-read was the root failure on MySQL).
- The operator sets the job execution status to `STOPPED` directly instead of `STOPPING` (which update(JobExecution) upgraded to STOPPED anyway).
- Re-enables `GracefulShutdownFunctionalTests` (disabled under #5308).

## Validation

JobOperatorFunctionalTests and GracefulShutdownFunctionalTests, 100 runs per vendor (MySQL 8 / PostgreSQL 16 in Docker, HSQLDB in-memory), after the fix:

| Test | Vendor | Baseline (no fix) | After this PR |
|---|---|---|---|
| `GracefulShutdownFunctionalTests` | HSQLDB | 3/100 | **0/100** |
| `GracefulShutdownFunctionalTests` | MySQL | 27/100 | **0/100** |
| `GracefulShutdownFunctionalTests` | PostgreSQL | 3/100 | **0/100** |
| `JobOperatorFunctionalTests` | HSQLDB | 1/100 | **0/100** |
| `JobOperatorFunctionalTests` | MySQL | 22/100 | **0/100** |
| `JobOperatorFunctionalTests` | PostgreSQL | 8/100 | **0/100** |

(0 failures / 600 runs. Before-fix numbers: see #5442.) Plus a new AbstractStepTests unit test asserting t updates to one step execution; full spring-batch-core suite passes.

Resolves #5308

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/JobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/JobOperator.java
@@ -187,12 +187,12 @@ public interface JobOperator extends JobLauncher {
 	boolean stop(long executionId) throws NoSuchJobExecutionException, JobExecutionNotRunningException;
 
 	/**
-	 * Send a stop signal to the supplied {@link JobExecution}. The signal is successfully
-	 * sent if this method returns true, but that doesn't mean that the job has stopped.
-	 * The only way to be sure of that is to poll the job execution status.
+	 * Stop the supplied {@link JobExecution}. The job execution is marked as STOPPED and
+	 * the currently running step is signalled to stop at the next chunk boundary.
+	 * Returning {@code true} means the stop was recorded; the worker thread may still be
+	 * winding down its current chunk when this method returns.
 	 * @param jobExecution the running {@link JobExecution}
-	 * @return true if the message was successfully sent (does not guarantee that the job
-	 * has stopped)
+	 * @return true if the stop was successfully recorded
 	 * @throws JobExecutionNotRunningException if the supplied {@link JobExecution} is not
 	 * running (so cannot be stopped)
 	 */
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java
@@ -331,9 +331,8 @@ public class SimpleJobOperator extends TaskExecutorJobLauncher implements JobOpe
 	@Override
 	public boolean stop(JobExecution jobExecution) throws JobExecutionNotRunningException {
 		Assert.notNull(jobExecution, "JobExecution must not be null");
-		// Indicate the execution should be stopped by setting it's status to
-		// 'STOPPING'. It is assumed that
-		// the step implementation will check this status at chunk boundaries.
+		// Indicate the execution should be stopped. The step implementation is
+		// expected to check this status at chunk boundaries.
 		BatchStatus status = jobExecution.getStatus();
 		if (!(status == BatchStatus.STARTED || status == BatchStatus.STARTING)) {
 			throw new JobExecutionNotRunningException(
@@ -342,8 +341,7 @@ public class SimpleJobOperator extends TaskExecutorJobLauncher implements JobOpe
 		if (logger.isInfoEnabled()) {
 			logger.info("Stopping job execution: " + jobExecution);
 		}
-		jobExecution.setStatus(BatchStatus.STOPPING); // will be upgraded to STOPPED in
-														// JobRepository.update
+		jobExecution.setStatus(BatchStatus.STOPPED);
 		jobExecution.setExitStatus(ExitStatus.STOPPED);
 		jobExecution.setEndTime(LocalDateTime.now());
 		jobRepository.update(jobExecution);
@@ -364,16 +362,20 @@ public class SimpleJobOperator extends TaskExecutorJobLauncher implements JobOpe
 								if (tasklet instanceof StoppableTasklet stoppableTasklet) {
 									StepSynchronizationManager.register(stepExecution);
 									stoppableTasklet.stop(stepExecution);
-									jobRepository.update(stepExecution);
-									jobRepository.updateExecutionContext(stepExecution);
+									taskletStep.callUnderLock(stepExecution, () -> {
+										jobRepository.update(stepExecution);
+										jobRepository.updateExecutionContext(stepExecution);
+									});
 									StepSynchronizationManager.release();
 								}
 							}
 							if (step instanceof StoppableStep stoppableStep) {
 								StepSynchronizationManager.register(stepExecution);
 								stoppableStep.stop(stepExecution);
-								jobRepository.update(stepExecution);
-								jobRepository.updateExecutionContext(stepExecution);
+								stoppableStep.callUnderLock(stepExecution, () -> {
+									jobRepository.update(stepExecution);
+									jobRepository.updateExecutionContext(stepExecution);
+								});
 								StepSynchronizationManager.release();
 							}
 						}
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/support/SimpleJobRepository.java
@@ -164,7 +164,6 @@ public class SimpleJobRepository extends SimpleJobExplorer implements JobReposit
 		this.jobExecutionDao.synchronizeStatus(jobExecution);
 
 		if (jobExecution.isStopped() || jobExecution.isStopping()) {
-			this.stepExecutionDao.synchronizeStatus(stepExecution);
 			stepExecution.setTerminateOnly();
 		}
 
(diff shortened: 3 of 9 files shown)


nikhiln64 (changes requested):
The per execution semaphore moved onto AbstractStep and shared with the stopping thread is a clean way to keep a chunk commit from racing the stop update inside one JVM, and the new AbstractStepTests case shows callUnderLock serialising concurrent callers on the same execution. My concern is the change that rides along with it in SimpleJobRepository.update, where stepExecutionDao.synchronizeStatus(stepExecution) is removed from the isStopped or isStopping branch.

That call was doing more than status reconciliation. It reloads VERSION from the row and, when the in memory version is stale, resets stepExecution version to the database value, which is exactly what keeps the next optimistic locking update from throwing OptimisticLockingFailureException when two holders of the same StepExecution both write. The new lock replaces that protection only for writers in the same JVM, since callUnderLock runs the action unlocked when getStepExecutionLock returns null and the comment there names the case, not executing in this JVM. So for a remote partitioned or remote chunking step, the worker updates the row in its own JVM while the operator on the manager stops the same execution through the null lock path with synchronizeStatus now gone, and the reconciliation that used to absorb that race is no longer there.

The added test exercises callUnderLock within a single execute() on one JVM, so it stays green even if the remote path regresses. Could you add a case that stops a step whose execution is not registered in the stopping thread's map, standing in for the remote worker, and asserts the stop update still succeeds against a concurrently bumped version. If that path does still need synchronizeStatus, keeping it only for the null lock branch would restore the old behaviour without giving up the new in JVM lock.

Two smaller things. The semaphore is put into the map at the top of execute() and removed in its finally, so a stop arriving before execute() populates the map, or after it clears, takes the unlocked branch, and for a stop that lands in that window against a still running worker the serialisation is silently skipped. And setting the execution straight to STOPPED instead of STOPPING drops the STOPPING observation entirely, so anything polling for the intermediate state, or restart logic that distinguishes the two, now only ever sees STOPPED. Both may be intended but are worth a line in the description.

kyungrae:
Thank you for the detailed review — it pushed me to study how `PartitionStep` and `RemoteStep` actually work, and I would like to share what I found before changing the PR.

### 1. On restoring `stepExecutionDao.synchronizeStatus(stepExecution)`

I don't think restoring it is the right fix, even when scoped to the unlocked branch.

The `VERSION` column exists to detect that another party has written the row, so that an update built on an outdated snapshot is *rejected* rather than applied. `OptimisticLockingFailureException` is that detection doing its job — it is the signal, not the defect.

`synchronizeStatus` re-reads the version and adopts it, so the update that follows is accepted even though the in-memory `StepExecution` was taken before the external write. The row is then written from that outdated snapshot. This does not resolve the concurrent modification — it only removes our ability to notice it.

If another party really has updated the row, what we *should* do depends on the state the step execution is in and on who performed that write, and I don't believe there is a generally safe rule for that. Defining such a rule is the genuinely hard part of this problem. Re-synchronizing the version silently replaces that decision with "write anyway", which I consider more dangerous than the exception it avoids.

So my preference is to eliminate the concurrent write rather than suppress its detection. (Section 2 below covers why I believe the remote step implementations do not have a concurrent writer on those rows in the first place.)

Some history on where that call comes from:

- e5fbc2a0387858f5f95009e3a032d2864398f9ac (#4023) made `stop()` mark the job and step executions `STOPPED` with an end time before updating the repository, which introduced a second writer on the step execution row.
- 644d7e6997c4e29822be580dab8e6f65713e17be (#5120, #5114) then added version synchronization in `SimpleJobRepository.update(StepExecution)` to suppress the resulting `OptimisticLockingFailureException`.
- 692acba437daa7459dc6b55b2c42714a8536ae6f (GH-5360) later replaced that full re-read with the cheaper `stepExecutionDao.synchronizeStatus(...)`.

So the line this PR removes was introduced as a mitigation for the lock contention Once the contention itself is removed, removing it should not change behaviour.

### 2. On remote partitioning and remote chunking

You are right that a `Semaphore` cannot cross JVM boundaries, and this PR does not address cross-JVM concurrency. Having studied both remote step implementations, however, I do not believe the removal regresses them.

**The worker's `StepExecution` row has a single writer after dispatch.** For both `PartitionStep` and `RemoteStep`, that row is written only by the JVM that runs `AbstractStep.execute()` for it. The manager writes it only *before* dispatch (`RemoteStep` creates and updates it at lines 96-101; `SimpleStepExecutionSplitter` creates it for partitions) and afterwards only reads it (`RemoteStep.pollRemoteStep`, `MessageChannelPartitionHandler.pollReplies`).

**`SimpleJobOperator.stop()` never reaches those rows.** [The loop at line 355](https://github.com/spring-projects/spring-batch/blob/3c875a47eec3cb18c755e681bec7487d33f2fef2/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java#L360) does iterate them — the splitter adds them to the same `JobExecution` — and they even pass the `isRunning()` check, because the manager's in-memory copies stay at their initial `STARTING` status. But [`stepLocator.getStep(stepExecution.getStepName())` at line 358](https://github.com/spring-projects/spring-batch/blob/3c875a47eec3cb18c755e681bec7487d33f2fef2/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java#L363) returns `null`: the execution name carries the partition suffix (`workerStep:partition0`, from `SimpleStepExecutionSplitter`), the manager's job only knows `managerStep`, and `PartitionStep` is not a `StepLocator`, so no delegation occurs. I verified this by replaying that lookup against a partitioned job:

```
stepExecution name=managerStep            getStep(name)=PartitionStep  -> stopped
stepExecution name=workerStep:partition2  getStep(name)=NULL  -> skipped
stepExecution name=workerStep:partition1  getStep(name)=NULL  -> skipped
stepExecution name=workerStep:partition0  getStep(name)=NULL  -> skipped
```

**Stop reaches workers only through the shared database.** `stop()` writes `BATCH_JOB_EXECUTION.STATUS`, and each worker observes it inside its own `SimpleJobRepository.update(StepExecution)`, where `jobExecutionDao.synchronizeStatus(jobExecution)` refreshes the parent status and `setTerminateOnly()` is then applied. Note that this happens on the worker thread, not on the stopping thread.

Even if the step name did resolve, `StoppableStep.stop()` only flips an in-memory flag on the manager's own copy of the object, so it could not interrupt a worker running in another JVM in any case. Cross-JVM interruption is by design a database-driven, chunk-boundary mechanism.

### 3. On the incomplete critical section

You are right that the critical section in this PR is not complete: it does not cover the window before `stepExecutionLocks.put(...)` and after `remove(...)`, and it cannot cover writers in other JVMs.

Rather than patching those gaps here, I think the underlying question is which approach to #5308 to take, and I would like the maintainer's guidance before investing further.

@fmbenhassine — there are two alternative directions on the table for #5308:

- **This PR (#5448)** keeps `stop()` writing the step execution and guards the two writers with a per-execution semaphore. As noted above, that boundary is inherently partial.
- **#5442** removes the second writer instead. `stop()` no longer persists step executions: it marks the job `STOPPING` in a short transaction, signals each running step, and then waits — outside any transaction — for the executing thread to terminate and persist its own stopped state, with a configurable timeout. `StoppableStep.stop()` then only sets `terminateOnly`, and the thread executing the step owns the `STOPPED` / exit status / end time transition, making it the sole writer of that row.

I currently favour the second direction, and it also covers the concern behind #4023 better than a purely non-blocking stop does. The dangling `STARTED` status there comes from the process being terminated without a shutdown hook — for example a pod being killed in Kubernetes — before the stopped state was ever persisted. Because `stop()` in #5442 blocks until the executing thread has persisted the terminal state, the caller has a durable point to synchronize graceful shutdown against, instead of returning while the status may still be in flight.

Could you indicate which direction you would prefer? I am happy to complete either one.

PR #5454, opened 2026-07-08
Fix #5415: add new KafkaConsumerRecodItemStreamReader

### LINKED ISSUE #5415: Extend the wrapper type to KafkaItemReader<T> like in JmsItemReader<T>
**Expected Behavior**

Basically I am asking that `KafkaItemReader`, without retrocompatibility break, can be declared as `KafkaItemReader<ConsumerRecord<Key, Value>>` and yields both the key and the value of the Kafka message, like a `JmsItemReader<Message<T>>` yields the whole message.

This would need to be done via an additional (optional) parameter `itemType` that can either be `V` (and its subclasses) or `ConsumerRecord.class`.

**Current Behavior**

I have clearly seen from the source code of `KafkaItemReader` that the value is directly extracted and returned from the `read` method without the type check implemented by `JmsItemReader`.

Unless reimplementing my own KafkaItemReader into my project, I see no way of getting the message key in a Spring Batch project.

**Context**

I need to extract **both** the message key and payload from `KafkaItemStreamReader`.

I have read and agree with https://github.com/spring-projects/spring-batch/pull/4308 that behaviour must be consistent across typed components.

I have read @fmbenhassine's answer to the PR. I agree with rejecting that particular PR that adds the `Message` type (which is part of JMS and not Kafka standard), but I have found the requested behaviour implemented into _today_'s implementation of `JmsItemReader`. 
To my understanding (git blame), that feature was added 9 months ago from today into the `JmsItemReader`.

I am available to contribute with a PR after a quick discussion. My idea is to add the same type check done by `JmsItemReader` with `itemType` into the `KafkaItemReader`.

djechelon:
As discussed in #5415, goal is to implement a new Kafka-based ItemReader that returns not only the message body (generic V), but the whole envelope (as ConsumerRecord<K,V>).

In order to maintain backwards compatibility, the following has been done:

- KafkaItemReader was mostly moved to an abstract class, with 3 generics. Of these, <K,V> are the usual Kafka key-value, and T is the actual type returned by the reader
- KafkaItemReader changed to extend the abstract class and set T to be V
- KafkaConsumerRecordItemReader to extend the abstract class and set T to be ConsumerRecord<K,V>
- Changed the builder to return the consumer record instance with a different build() method
- Worked on the unit tests
- Signed off commit and updated copyright notice

Please consider that locally I am experiencing `mvn verify` even on main as I started working on the fork. I was able to verify that **my** affected tests work, failing tests occur on unaffected classes.

I wish to thank whoever will spend their precious time paying attention to this contribution.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/kafka/AbstractKafkaItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/kafka/AbstractKafkaItemReader.java
@@ -1,0 +1,236 @@
+/*
+ * Copyright 2019-2026 the original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *          https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.batch.infrastructure.item.kafka;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+import org.apache.kafka.clients.consumer.ConsumerConfig;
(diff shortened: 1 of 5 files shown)


quaff:
`JmsItemReader` should be refactored like this if the PR is accepted.

djechelon:
About that, consider also that `AmqpItemReader` shows the same mechanism as Jms.

Refactoring Jms and Amqp would mean that the `itemType` mechanism becomes broken (or turns into a backwards-compatibility-porkaround), as currently it is used to choose between returning the `Message<T>` wrapper or the plain `T` value

quaff:
> About that, consider also that `AmqpItemReader` shows the same mechanism as Jms.
> 
> Refactoring Jms and Amqp would mean that the `itemType` mechanism becomes broken (or turns into a backwards-compatibility-porkaround), as currently it is used to choose between returning the `Message<T>` wrapper or the plain `T` value

We could deprecate current reader and add new readers.

PR #5457, opened 2026-07-10
Improve DefaultJobParametersConverter and JsonJobParametersConverter

quaff:
1. let `decode()` supports generics.
2. add `@Nullable` to `JobParameterDefinition`.
3. let `JsonJobParametersConverter` encode `identifying` as boolean instead of String, and add tests to verify both boolean and String can be decoded to java `Boolean`.


fmbenhassine:
LGTM 👍 Rebased and merged as 41ddb67eef71535f5e960ee7d1f6a51af7c9d50c. Thank you for your contribution!

PR #5467, opened 2026-07-22
Fix duplicate listener registration in ChunkOrientedStepBuilder

### LINKED ISSUE #5466: Interface listener callbacks are invoked twice when an item handler also has a listener annotation
**Bug description**

If an `ItemReader`, `ItemProcessor`, or `ItemWriter` implements a `StepListener` sub-interface (e.g. `ItemProcessListener`) and also has a method annotated with a listener annotation (e.g. `@BeforeStep`), the object is registered as a listener twice, so every interface callback fires twice per item. The annotated method itself fires once; only the interface callbacks are doubled.

`ChunkOrientedStepBuilder#addAsStreamAndListener` has two non-exclusive branches: `instanceof StepListener` adds the raw object, and `StepListenerFactoryBean.isListener(...)` adds a proxy. Since `AbstractListenerFactoryBean#getObject` resolves callbacks **by interface as well as by annotation**, the proxy delegates the same interface callbacks back to the same target.

This is a regression from 5.2.x, where `SimpleStepBuilder#registerAsStreamsAndListeners` used a single `StepListenerFactoryBean.isListener(...)` branch, so the object was registered once. It fails silently: a listener that accumulates state (counters, files) produces doubled results while the job still completes successfully.

**Environment**

Spring Batch 6.0.4 (Spring Boot 4.1.0), Java 17+, no database. Still present on `main`.

**Steps to reproduce**

Run the reproducer below (`./gradlew bootRun`). Two identical jobs run: one processor only implements `ItemProcessListener`, the other is the same processor plus one `@BeforeStep` method.

```text
without a listener annotation = [beforeProcess:1, afterProcess:1, beforeProcess:2, afterProcess:2]
with @BeforeStep added        = [beforeStep, beforeProcess:1, beforeProcess:1, afterProcess:1, afterProcess:1, beforeProcess:2, beforeProcess:2, afterProcess:2, afterProcess:2]
```

`./gradlew test` asserts the same: `callbacksFireOnceWhenOnlyTheInterfaceIsImplemented` passes, `callbacksShouldAlsoFireOnceWhenAListenerAnnotationIsAdded` fails.

**Expected behavior**

Adding an annotated listener method must not change how the interface callbacks are invoked:

```text
with @BeforeStep added        = [beforeStep, beforeProcess:1, afterProcess:1, beforeProcess:2, afterProcess:2]
```

https://github.com/benelog/spring-batch-practice/tree/main/issue-duplicate-listener

**Suggested fix**

Make the two branches mutually exclusive, as in 5.2.x. `StepListenerFactoryBean.isListener` already returns `true` for interface-only implementations, so the factory branch covers both cases:

```java
if (StepListenerFactoryBean.isListener(itemHandler)) {
    this.stepListeners.add(StepListenerFactoryBean.getListener(itemHandler));
}
else if (itemHandler instanceof StepListener listener) {
    this.stepListeners.add(listener);
}
```

**Related**

- #5451 — `@BeforeStep`/`@AfterStep` ignored by `ChunkOrientedStepBuilder#listener(Object)`: different code path, same class.
- #2077 (BATCH-1493) — the same "registered twice, invoked twice" symptom via manual registration, fixed in 2.x.

benelog:
Resolves #5466

`ChunkOrientedStepBuilder#addAsStreamAndListener` registered an item handler as a listener in two non-exclusive branches: once as the raw object (`instanceof StepListener`) and once as a `StepListenerFactoryBean` proxy (`isListener(...)`). Since `AbstractListenerFactoryBean#getObject` resolves callbacks by interface as well as by annotation, an item handler that implements a listener interface **and** has a listener-annotated method was registered twice, and every interface callback fired twice per item.

This change registers through `StepListenerFactoryBean` only, restoring the 5.2.x behavior of `SimpleStepBuilder#registerAsStreamsAndListeners`:

- `isListener(...)` already returns `true` for interface-only implementations, and
- `getObject()` returns the delegate unchanged when no annotated methods are present (`synthetic == false`), so interface-only handlers are still registered as before.

The added integration test fails on `main` with `expected: <1> but was: <2>` for `beforeProcess`/`afterProcess` counts and passes with this fix.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/ChunkOrientedStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/ChunkOrientedStepBuilder.java
@@ -485,13 +485,14 @@ public class ChunkOrientedStepBuilder<I, O> extends StepBuilderHelper<ChunkOrien
 		if (itemHandler instanceof ItemStream itemStream) {
 			this.streams.add(itemStream);
 		}
-		// Register as listener if implements the interface
-		if (itemHandler instanceof StepListener listener) {
-			this.stepListeners.add(listener);
-		}
-		// Register as listener if annotated methods are present
+		// Register as listener if it implements a listener interface or has annotated
+		// methods. The factory covers both cases, so the item handler is registered
+		// only once even if it mixes interfaces and annotations.
 		if (StepListenerFactoryBean.isListener(itemHandler)) {
 			StepListener listener = StepListenerFactoryBean.getListener(itemHandler);
+			this.stepListeners.add(listener);
+		}
+		else if (itemHandler instanceof StepListener listener) {
 			this.stepListeners.add(listener);
 		}
 	}
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/step/item/ChunkOrientedStepIntegrationTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/step/item/ChunkOrientedStepIntegrationTests.java
@@ -15,10 +15,12 @@
  */
 package org.springframework.batch.core.step.item;
 
+import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.annotation.BeforeStep;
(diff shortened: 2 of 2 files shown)


fmbenhassine:
Rebased and merged as 2708101c9c6a1ab2d129cc6de301e12186377159. Thank you for your contribution  🙏

PR #5469, opened 2026-07-28
Add missing StepListenerMetaData entry for @OnChunkError

### LINKED ISSUE #5468: @OnChunkError is never invoked on a chunk-oriented step
**Bug description**

`@OnChunkError` is never invoked on a chunk-oriented step.

`ChunkOrientedStepBuilder#listener(Object)` scans for `@OnChunkError` and builds a listener proxy for it, but `StepListenerMetaData` has no entry for that annotation. The proxy is registered and the callback is never wired, so the method is silently skipped. No exception is thrown and no warning is logged.

`@BeforeChunk` and `@AfterChunk` on the same class work, which makes the gap easy to miss.

**Environment**

* Spring Batch 6.0.4
* Spring Boot 4.1.0
* Java 17

**Steps to reproduce**

Minimal reproducer: https://github.com/benelog/spring-batch-practice/tree/main/issue-chunk-error-listener

```java
public class AnnotatedChunkListener {

	@BeforeChunk
	public void beforeChunk(Chunk<Integer> chunk) { ... }

	@AfterChunk
	public void afterChunk(Chunk<Integer> chunk) { ... }

	@OnChunkError                                       // never called
	public void onChunkError(Exception exception, Chunk<Integer> chunk) { ... }
}
```

The listener is registered with `listener(Object)` on a chunk-oriented step whose writer throws on the second chunk.

```bash
./gradlew bootRun
```

Actual output:

```text
[DIAG] ChunkListener interface    = [interface:beforeChunk, interface:afterChunk, interface:beforeChunk, interface:onChunkError]
[DIAG] @OnChunkError              = [annotation:beforeChunk, annotation:afterChunk, annotation:beforeChunk]
[DIAG] @AfterChunkError (5.x)     = []
```

**Expected behavior**

```text
[DIAG] @OnChunkError              = [annotation:beforeChunk, annotation:afterChunk, annotation:beforeChunk, annotation:onChunkError]
```

The same expectation is asserted as a JUnit test in the reproducer (`./gradlew test`): the two control tests pass and the two tests for this issue fail.

**Root cause**

`StepListenerMetaData` (6.0.4) has no `OnChunkError` constant. `AFTER_CHUNK_ERROR` still maps to the deprecated `ChunkListener#afterChunkError(ChunkContext)`:

```java
BEFORE_CHUNK("beforeChunk", "before-chunk-method", BeforeChunk.class, ChunkListener.class, Chunk.class),
AFTER_CHUNK("afterChunk", "after-chunk-method", AfterChunk.class, ChunkListener.class, Chunk.class),
AFTER_CHUNK_ERROR("afterChunkError", "after-chunk-error-method", AfterChunkError.class, ChunkListener.class,
		ChunkContext.class),
// no entry for OnChunkError
```

`AbstractListenerFactoryBean#getObject` iterates over these constants, so the proxy it returns has nothing bound to `onChunkError(Exception, Chunk)`. `ChunkOrientedStep` calls only `onChunkError(Exception, Chunk)`; the deprecated `afterChunkError(ChunkContext)` is called by `TaskletStep` alone.

`@BeforeChunk` / `@AfterChunk` were migrated to the `Chunk` signature in #4961, and `@OnChunkError` was added in 6.0 with a javadoc that documents `void onChunkError(Exception, Chunk)`. The metadata entry appears to have been left out of both changes.

`@AfterChunkError` is a separate, smaller gap: `ChunkOrientedStepBuilder#listener(Object)` does not scan for it at all, so a class carrying only that annotation registers nothing.

**Suggested fix**

Add the missing constant:

```java
ON_CHUNK_ERROR("onChunkError", "on-chunk-error-method", OnChunkError.class, ChunkListener.class,
		Exception.class, Chunk.class),
```

I am happy to submit a pull request for this.

**Impact**

An annotation-based chunk error handler compiles, registers without complaint, and does nothing. Error notifications, cleanup, and failure counters written this way are silently lost, and the step reports the failure normally, so nothing points at the listener.

**Workaround**

Implement `ChunkListener` and override `onChunkError(Exception, Chunk)`.

**Related issues**

* #4961 / #4969 — `@BeforeChunk` / `@AfterChunk` fixed to accept `Chunk` as documented. `AFTER_CHUNK_ERROR` was not migrated in that change.
* #5297 — `@AfterChunkError` with a `ChunkContext` parameter on `SimpleStepBuilder`. Different path: that one throws `IllegalArgumentException`, this one fails silently.
* #5226 — `ChunkListener` callbacks on an `ItemReader` not invoked in 6.x.

benelog:
Resolves #5468

The parameter types follow `ON_WRITE_ERROR`, which has the same `(Exception, Chunk)` shape. No other change was needed: `ChunkOrientedStepBuilder#listener(Object)` already scans for the annotation and `StepListenerFactoryBean` reads `StepListenerMetaData.values()`.

## Points for review

- The new constant sits after `AFTER_CHUNK_ERROR` to keep the chunk callbacks together, which shifts the ordinals below it. Nothing reads the ordinal, but say the word if you would rather have it appended at the end.
- `taskletListenerMetaData()` is untouched. It feeds the XML namespace for `TaskletStep`, which calls `afterChunkError(ChunkContext)` and not `onChunkError`. Happy to expose `on-chunk-error-method` there too if you want it in the namespace.
- `@AfterChunkError` on a chunk-oriented step stays broken: `listener(Object)` does not scan for it. Since it maps to a method deprecated for removal in 6.2 that `ChunkOrientedStep` never calls, wiring it up seemed like the wrong direction. Noted in the issue rather than fixed here.

## Verification

Besides the unit test in the diff, the reproducer from #5468 was re-run against a locally built `6.0.5-SNAPSHOT`:

```text
[DIAG] @OnChunkError = [annotation:beforeChunk, annotation:afterChunk, annotation:beforeChunk, annotation:onChunkError]
```

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/listener/StepListenerMetaData.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/listener/StepListenerMetaData.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2025 the original author or authors.
+ * Copyright 2002-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.batch.core.annotation.BeforeProcess;
 import org.springframework.batch.core.annotation.BeforeRead;
 import org.springframework.batch.core.annotation.BeforeStep;
 import org.springframework.batch.core.annotation.BeforeWrite;
+import org.springframework.batch.core.annotation.OnChunkError;
 import org.springframework.batch.core.annotation.OnProcessError;
 import org.springframework.batch.core.annotation.OnReadError;
 import org.springframework.batch.core.annotation.OnSkipInProcess;
@@ -59,6 +60,8 @@ public enum StepListenerMetaData implements ListenerMetaData {
 	AFTER_CHUNK("afterChunk", "after-chunk-method", AfterChunk.class, ChunkListener.class, Chunk.class),
 	AFTER_CHUNK_ERROR("afterChunkError", "after-chunk-error-method", AfterChunkError.class, ChunkListener.class,
 			ChunkContext.class),
+	ON_CHUNK_ERROR("onChunkError", "on-chunk-error-method", OnChunkError.class, ChunkListener.class, Exception.class,
+			Chunk.class),
 	BEFORE_READ("beforeRead", "before-read-method", BeforeRead.class, ItemReadListener.class),
 	AFTER_READ("afterRead", "after-read-method", AfterRead.class, ItemReadListener.class, Object.class),
 	ON_READ_ERROR("onReadError", "on-read-error-method", OnReadError.class, ItemReadListener.class, Exception.class),
--- a/spring-batch-core/src/test/java/org/springframework/batch/core/listener/StepListenerFactoryBeanTests.java
+++ b/spring-batch-core/src/test/java/org/springframework/batch/core/listener/StepListenerFactoryBeanTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2025 the original author or authors.
+ * Copyright 2002-2026 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 8bf3bada80daeda858af859df7e4e108011c9d49. Thank you for your contribution!

PR #5476, opened 2026-08-03
Fix the tracing example in the reference documentation

### LINKED ISSUE #5475: Reference documentation's tracing example registers a handler that creates no spans and fails the job
**Bug description**

The [Tracing section](https://docs.spring.io/spring-batch/reference/spring-batch-observability/micrometer.html#tracing) of the reference documentation says to register an `ObservationHandler` "that supports tracing, such as `TracingAwareMeterObservationHandler`", and shows that handler as the only example.

That handler does not create spans. Its javadoc describes it as "a handler that can wrap another one and makes the tracing data available for it (e.g.: exemplars)". Registering it alone, exactly as documented, creates no spans and throws when the job observation is stopped.

**Environment**

Spring Batch 6.0.4 (the same text is in `main`), Micrometer Tracing 1.6.6 with the OpenTelemetry bridge, OpenTelemetry SDK 1.54.1, Java 21. No Spring Boot, since the documented configuration is a manual one, and no database: the job runs on a `ResourcelessJobRepository`.

**Steps to reproduce**

Configure the registry as documented and run any job:

```java
ObservationRegistry observationRegistry = ObservationRegistry.create();
observationRegistry.observationConfig()
        .observationHandler(new TracingAwareMeterObservationHandler<>(
                new DefaultMeterObservationHandler(meterRegistry), tracer));
```

The job and its steps are built with `.observationRegistry(observationRegistry)`, and spans are collected with the OpenTelemetry SDK `InMemorySpanExporter`. No span is created, and the exception thrown when the job observation is stopped propagates out of `JobLauncher.run()`:

```
java.lang.IllegalArgumentException: Context does not have an entry for key
  [class io.micrometer.tracing.handler.TracingObservationHandler$TracingContext]
	at io.micrometer.tracing.handler.TracingAwareMeterObservationHandler.onStop(TracingAwareMeterObservationHandler.java:78)
	at org.springframework.batch.core.job.AbstractJob.stopObservation(AbstractJob.java:370)
```

**Expected behavior**

What the same section promises: "a trace for each job execution and a span for each step execution".

https://github.com/benelog/spring-batch-practice/tree/main/issue-tracing-handler

Two tasklet steps, no Spring Boot. `./gradlew run` runs the same job with the documented handler, with a tracing handler and with both. `./gradlew test` asserts all three: it fails only on the documented one.

**Cause**

`TracingAwareMeterObservationHandler.onStop()` requires a `TracingContext`, which only a `TracingObservationHandler` implementation puts into the observation context. It is a decorator, usable only alongside a span-producing handler. Boot gets this right: `TracingAndMeterObservationHandlerGroup.registerMembers()` registers the `TracingObservationHandler` beans and only wraps `MeterObservationHandler` instances. So this affects the manual configuration shown in the documentation.

**Suggested fix**

Either register a span-producing handler:

```java
observationRegistry.observationConfig()
        .observationHandler(new DefaultTracingObservationHandler(tracer));
```

or, if the example was meant to cover metrics and tracing together (the metrics section right above registers a `DefaultMeterObservationHandler` on a bean of the same name), register both:

```java
observationRegistry.observationConfig()
        .observationHandler(new DefaultTracingObservationHandler(tracer))
        .observationHandler(new TracingAwareMeterObservationHandler<>(
                new DefaultMeterObservationHandler(meterRegistry), tracer));
```

The reproducer covers both: 3 spans with the first, 3 spans and 6 `spring.batch.*` meters with the second. The sentence above the snippet needs to change as well. I am happy to submit a PR for whichever the team prefers.

The example was introduced in 6.0 (commit ef1de538, "Remove usage of Micrometer's global static meter registry", #4968); the 5.x documentation had no handler example.

benelog:
Resolves #5475

The example now registers a `DefaultTracingObservationHandler`, and the sentence above it no longer presents `TracingAwareMeterObservationHandler` as a handler that supports tracing on its own.

I kept the metrics handler in the example rather than replacing it, because both sections define a bean named `observationRegistry`: a tracing-only snippet here would quietly drop the metrics from the section above. Happy to reduce it to the tracing handler alone if you prefer the shorter example.

Handler order is not a requirement, the reversed order produced the same spans and meters, but I registered the tracing handler first to match `TracingAndMeterObservationHandlerGroup` in Spring Boot.

### DIFF
--- a/spring-batch-docs/modules/ROOT/pages/spring-batch-observability/micrometer.adoc
+++ b/spring-batch-docs/modules/ROOT/pages/spring-batch-observability/micrometer.adoc
@@ -89,8 +89,9 @@ public class MyTimedTasklet implements Tasklet {
 == Tracing
 
 As of version 5, Spring Batch provides tracing through Micrometer's `Observation` API. By default, tracing is disabled.
-To enable it, you need to define an `ObservationRegistry` bean configured with an `ObservationHandler` that supports tracing,
-such as `TracingAwareMeterObservationHandler`:
+To enable it, you need to define an `ObservationRegistry` bean configured with a `TracingObservationHandler`,
+such as `DefaultTracingObservationHandler`. The following example registers it next to the metrics handler
+shown above, so that the same registry collects both metrics and traces:
 
 [source, java]
 ----
@@ -99,10 +100,15 @@ public ObservationRegistry observationRegistry(MeterRegistry meterRegistry, Trac
     DefaultMeterObservationHandler observationHandler = new DefaultMeterObservationHandler(meterRegistry);
     ObservationRegistry observationRegistry = ObservationRegistry.create();
     observationRegistry.observationConfig()
+            .observationHandler(new DefaultTracingObservationHandler(tracer))
             .observationHandler(new TracingAwareMeterObservationHandler<>(observationHandler, tracer));
     return observationRegistry;
 }
 ----
+
+NOTE: `TracingAwareMeterObservationHandler` does not create spans. It wraps a `MeterObservationHandler` to make
+tracing data available to it (for exemplars, for example), so it requires a `TracingObservationHandler` to be
+registered as well.
 
 With that in place, Spring Batch will create a trace for each job execution and a span for each step execution.
 


fmbenhassine:
Thank you for reporting the issue and for contributing a PR! LGTM 👍

PR #5480, opened 2026-08-05
GH-5470: Avoid nested query in JdbcStepExecutionDao.getLastStepExecution

### LINKED ISSUE #5470: JdbcStepExecutionDao.getLastStepExecution nests a query inside an open row-limited ResultSet, breaking every restart on CockroachDB
### Summary

`JdbcStepExecutionDao.getLastStepExecution` issues a nested query from inside an open, row-limited `ResultSet`. On PostgreSQL this is fine. On CockroachDB it is rejected outright, so **any step that already has a prior `StepExecution` under the same `JobInstance` fails**, and the connection is discarded as broken. In practice that means restarts.

This is a portability defect rather than a performance one, and it is not caught by the existing test suite because it only manifests on a database that refuses multiple simultaneously-open portals.

### Affected versions

Present in **6.0.4** (current release) and on **`main`** as of today.

### The code

`spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcStepExecutionDao.java`, in `getLastStepExecution`:

```java
return getJdbcTemplate().query(getQuery(GET_LAST_STEP_EXECUTION), (PreparedStatementCallback<StepExecution>) statement -> {
    statement.setMaxRows(1);                       // <-- opens a row-limited portal
    ...
    try (ResultSet rs = statement.executeQuery()) {
        if (rs.next()) {
            Long jobExecutionId = rs.getLong(19);
            JobExecution jobExecution = new JobExecution(jobExecutionId, jobInstance,
                    jobExecutionDao.getJobParameters(jobExecutionId));   // <-- nested query, cursor still open
            ...
```

`JdbcJobExecutionDao.getJobParameters` prepares and executes a second statement on the same connection while the first cursor is still open.

### Why `setMaxRows(1)` matters

The nesting alone is not sufficient. The distinction that matters is **suspended and still executable versus completed**, not simply open versus closed.

`setMaxRows(1)` makes pgjdbc send `Execute` with a protocol row count of 1. Per the protocol, if a non-zero row count stops execution before the portal completes, the backend replies `PortalSuspended` rather than `CommandComplete`. pgjdbc has already sent `Sync`, but because the connection is inside an explicit transaction that `Sync` yields `ReadyForQuery(T)` without ending the transaction or disposing the suspended portal. The nested query then attempts its protocol preparation while that portal is still actively suspended, and CockroachDB rejects it.

By contrast `LIMIT 1` in the SQL, with no JDBC max rows, produces `Execute(portal, 0)`, which fetches to completion and yields `CommandComplete`. That is why the two are not interchangeable here.

Relevant detail: on this `setMaxRows` / `fetchSize == 0` path pgjdbc uses the *unnamed* portal and returns no named `ResultCursor`, so its explicit portal-cleanup machinery does not apply. Closing the Java `ResultSet` therefore does not necessarily emit a wire-level `Close` for the suspended unnamed portal.

Verified empirically in an equivalent implementation: `LIMIT 1` works, switching to `statement.setMaxRows(1)` reintroduces the failure, removing it resolves it.

Qualifications, so this is not overstated:
- `setMaxRows(1)` does not *guarantee* `PortalSuspended`. A zero-row result reaches `CommandComplete`.
- A positive `fetchSize` can create the same condition via a named fetch portal, so `setMaxRows` is not the only trigger.
- The explanation does not apply in simple-query mode.
- Prepared-statement caching and `prepareThreshold` can affect reproduction, since a subsequent unnamed `Bind` normally replaces the previous unnamed portal.
- The wire state above is **inferred from behaviour, not captured**. A pgjdbc `FINEST` trace or pgwire capture showing `Execute(portal=null, limit=1)` then `PortalSuspended` then `Sync` then `ReadyForQuery(T)` then `Parse` then `ErrorResponse` would confirm it. I have not captured one.

### Observed failure

```
org.postgresql.util.PSQLException: ERROR: unimplemented: multiple active portals is in preview,
please set session variable multiple_active_portals_enabled to true to enable them
  Detail: cannot perform operation sql.PrepareStmt while a different portal is open
  Hint: You have attempted to use a feature that is not yet implemented.
See: https://go.crdb.dev/issue-v/40195/v26.2
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2904)
	...
	at org.springframework.batch.core.repository.dao.jdbc.JdbcJobExecutionDao.getJobParameters(JdbcJobExecutionDao.java:450)
	at org.springframework.batch.core.repository.dao.jdbc.JdbcStepExecutionDao.lambda$getLastStepExecution$0(JdbcStepExecutionDao.java:341)
	at org.springframework.batch.core.repository.dao.jdbc.JdbcStepExecutionDao.getLastStepExecution(JdbcStepExecutionDao.java:332)
	at org.springframework.batch.core.repository.explore.support.SimpleJobExplorer.getLastStepExecution(SimpleJobExplorer.java:267)
	at org.springframework.batch.core.job.SimpleStepHandler.handleStep(SimpleStepHandler.java:95)
```

HikariCP then marks the connection broken (`SQLSTATE 0A000`), which surfaces downstream as `Connection is closed` and `JDBC rollback failed`, so the original cause is easy to misdiagnose.

### Impact and why it looks intermittent

To be precise about the trigger, since "restart" is not quite the right framing: `SimpleStepHandler.handleStep` always runs with a `JobInstance` present and calls `getLastStepExecution` on ordinary first attempts too, simply getting `null` back.

The dangerous condition is narrower: **a prior `StepExecution` exists for the same `JobInstance` and the same step name**, so `rs.next()` is true and the nested hydration query actually runs. A first execution with a fresh step name is safe because the outer query returns zero rows, the `Execute` completes rather than suspending, and `rs.next()` is false.

In our case this presented as one service failing every attempt while others were unaffected, which sent us looking at the service rather than the framework. `SimpleStepExecutionSplitter` calls `getLastStepExecution` once per partition, so partitioned steps multiply the exposure, but partitioning is a multiplier rather than the cause.

### Reproduction

Against `cockroachdb/cockroach:v26.2.3` (Testcontainers), with a JDBC `JobRepository` and no CockroachDB session flags set:

1. Run any job to completion or failure so a `JobInstance` exists.
2. Relaunch the same job instance so Spring Batch takes the restart path.
3. `getLastStepExecution` throws the error above.

### On PR #5133

[#5133](https://github.com/spring-projects/spring-batch/pull/5133) refactors this method to reuse `JdbcJobExecutionDao::getJobExecution`, and at first glance looks like it would resolve this. It does not: the replacement call is still made from inside the open `ResultSet`.

```java
+   return new StepExecutionRowMapper(
+           jobExecutionDao.getJobExecution(rs.getLong("JOB_EXECUTION_ID")))
+       .mapRow(rs, 0);
```

Since `getJobExecution` issues more statements than the `getJobParameters` call it replaces, the nesting would arguably widen. Worth noting in case that PR is considered a fix for this.

### Suggested direction

Hoist the dependent read out of the cursor rather than changing which call is nested: read the row (or just its id) first, let that statement reach `CommandComplete`, then resolve the `JobExecution`. `getStepExecution(long)` already performs the sequential equivalent and could be delegated to.

Important subtlety: closing the Java `ResultSet` is **not** sufficient on its own. The first query must actually complete, which means limiting via SQL `LIMIT 1` rather than JDBC `setMaxRows(1)`. Otherwise the id query leaves its own suspended portal and the problem simply moves.

Happy to open a PR if the maintainers would like it in that shape.

### Workaround for anyone hitting this

Either enable `multiple_active_portals_enabled` on the CockroachDB connection the `JobRepository` uses (it is a preview feature), or override `JdbcStepExecutionDao.getLastStepExecution` via a `JdbcJobRepositoryFactoryBean` subclass so it reads the id first and then delegates to `getStepExecution(long)`. We took the latter route to avoid the preview flag.

---

*Wire-protocol account in this report was reviewed and corrected after an independent technical review; an earlier revision attributed the behaviour to a deferred `Close` at the next `Sync`, which is wrong on two counts (pgjdbc sends `Sync` immediately, including inside explicit transactions, and `Sync` is not a deferred `Close`). The observable behaviour and the fix are unchanged.*

arimu1:
## Fixes #5470

`JdbcStepExecutionDao.getLastStepExecution` previously called
`jobExecutionDao.getJobParameters` (a second JDBC statement) while a
`ResultSet` opened with `setMaxRows(1)` was still active. On databases
that refuse multiple simultaneously open portals (notably CockroachDB),
that nested prepare fails and restarts break whenever a prior
`StepExecution` exists for the same job instance and step name.

### Changes

- Select only `STEP_EXECUTION_ID` for the “last step execution” lookup
- Keep portable `statement.setMaxRows(1)` (no SQL `LIMIT`, which is not
  supported by all databases Spring Batch targets)
- After that statement completes and is closed by `JdbcTemplate`, load
  the full `StepExecution` via `getStepExecution(long)`

This matches the sequential approach suggested in the issue and the
workaround of reading the id first then delegating to
`getStepExecution(long)`.

Note: open PR #5133 still performs `getJobExecution` **inside** the open
`ResultSet` callback, so it does not address GH-5470.

### Testing

- `JdbcStepExecutionDaoTests` (6 tests, including new coverage for same-job
  and cross-job-execution “last step” selection)
- `./mvnw -pl spring-batch-core -am test -Dtest=JdbcStepExecutionDaoTests`
- JDK: Eclipse Temurin 25 (project requires 17; NullAway on this branch needs 22+)

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcStepExecutionDao.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/jdbc/JdbcStepExecutionDao.java
@@ -27,7 +27,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.batch.core.BatchStatus;
-import org.springframework.batch.core.ExitStatus;
 import org.springframework.batch.core.job.JobExecution;
 import org.springframework.batch.core.job.JobInstance;
 import org.springframework.batch.core.step.StepExecution;
@@ -95,7 +94,7 @@ public class JdbcStepExecutionDao extends AbstractJdbcBatchMetadataDao implement
 			""";
 
 	private static final String GET_LAST_STEP_EXECUTION = """
-			SELECT SE.STEP_EXECUTION_ID, SE.STEP_NAME, SE.START_TIME, SE.END_TIME, SE.STATUS, SE.COMMIT_COUNT, SE.READ_COUNT, SE.FILTER_COUNT, SE.WRITE_COUNT, SE.EXIT_CODE, SE.EXIT_MESSAGE, SE.READ_SKIP_COUNT, SE.WRITE_SKIP_COUNT, SE.PROCESS_SKIP_COUNT, SE.ROLLBACK_COUNT, SE.LAST_UPDATED, SE.VERSION, SE.CREATE_TIME, JE.JOB_EXECUTION_ID, JE.START_TIME, JE.END_TIME, JE.STATUS, JE.EXIT_CODE, JE.EXIT_MESSAGE, JE.CREATE_TIME, JE.LAST_UPDATED, JE.VERSION
+			SELECT SE.STEP_EXECUTION_ID
 			FROM %PREFIX%JOB_EXECUTION JE
 				JOIN %PREFIX%STEP_EXECUTION SE ON SE.JOB_EXECUTION_ID = JE.JOB_EXECUTION_ID
 			WHERE JE.JOB_INSTANCE_ID = ? AND SE.STEP_NAME = ?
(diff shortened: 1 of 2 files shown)


arimu1:
Follow-up for the portal-completion case called out in #5470:

Nesting alone was not enough. `setMaxRows(1)` can leave a suspended portal (`PortalSuspended` rather than `CommandComplete`) on pgjdbc-style wire protocol, so a subsequent prepare still fails on CockroachDB even after the Java `ResultSet` is closed.

This push:
- drops `statement.setMaxRows(1)` on the id-only lookup
- keeps portable `ORDER BY SE.CREATE_TIME DESC, SE.STEP_EXECUTION_ID DESC` (no SQL `LIMIT`)
- reads only the first ordered row in Java, lets the statement complete, then loads via `getStepExecution(long)`
- documents why default fetch size + first-row consumption is acceptable for large history

`JdbcStepExecutionDaoTests`: 6/6 pass.

PR #5483, opened 2026-08-06
Fix incorrect nullability contract for encoding in StaxEventItemReader

Gorock:
The encoding can be null Based on the Javadoc but Jspecify gives a warning when passing null. Which is confusing

this is my first Contribution to Spring Batch so please tell me if i forget something.


fmbenhassine:
Good catch! Thank you for reporting this and for contributing a fix.

Rebased and merged as ff4e7cf9609db7dd83fb1335e5ce2af0206e5d6e.

Thank you for your first contribution 🙏

PR #5488, opened 2026-08-09
Add ResourcesItemReaderBuilder with support for a file pattern

### LINKED ISSUE #5487: Add a builder for ResourcesItemReader with support for specifying resources via a pattern
**Expected Behavior**

A builder for `ResourcesItemReader`, consistent with the other readers and writers in the `org.springframework.batch.infrastructure.item.file.builder` package, that also supports specifying the resources with a file pattern:

```java
ResourcesItemReader reader = new ResourcesItemReaderBuilder()
    .filesPattern("/data/inputs/*.csv")
    .name("resourcesReader")
    .build();
```

**Current Behavior**

`ResourcesItemReader` is the only reader in the `org.springframework.batch.infrastructure.item.file` package without a builder. It can only be configured through the default constructor and `setResources(Resource[])`.

Its Javadoc even suggests configuring it with a pattern ("Hint: use a pattern to configure."), which worked out of the box in XML configuration thanks to `ResourceArrayPropertyEditor`. In Java configuration, however, resolving the pattern is left to the user:

```java
var resourcePatternResolver = new PathMatchingResourcePatternResolver();
ResourcesItemReader reader = new ResourcesItemReader();
reader.setResources(resourcePatternResolver.getResources("file:/data/inputs/*.csv"));
reader.setName("resourcesReader");
```

**Context**

This is a companion of #5056, which proposes the same pattern support for `MultiResourceItemReaderBuilder` (PR #5071). The proposed `filesPattern(String)` method follows the same naming and semantics, so the two builders would stay consistent.

I will submit a pull request for this shortly.

benelog:
Resolves #5487

The new `ResourcesItemReaderBuilder` follows the structure of the existing builders in the same package. Since `ResourcesItemReader` has no `saveState` or `strict` properties, the builder only exposes `resources(Resource...)`, `filesPattern(String)` and `name(String)`. The `name` is optional because the reader falls back to a class-name-based execution context key.

The `filesPattern(String)` method has the same name and semantics as the one proposed for `MultiResourceItemReaderBuilder` in #5071: the pattern is resolved by `PathMatchingResourcePatternResolver` with a "file:" prefix at build time. The test resource files are named `resource1.txt` and `resource2.txt` so that they do not clash with the `test1.txt`/`test2.txt` files added by #5071 in the same directory, in case both PRs get merged.


fmbenhassine:
LGTM 👍 The only thing I have to say is to keep the builder agnostic to resources type, and let the user choose the prefix. For that, we can use a method overload rather than a new method (similar to #5071)

Merged as 1209216e14d686dd5f3582502834675e57e0b925, and refined in 930ef884df5362563b1db39353f5f9c7d1057997.

Thank you for all your contributions!

PR #5489, opened 2026-08-09
GH-5486: Fix NPE in JpaCursorItemReader.doClose() when reader was nev…

### LINKED ISSUE #5486: JpaCursorItemReader.doClose() missing null check on entityManager
**Bug description**

`JpaCursorItemReader#doClose()` calls `this.entityManager.close()` without a null check, even though the `entityManager` field is annotated `@Nullable` and is only assigned inside `doOpen()`. If `close()` is invoked on a reader whose `open(ExecutionContext)` was never called, this throws a `NullPointerException`, which `AbstractItemCountingItemStreamItemReader#close()` catches and re-wraps as `ItemStreamException: Error while closing item reader`.

Calling `close()` on a reader that was never opened is not an edge case. It's the normal Spring bean lifecycle. The `close()` method is the container's inferred `destroy` method, invoked on every singleton `ItemReader` bean when the `ApplicationContext` shuts down, regardless of whether that reader's step ever executed. The problem is that when in happens in applications that define many jobs/steps as beans but only run one job per application run (a common pattern for CLI-driven batch apps). In this case, every `JpaCursorItemReader` belonging to a step that didn't run gets closed anyway and now throws.

The null check existed before and was dropped in commit 551e26d1c31e561af62dc145076eeeccdb6ecdbf ("Migrate nullability annotations to JSpecify", issue #4673), whose purpose I believe was only to migrate nullability annotations, not to change runtime behavior:

Before (parent commit 76e723e):

```java
private EntityManager entityManager;
...
@Override
protected void doClose() {
    if (this.entityManager != null) {
        this.entityManager.close();
    }
}
```

After (551e26d, and still present on current main / released in 6.0.4):

```java
private @Nullable EntityManager entityManager;
...
@Override
protected void doClose() {
    this.entityManager.close();
}
```

The `entityManager` field is explicitly annotated `@Nullable` in the very same commit that removed the code path handling that null.

**Environment**

- Spring Batch: 6.0.4
- Spring Boot: 4.1.0
- Java: 17
- Database: any relational database

**Steps to reproduce**

Alternative 1:

1. Download and extract the MCVE zip
2. Run any (or both) jobs with: `mvn spring-boot:run -Dspring-boot.run.arguments="jobName=productJob"` or `mvn spring-boot:run -Dspring-boot.run.arguments="jobName=customerJob"`
3. During application shutdown, observe the `ItemStreamException: Error while closing item reader.` warning log

Alternative 2:

1. Create a Spring Boot-based project with Spring Batch starter
2. Create two jobs that have have a step that uses a `JpaCursorItemReader`
3. Create a `CommandLineRunner` that gets the job to run via application args, and runs the requested job
4. During application shutdown, observe the `ItemStreamException: Error while closing item reader.` warning log

Alternative 3:

1. Build a `JpaCursorItemReader` (e.g. via `JpaCursorItemReaderBuilder`) but never call `open(ExecutionContext)` on it.
2. Call `close()` on it — directly, or implicitly via a Spring container destroying it as a singleton bean whose owning step never ran.
3. Observe a `NullPointerException`, wrapped as `ItemStreamException: Error while closing item reader.`

**Expected behavior**

Calling `close()` on a `JpaCursorItemReader` that was never opened should be a no-op, as it was prior to 6.0.

MCVE (Minimal Complete Verifiable Example):

- [spring-batch-mcve.zip](https://github.com/user-attachments/files/30836105/spring-batch-mcve.zip)

Run with:

```bash
mvn spring-boot:run -Dspring-boot.run.arguments="jobName=productJob"
mvn spring-boot:run -Dspring-boot.run.arguments="jobName=customerJob"
```

Expected log:

```
WARN --- [           main] o.s.b.f.support.DisposableBeanAdapter    : Custom destroy method 'close' on bean with name 'productReader' propagated an exception: org.springframework.batch.infrastructure.item.ItemStreamException: Error while closing item reader
```

Other minimal example as an alternative:

```java
import jakarta.persistence.EntityManagerFactory;
import org.springframework.batch.infrastructure.item.ItemStreamException;
import org.springframework.batch.infrastructure.item.database.JpaCursorItemReader;
import org.springframework.batch.infrastructure.item.database.builder.JpaCursorItemReaderBuilder;

import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.mockito.Mockito.mock;

class JpaCursorItemReaderCloseWithoutOpenTest {

	@Test
	void closingAReaderThatWasNeverOpenedThrows() {
		EntityManagerFactory entityManagerFactory = mock(EntityManagerFactory.class); // never invoked before close()

		JpaCursorItemReader<Object> reader = new JpaCursorItemReaderBuilder<>()
			.name("reader")
			.entityManagerFactory(entityManagerFactory)
			.queryString("select o from Object o")
			.build();

		// open(executionContext) is never called here, mirroring a bean whose
		// step never executed in the current process run
		assertThrows(ItemStreamException.class, reader::close);
	}
}
```

Resulting stack trace:

```
org.springframework.batch.infrastructure.item.ItemStreamException: Error while closing item reader
	at org.springframework.batch.infrastructure.item.support.AbstractItemCountingItemStreamItemReader.close(...)
Caused by: java.lang.NullPointerException: Cannot invoke "jakarta.persistence.EntityManager.close()" because "this.entityManager" is null
	at org.springframework.batch.infrastructure.item.database.JpaCursorItemReader.doClose(...)
```

**Suggested fix**

Restore the null check removed in 551e26d1c31e561af62dc145076eeeccdb6ecdbf:

```java
@Override
protected void doClose() {
    if (this.entityManager != null) {
        this.entityManager.close();
    }
}
```

dhritimanchakra:
…er opened
`entityManager` is only assigned in `doOpen()` and is `@Nullable`, but
`doClose()` called `entityManager.close()` with no null check. Spring
calls `close()` on every singleton `ItemReader` bean at context
shutdown regardless of whether its step ran, causing an NPE wrapped
as `ItemStreamException` for readers that were never opened.

The check was accidentally dropped in 551e26d when `@Nullable` was
added without preserving the runtime guard.

Fixes #5486

Thank you for taking time to contribute this pull request!
You might have already read the [contributor guide][1], but as a reminder, please make sure to:

* Rebase your changes on the latest `main` branch and squash your commits
* Add/Update unit tests as needed
* Run a build and make sure all tests pass prior to submission
* Sign-off commits according to the [Developer Certificate of Origin](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring)

For more details, please check the [contributor guide][1].
Thank you upfront!

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JpaCursorItemReader.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/database/JpaCursorItemReader.java
@@ -174,7 +174,9 @@ public class JpaCursorItemReader<T> extends AbstractItemCountingItemStreamItemRe
 	@SuppressWarnings("DataFlowIssue")
 	@Override
 	protected void doClose() {
-		this.entityManager.close();
+		if (this.entityManager != null) {
+			this.entityManager.close();
+		}
 	}
 
 }


fmbenhassine:
Thank you for the PR. LGTM 👍

Issue #4697, opened 2024-10-30
Jackson2ExecutionContextStringSerializer ignores QName prefix


jpraet:
**Bug description**
See https://github.com/FasterXML/jackson-databind/issues/4771: jackson QName (de)serialization ignores prefix.

**Environment**
Spring Batch 5.1.2, JDK 21

**Steps to reproduce**

```java
    @Test
    void qname() throws Exception {
        Jackson2ExecutionContextStringSerializer serializer = new Jackson2ExecutionContextStringSerializer();
        ExecutionContext context = new ExecutionContext();
        QName qname = new QName("http://namespace", "test", "p");
        context.put("qname", qname);
        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
        serializer.serialize(new HashMap<>(context.toMap()), outputStream);
        LOGGER.info(outputStream.toString(StandardCharsets.UTF_8));
        // <- {"@class":"java.util.HashMap","qname":["javax.xml.namespace.QName","{http://namespace}test"]}
        Map<String, Object> result = serializer.deserializeFromByteArray(outputStream.toByteArray());
        QName value = (QName) result.get("qname");
        assertThat(value.getLocalPart()).isEqualTo(qname.getLocalPart());
        assertThat(value.getNamespaceURI()).isEqualTo(qname.getNamespaceURI());
        assertThat(value.getPrefix()).isEqualTo(qname.getPrefix()); // <- fails
    }
```

**Expected behavior**
QName prefix is preserved in context (de)serialization roundtrip.

Spring Batch StaxEventItemWriter depends on QName prefix being preserved correctly:

https://github.com/spring-projects/spring-batch/blob/fc1f3fcfc791196273b1249157c4e860b1df9025/spring-batch-infrastructure/src/main/java/org/springframework/batch/item/xml/StaxEventItemWriter.java#L796-L799

jjh75607:
Some findings, since the situation has changed since this was reported.

[...]

PR #5502, opened 2026-08-26
Fix QName deserialization in JacksonExecutionContextStringSerializer

### LINKED ISSUE #4044: StaxEventItemWriter.unclosedHeaderCallbackElements prevents new job execution
While upgrading to spring-batch 4.3.4, my job started to fail with the exception :
`The class with javax.xml.namespace.QName and name of javax.xml.namespace.QName is not trusted.`

After investigating, it turns out that `StaxEventItemWriter` (with saveState=true) writes unclosed elements from the header callback to the step context as `QNames`. On the next job execution, it will first try to read the last job execution and this is where the exception appears as the step execution from `StaxEventItemWriter` cannot be deserialized.

This was not a problem before 4.3.4 as `SimpleJobExplorer#getLastJobExecution` did not fetch step executions (see #3943) but now that it does, it means that `StaxEventItemWriter` is not usable with a header callback out-of-the-box. EIther we have to set `saveState=false` on the writer, or we must create a BatchConfigurer to add the mixIn for `QName` to the `ObjectMapper`.

I think that at the very least, documentation should reflect this.

Minimal example : https://github.com/Alex4i/staxwriter_qname
Run with `gradlew bootRun`

jjh75607:
`JacksonExecutionContextStringSerializer` allows `java.xml.` in its `BasicPolymorphicTypeValidator`. But `java.xml` is a JPMS module name, not a package. The JDK package is `javax.xml.`, so the entry matches nothing and every `javax.xml.*` type is denied.

The type this hits is `javax.xml.namespace.QName`, which Spring Batch writes to the execution context itself. With `saveState` enabled, `StaxEventItemWriter` stores the elements its header callback left open as a `List<QName>`, and reads them back in `open()`.

Serialization succeeds, so the run completes normally. Only the restart fails, with `InvalidTypeIdException ... denied resolution`.

Both other serializers allow the class deliberately: `Jackson2ExecutionContextStringSerializer` since 398d52a5, which fixed #4044, and `DefaultExecutionContextSerializer` since 46768dc7.

The change is one line, `java.xml.` to `javax.xml.`. The added test round-trips a `QName` through the serializer. It fails on `main` with the exception above and passes with the change. The `spring-batch-infrastructure`, `spring-batch-core`, `spring-batch-test` and `spring-batch-integration` suites all pass.

If you would rather match the other two serializers exactly, `allowIfSubType("javax.xml.namespace.QName")` also works and I am happy to change it.

The same line is present in `v6.0.0` and on `6.0.x`, so I have left the backport judgement to you.

Related to but distinct from #4697. That issue is about the deprecated `Jackson2ExecutionContextStringSerializer` dropping a `QName` prefix. That behaviour is unchanged here and I will follow up there separately.

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JacksonExecutionContextStringSerializer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JacksonExecutionContextStringSerializer.java
@@ -56,7 +56,7 @@ public class JacksonExecutionContextStringSerializer implements ExecutionContext
 			.allowIfSubType("java.math.")
 			.allowIfSubType("java.time.")
 			.allowIfSubType("java.net.")
-			.allowIfSubType("javax.xml.")
+			.allowIfSubType("javax.xml.namespace.QName")
 			.allowIfSubType("org.springframework.batch.")
 			.build();
 		this.jsonMapper = JsonMapper.builder().activateDefaultTyping(polymorphicTypeValidator).build();
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JacksonExecutionContextStringSerializer.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/repository/dao/JacksonExecutionContextStringSerializer.java
@@ -36,6 +36,7 @@ import org.springframework.batch.core.repository.ExecutionContextSerializer;
  * provide your own {@link JsonMapper} instance through the constructor.
  *
  * @author Mahmoud Ben Hassine
+ * @author Soonjae Jung
  * @since 6.0.0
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Thank you for this PR! That is a valid point.

> If you would rather match the other two serializers exactly, allowIfSubType("javax.xml.namespace.QName") also works and I am happy to change it.

Yes please, for consistency with other serializers

jjh75607:
Done. `spring-batch-core` passes, 819 tests.

fmbenhassine:
Thank you for the update. Squashed and merged as c7d690e7fe9cd3263b397cdb0c74307d198f7b8b.

Back-ported to 6.0.x as well.

Issue #5507, opened 2026-08-30
TransientTransactionError in MongoSequenceIncrementer during parallel job executions


bcaillard:
**Bug description**
Issue [#4960](https://github.com/spring-projects/spring-batch/issues/4960) is not fully resolved. When running multiple batch instances in parallel with the default configuration, a transaction error is now consistently raised when incrementing a sequence.

```
Caused by: com.mongodb.MongoCommandException: Command execution failed on MongoDB server with error 251 (NoSuchTransaction): 'Transaction with { txnNumber: 2 } has been aborted.' on server xxx:27017. The full response is {""errorLabels"": [""TransientTransactionError""], ""ok"": 0.0, ""errmsg"": ""Transaction with { txnNumber: 2 } has been aborted."", ""code"": 251, ""codeName"": ""NoSuchTransaction"", ""$clusterTime"": {""clusterTime"": {""$timestamp"": {""t"": 1787842933, ""i"": 1}}, ""signature"": {""hash"": {""$binary"": {""base64"": ""9IjGm+ZQOCuAm/GkrrAqLJ9be9M="", ""subType"": ""00""}}, ""keyId"": 7641827476305346561}}, ""operationTime"": {""$timestamp"": {""t"": 1787842933, ""i"": 1}}}
  at com.mongodb.internal.connection.ProtocolHelper.getCommandFailureException(ProtocolHelper.java:210)
  at com.mongodb.internal.connection.InternalStreamConnection.receiveCommandMessageResponse(InternalStreamConnection.java:570)
  at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceiveInternal(InternalStreamConnection.java:494)
  at com.mongodb.internal.connection.InternalStreamConnection.lambda$sendAndReceive$1(InternalStreamConnection.java:383)
  at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceive(InternalStreamConnection.java:386)
  at com.mongodb.internal.connection.UsageTrackingInternalConnection.sendAndReceive(UsageTrackingInternalConnection.java:111)
  at com.mongodb.internal.connection.DefaultConnectionPool$PooledConnection.sendAndReceive(DefaultConnectionPool.java:757)
  at com.mongodb.internal.connection.CommandProtocolImpl.execute(CommandProtocolImpl.java:60)
  at com.mongodb.internal.connection.DefaultServer$DefaultServerProtocolExecutor.execute(DefaultServer.java:207)
[...]

PR #5508, opened 2026-08-31
Allow to use Mongo JobRepository with any PlatformTransactionManager

quaff:
Align to JDBC implementation, it's not required to couple with `MongoTransactionManager`, `ResourcelessTransactionManager` or other custom `PlatformTransactionManager`s should be supported.

See GH-5507


fmbenhassine:
Rebased and merged as 4534099f47dbce965e85203e5131f93a1247d6e7.

Issue #5512, opened 2026-09-01
Inconsistent builder method name: MultiResourceItemReaderBuilder.setStrict()


noojung:
`MultiResourceItemReaderBuilder` exposes its strict-mode toggle as `setStrict(boolean)`:
```java
public MultiResourceItemReaderBuilder<T> setStrict(boolean strict) {
	this.strict = strict;
	return this;
}
```

However, almost every other builder in Spring Batch (e.g., FlatFileItemReaderBuilder, StaxEventItemReaderBuilder, JsonItemReaderBuilder) uses the fluent `strict(boolean)` naming convention without the set prefix.

This appears to be an inconsistency in the builder API design.

This is my propsal:
- Add strict(boolean) to MultiResourceItemReaderBuilder to align with the rest of the builder APIs.
- Deprecate setStrict(boolean) for backward compatibility.

PR #5515, opened 2026-09-01
Add strict() method and deprecate setStrict() in MultiResourceItemReaderBuilder

UsmanEjaz10:
## Issue

MultiResourceItemReaderBuilder exposes its strict-mode toggle as setStrict(boolean), while almost every other builder in Spring Batch uses the fluent strict(boolean) naming convention without the set prefix.

## Solution

- Added strict(boolean) method to align with other Spring Batch builder APIs
- Deprecated setStrict(boolean) with @Deprecated(since = "6.1.0", forRemoval = true) for backward compatibility
- Added comprehensive test coverage for both methods

### MultiResourceItemReaderBuilder.java
- New method: strict(boolean strict) - follows fluent API convention
- Deprecated method: setStrict(boolean strict) with proper version annotation

### MultiResourceItemReaderBuilderTests.java
- testStrictMethodChaining() - verifies fluent API chaining works
- testSetStrictStillWorks() - confirms backward compatibility of deprecated method

## Backward Compatibility
- Old setStrict() method remains fully functional
- Deprecation warnings guide developers to new API
- IDE quick fixes can automate migration

This fix aligns MultiResourceItemReaderBuilder with Spring Batch's consistent builder API pattern, improving developer experience and API consistency across the framework.

### DIFF
--- a/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/file/builder/MultiResourceItemReaderBuilder.java
+++ b/spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/file/builder/MultiResourceItemReaderBuilder.java
@@ -34,6 +34,7 @@ import org.springframework.util.StringUtils;
  * @author Glenn Renfro
  * @author Drummond Dawson
  * @author Stefano Cordio
+ * @author Usman Ijaz
  * @since 4.0
  * @see MultiResourceItemReader
  */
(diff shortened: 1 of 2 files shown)


spring-batch-infrastructure/src/main/java/org/springframework/batch/infrastructure/item/file/builder/MultiResourceItemReaderBuilder.java
```diff
@@ -110,13 +111,30 @@ public MultiResourceItemReaderBuilder<T> delegate(ResourceAwareItemReaderItemStr
 	 * @param strict false by default.
 	 * @return this instance for method chaining.
 	 * @see MultiResourceItemReader#setStrict(boolean)
+	 * @deprecated as of 6.1.0 in favor of {@link #strict(boolean)}
 	 */
+	@Deprecated(since = "6.1.0", forRemoval = false)
```
quaff:
```suggestion
	@Deprecated(since = "6.1.0", forRemoval = true)
```

noojung:
I think this test code might be unnecessary. 
`strict` mode is already covered in `MultiResourceItemReaderIntegrationTests`, and the builder simply toggles a boolean flag.

noojung:
This PR seems to address issue #5512.

UsmanEjaz10:
> I think this test code might be unnecessary. `strict` mode is already covered in `MultiResourceItemReaderIntegrationTests`, and the builder simply toggles a boolean flag.

Yes the 2 base test cases are redundant. And I should remove them
I think we should keep these tests (`testStrictMethodChaining()` and `testSetStrictStillWorks()`) because they verify API contract compliance - fluent method chaining and backward compatibility - which are builder-specific concerns not covered by integration tests.

These tests verify:
1. Fluent API works correctly (chaining pattern is a key builder feature) - `testStrictMethodChaining()`
2. Deprecated API remains functional (critical for backward compatibility in a framework) -  `testSetStrictStillWorks()`

UsmanEjaz10:
Hi @quaff this PR will resolve the issue 5512. Could you please review it.

fmbenhassine:
LGTM 👍 Rebased, squashed and merged as 7ef98f9f00d9cbff37f9482f1b802c2e93b73039.

PR #5532, opened 2026-09-09
Fix skip annotation listener detection

### LINKED ISSUE #5531: @OnSkipInRead/@OnSkipInProcess/@OnSkipInWrite methods are ignored by ChunkOrientedStepBuilder#listener(Object)
**Bug description**

`ChunkOrientedStepBuilder#listener(Object)` does not scan for `@OnSkipInRead`, `@OnSkipInProcess` and `@OnSkipInWrite`, so an object whose only listener methods carry skip annotations is never wrapped in a listener proxy nor registered. The job completes with correct skip counts, but no skip callback is invoked.

`StepListenerMetaData` maps all three annotations to `SkipListener` and `registerTypedListener` registers it (#5347), so only the detection list is missing them. Same gap as #5451 (`@BeforeStep`/`@AfterStep`); in 5.x this was handled by `FaultTolerantStepBuilder#listener(Object)`.

**Environment**

Spring Batch 6.0.5 and current main, Java 25, no database.

**Expected behavior**

Each annotated method is invoked once, as when the object implements `SkipListener`.

```java
@Test
void skipAnnotationsShouldBeDetected() throws Exception {
	AtomicInteger skips = new AtomicInteger();
	Object listener = new Object() {

		@OnSkipInRead
		public void onSkipInRead(Throwable t) {
			skips.incrementAndGet();
		}

	};
	Iterator<String> items = List.of("fail", "ok").iterator();
	ChunkOrientedStep<String, String> step = new StepBuilder("step", new ResourcelessJobRepository())
		.<String, String>chunk(1)
		.reader(() -> {
			if (!items.hasNext()) {
				return null;
			}
			String item = items.next();
			if (item.equals("fail")) {
				throw new IllegalStateException();
			}
			return item;
		})
		.writer(chunk -> {
		})
		.faultTolerant()
		.skip(IllegalStateException.class)
		.listener(listener)
		.build();
	StepExecution stepExecution = new StepExecution(1L, "step",
			new JobExecution(1L, new JobInstance(1L, "job"), new JobParameters()));

	step.execute(stepExecution);

	assertEquals(1, stepExecution.getSkipCount()); // passes
	assertEquals(1, skips.get()); // fails: expected <1> but was <0>
}
```

`@OnSkipInProcess` and `@OnSkipInWrite` behave the same way.

acktsap:
Resolves #5531

### DIFF
--- a/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/ChunkOrientedStepBuilder.java
+++ b/spring-batch-core/src/main/java/org/springframework/batch/core/step/builder/ChunkOrientedStepBuilder.java
@@ -37,6 +37,9 @@ import org.springframework.batch.core.annotation.BeforeWrite;
 import org.springframework.batch.core.annotation.OnChunkError;
 import org.springframework.batch.core.annotation.OnProcessError;
 import org.springframework.batch.core.annotation.OnReadError;
+import org.springframework.batch.core.annotation.OnSkipInProcess;
+import org.springframework.batch.core.annotation.OnSkipInRead;
+import org.springframework.batch.core.annotation.OnSkipInWrite;
 import org.springframework.batch.core.annotation.OnWriteError;
 import org.springframework.batch.core.listener.ChunkListener;
 import org.springframework.batch.core.listener.ItemProcessListener;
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Thank you for reporting this issue and for contributing a fix!

Issue #4551, opened 2024-02-15
Improve recommendations for indexing metadata tables


vpavic:
[_Recommendations for Indexing Metadata Tables_](https://docs.spring.io/spring-batch/reference/schema-appendix.html#recommendationsForIndexingMetaDataTables) section of the reference manual provides some guidance for indexing Spring Batch tables. However, this is focused solely on where clauses in SQL statements while it should arguably also consider foreign keys as indexing those plays important role in speeding up queries involving joins.

On PostgreSQL, detecting such columns can be done using [something like this](https://wiki.postgresql.org/wiki/Unindexed_foreign_keys).

wondywang:
I hope someone can provide the best solution for MySQL or Oracle.

PR #5550, opened 2026-09-19
Update the indexing recommendations in the schema appendix

### LINKED ISSUE #5549: Indexing recommendations table in the schema appendix does not match the queries issued since 4.2
The [Recommendations for Indexing Metadata Tables](https://docs.spring.io/spring-batch/reference/schema-appendix.html#recommendationsForIndexingMetaDataTables) section lists the `WHERE` clauses that the DAO implementations use, so that users can decide which columns to index. The table has not changed since it was written and no longer matches the queries in `JdbcStepExecutionDao` and `JdbcJobExecutionDao`. I checked against 6.0.5 and `main`, where the queries below are identical.

### `BATCH_STEP_EXECUTION`, `STEP_NAME = ? and JOB_EXECUTION_ID = ?`, "Before each step execution"

No query in `JdbcStepExecutionDao` uses this clause. Since 4.2.0 (commit 62a8f441, BATCH-2716), `SimpleStepHandler` finds the last step execution through `JobRepository#getLastStepExecution(JobInstance, String)`, and the DAO runs a join. `countStepExecutions` (4.3.0, commit b50c665f) uses the same join:

```sql
SELECT ...
FROM BATCH_JOB_EXECUTION JE
    JOIN BATCH_STEP_EXECUTION SE ON SE.JOB_EXECUTION_ID = JE.JOB_EXECUTION_ID
WHERE JE.JOB_INSTANCE_ID = ? AND SE.STEP_NAME = ?
```

So the columns hit before each step execution are `BATCH_JOB_EXECUTION.JOB_INSTANCE_ID`, `BATCH_STEP_EXECUTION.JOB_EXECUTION_ID` (the join column) and `BATCH_STEP_EXECUTION.STEP_NAME`.

### `BATCH_STEP_EXECUTION`, `VERSION = ?`, "On commit interval"

The actual clause is `STEP_EXECUTION_ID = ? AND VERSION = ?` (`UPDATE_STEP_EXECUTION` and `DELETE_STEP_EXECUTION`). The primary key already narrows it to one row, so an index on `VERSION` does not help. The table's title says primary keys are excluded, and this row should be excluded for the same reason.

### `BATCH_JOB_EXECUTION_PARAMS`, `JOB_EXECUTION_ID = ?` is missing

`JdbcJobExecutionDao#getJobExecution(long)` loads the parameters with `FIND_PARAMS_FROM_ID` every time a job execution is loaded. On a launch for an existing job instance, `TaskExecutorJobLauncher` loads every execution of the instance, so the query runs once per execution. The table has no primary key, and this is the column #5360 was about: since 6.0.4 the Oracle script creates `BATCH_JOB_EXEC_PARAMS_IDX` for it, the other scripts do not, and the appendix does not mention it.

### `BATCH_JOB_EXECUTION`, `JOB_INSTANCE_ID = ?`, "Every time a job is restarted"

Still accurate for `GET_LAST_JOB_EXECUTION_ID` and `GET_JOB_EXECUTION_IDS_BY_INSTANCE_ID`, but `TaskExecutorJobLauncher` runs both on every launch for an existing job instance, not only on a restart. Minor compared with the rows above.

## Related

#4551 asks to add foreign key columns to this section (PRs #5419 and #5425). This issue is about the existing rows not matching the code, so I kept it separate. I will submit a PR that updates the table.

benelog:
Resolves #5549

The table now lists the clauses that `JdbcJobInstanceDao`, `JdbcJobExecutionDao` and `JdbcStepExecutionDao` run in 6.x. The `BATCH_STEP_EXECUTION` row with `STEP_NAME = ? and JOB_EXECUTION_ID = ?` is replaced by the join on `BATCH_JOB_EXECUTION` that `getLastStepExecution` and `countStepExecutions` use, and the `VERSION = ?` row is removed because that clause always comes with the primary key. Two rows are added for the queries that run every time a job execution is loaded: `JdbcJobExecutionDao#getJobExecution(long)` reads `BATCH_JOB_EXECUTION_PARAMS` by `JOB_EXECUTION_ID`, and `SimpleJobExplorer#fillJobExecutionDependencies` reads `BATCH_STEP_EXECUTION` by the same column to attach the step executions.

I also added a note that the unique constraint on `JOB_NAME` and `JOB_KEY` is usually backed by an index and that the Oracle script already creates one for `BATCH_JOB_EXECUTION_PARAMS`, so that readers do not create duplicate indexes. Happy to drop the note if you prefer the table alone.

The foreign key columns discussed in #4551 are left to that issue.


benelog:
Added one more row: `SimpleJobExplorer#fillJobExecutionDependencies` loads the step executions of every job execution it returns with a query on `BATCH_STEP_EXECUTION` by `JOB_EXECUTION_ID`, so that column is hit as often as the job parameters lookup. I will squash the two commits once the review is done.

